PR #4, opened 2017-11-03
Add Flow / ESLint / Prettier

bpierre:


.eslintrc
```diff
@@ -0,0 +1,37 @@
+{
+  "env": {
+    "browser": true,
+    "es6": true
+  },
+  "extends": [
```
harshjv: can we use standard-js? 

```
- eslint-config-standard
- eslint-config-standard-react
```

bpierre: Sure! Adding it

harshjv: Yea!

PR #9, opened 2017-11-19
Set the assets public URL at runtime

bpierre:
There are two different ways to handle a library as a dependency:

- As a source library
- As a runtime library, transpiled to “universal JS”

The problem with the source is that it brings many problems to make it work with the parent app:

- Having to install the same dependencies in the parent project.
- Having to install the same build dependencies.
- Having to configure the build exactly the same way.
- Update these every time the library is updated.

Using the transpiled library solves all this problems, except one. Most modern build systems have some mechanism to import non-JS dependencies, which allows, in the context of a components library, to have the assets of a component in the same directory (instead of using a global folder).

When the modules are transpiled, the link between the importing module and the assets is lost in the process, and all the links to the assets are broken.

The solution implemented by this commit is the following:

- The library is transpiled to “universal JS”, with and without ES6 modules (for tree shaking, if the parent project build system supports it).
- The parent project imports the transpiled version of Aragon UI, and also copy the Aragon UI assets into a directory where they can be served by a web server.
- The parent project wraps the Aragon UI app into AragonApp, which allows to set the public URL of the toolkit assets: `<AragonApp publicUrl="/">`
- That way, each Aragon UI component using an asset is aware of the URL where they can be fetched from.

### DIFF
--- a/gallery/src/App.js
+++ b/gallery/src/App.js
@@ -1,7 +1,7 @@
 import React from 'react'
 import createHistory from 'history/createBrowserHistory'
 import styled from 'styled-components'
-import { BaseStyles } from '@aragon/ui'
+import { AragonApp } from '@aragon/ui'
 import Sidebar from 'comps/Sidebar/Sidebar'
 import PageHome from 'pages/PageHome'
 import PageColors from 'pages/PageColors'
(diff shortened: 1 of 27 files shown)


luisivan:
I think this makes sense, thanks Pierre

PR #15, opened 2017-11-27
Button: add flowtypes

sohkai:
Adds flow types to Button's props.

### DIFF
--- a/src/components/Button/Button.js
+++ b/src/components/Button/Button.js
@@ -1,4 +1,5 @@
 /* @flow */
+import type { ComponentType } from 'react'
 import styled, { css } from 'styled-components'
 import theme from '../../theme'
 import { fontStyle } from '../../shared-styles'
@@ -106,7 +107,17 @@ const negativeStyle = css`
   background: url(${asset(cross)}) no-repeat 10px calc(50% - 1px);
 `
 
-const StyledButton = getPublicUrl(styled.button`
+type Mode = 'normal' | 'strong' | 'outline' | 'text'
+type Emphasis = 'positive' | 'negative'
+type Props = {
+  compact?: boolean,
+  emphasis?: Emphasis,
+  mode?: Mode,
+  wide?: boolean,
+}
+
+// Flow declaration: see https://github.com/styled-components/styled-components/issues/570#issuecomment-332087358
+const StyledButton: ComponentType<Props> = getPublicUrl(styled.button`
   width: ${({ wide }) => (wide ? '100%' : 'auto')};
   padding: 10px 15px;
   white-space: nowrap;
--- a/src/components/Button/Button.js
+++ b/src/components/Button/Button.js
@@ -127,11 +127,11 @@ const StyledButton = getPublicUrl(styled.button`
     border: 0;
   }
 
-  ${({ mode = 'normal' }) => {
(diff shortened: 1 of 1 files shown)


bpierre:
What about doing both? If the mode doesn’t exist:

- Throw a warning to let the author know that the mode is being ignored.
- Default to normal mode.

And if the parent project is using flow, the type definition (that we need to add) triggers an error before even running it.

The reason why I think we should default to normal mode is that the base styles are not meant to be used without any mode style (`modeNormal`, `modeStrong`, etc.).

Another possibility would be to remove any styling if the mode is not defined, leaving the default style of the HTML button. But since we are defining a set of base components, I think the default should be our own default which is `<Button mode="normal" />`.

We could also throw an error as you are suggesting. It might be a bit strict, but I think it’s a good solution too. If we go for this solution, we should probably align the other properties and components to this level of strictness.

What do you think?

sohkai:
> What about doing both? If the mode doesn’t exist:

> Throw a warning to let the author know that the mode is being ignored.
> Default to normal mode.

❤️. The proptype or flow warnings should be enough of an error, and yeah, applying the default style is a good fallback. I originally wanted to remove the styling to make the button pop out for the dev, but explicit errors are way better.

src/components/Button/Button.js
```diff
@@ -106,7 +107,17 @@ const negativeStyle = css`
...
+type Emphasis = 'positive' | 'negative'
+type Props = {
+  compact?: boolean,
+  emphasis?: Emphasis,
+  mode?: Mode,
+  wide?: boolean,
+}
+
+// Flow declaration: see https://github.com/styled-components/styled-components/issues/570#issuecomment-332087358
+const StyledButton: ComponentType<Props> = getPublicUrl(styled.button`
```
sohkai: Note that doing this [apparently breaks `styled-component`'s exposed component API](https://github.com/styled-components/styled-components/issues/570#issuecomment-332087358), but that's not a problem here since `getPublicUrl()` already does that.

src/components/Button/Button.js
```diff
@@ -106,7 +107,17 @@ const negativeStyle = css`
...
 
-const StyledButton = getPublicUrl(styled.button`
+type Mode = 'normal' | 'strong' | 'outline' | 'text'
+type Emphasis = 'positive' | 'negative'
+type Props = {
+  compact?: boolean,
+  emphasis?: Emphasis,
+  mode?: Mode,
+  wide?: boolean,
+}
```
sohkai: We could manually duplicate flowtypes into proptypes for the few cases where we have enums like this, or we could transform all our flowtypes into proptypes via something like a [babel-plugin](https://github.com/brigand/babel-plugin-flow-react-proptypes).

I like the plugin, since it's less hassle once set up and provides proptypes for all components. I only worry it requires everything to be well typed, which we currently don't do (right now our flowtypes are more like documentation than full type checks 😅).

Thoughts @bpierre?

bpierre: Nothing against setting up the babel plugin, I usually duplicate the enums for checking, since I can’t access the types at runtime.

src/components/Button/Button.js
bpierre: Yes it shouldn’t be a problem, since we don’t want to expose this API outside.

PR #16, opened 2017-11-27
Text: refactor to always render as <span>

sohkai:
Refactors `Text` to be more similar to `Button` (i.e. no element-switching).

### DIFF
--- a/src/components/Text/Text.js
+++ b/src/components/Text/Text.js
@@ -1,11 +1,24 @@
 /* @flow */
-import type { Node } from 'react'
+import type { ComponentType, ElementType, Node } from 'react'
 import React from 'react'
 import styled from 'styled-components'
 import theme from '../../theme'
 import { fontStyle } from '../../shared-styles'
 
-const StyledText = styled.p`
+type Props = {
+  children?: Node,
+  color?: string,
+  size?: string,
+  smallcaps?: boolean,
+  weight?: string,
+}
+
+type TextComponentType = {
+  Block: ComponentType<Props>,
+  Paragraph: ComponentType<Props>,
+}
+
+const StyledText = styled.span`
   ${({ size, weight }) => fontStyle({ size, weight })};
   ${({ smallcaps }) => {
     if (!smallcaps) return ''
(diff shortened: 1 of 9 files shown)


src/components/Text/Text.js
```diff
@@ -32,60 +31,39 @@ const styledHeadings = [
...
-    )
-  }
-  return StyledText
-}
-
 type Props = {
   block: boolean,
   inline: boolean,
-  heading: string | number,
+  heading: string | number, // Make this just a number?
```
sohkai: We'd probably have to fix all the `Text` instances using `<Text heading="1">` across a few repos, but I don't see much value in supporting both a string and number.

`<Text heading={1}>` seems pretty normal to me ¯\_(ツ)_/¯

src/components/Text/Text.js
```diff
@@ -32,60 +31,39 @@ const styledHeadings = [
...
-  color: null,
-}
+const Text = ({ block, inline, heading, ...props }: Props) => {
+  /*
+  const Component =
+    (block && StyledBlock) ||
+    (inline && StyledInline) ||
+    (heading >= 1 && heading <= 6 && styledHeadings[heading - 1]) ||
+    StyledText
+  */
```
sohkai: Changing `heading` to only accepting numbers unlocks this refactor of the component selecting logic!

bpierre:
I am starting to think that our components should try to avoid swapping the HTML element based on properties, when possible. I have three solutions in mind:

## Shared styles

We could recommend to always use the `font()` function, and let people define their own text components. It could work, but it would break the idea of being able to build an Aragon app using only default components. 👎 

## Using `<span>` as an inline, neutral element

In this case, we could use a `<span>` element only.

It would give us the following advantages compared to the current implementation:

- A single styled component is easier to manage.
- The role of the component is clearer, as it would only deal with styles.
- The user would be free to decide what HTML semantics to apply (which is quite useful in the particular case of HTML h1…6 elements).

Instead of this:

```jsx
<Text heading={1}>Hello World</Text>
```

We could have like this (where `mode` would contain a series of text styles):

```jsx
<h1><Text mode="heading-large">Hello World</Text></h1>
```

## As a styling container (à la React Native `Text`)

I like how the `Text` component works in React Native: https://facebook.github.io/react-native/docs/text.html

A way to implement this could be to use a `<span>` if `children` contains any combination of strings and inline elements, and a `<div>` otherwise.

The problem is that we implicitly swap the HTML element. To avoid this, we could render `<Text>` as a `<span>`, and have another component that would render as a `<div>` (e.g. `<TextBlock>`).

What do you think? 🤔

sohkai:
I prefer `Text` always rendering to a `span`, especially given our current logic for `Text` (i.e. always the same styles, just different backing elements).

`TextBlock` (or `Paragraph`) will likely be convenient, but in my mind they're not absolutely necessary given they're really simple _container_ elements that map 1:1 to an element (`<div>` and `<p>`).

I also wouldn't mind keeping the current props to let the user control them as they like. Rather than 
```js
<h1><Text mode="heading-large">Hello World</Text></h1>
```
 I think
```js
<h1><Text size="large" weight="bold">Hello World</Text></h1>
```
is perfectly fine.

If we usually use the same styling for headers, we could also provide convenience components for those (with escape hatches to pass down props to the backing `<Text>`, so the above could look like
```js
<H1>Hello World</H1>
```

bpierre:
> TextBlock (or Paragraph) will likely be convenient, but in my mind they're not absolutely necessary given they're really simple container elements that map 1:1 to an element (`<div>` and `<p>`).

Agree, let’s try with the `<span>` solution and see how it works for us.

> I also wouldn't mind keeping the current props to let the user control them as they like.

Sorry I wasn’t clear, the idea of the `mode` property was to have some presets defining some named styles that we often use, but nothing that can’t be done using the other properties. It would not replace `size`, `weight`, etc., and should be overridden by them in case of conflict.

> we could also provide convenience components for those (with escape hatches to pass down props to the backing `<Text>`)

We could also do that to provide these styles, maybe by attaching these components to `Text`: `Text.HeadingLarge`, `Text.SectionLabel`, etc. (I am trying to avoid names like `H1` to make it clear that it doesn’t render as a semantic element)

sohkai:
Ahh, gotcha. I prefer attaching oft-used components to `Text`; they could be really simple components that just have default props to set styles. It's a lot simpler than `mode`, and makes the intent obvious (in general, I think this is a pretty neat pattern: export a base component, and then attach default behaviour / styling as static "children" components).

bpierre:
OK, let’s use attached components for the predefined styles. 👍 

And since we are attaching components to `Text`, what about adding `Text.Block` (or `Text.Div`) as well?

src/components/Text/Text.js
```diff
@@ -16,96 +28,39 @@ const StyledText = styled.p`
...
+    children,
+    color,
+    size,
+    smallcaps,
+    weight,
+    ...props
+  }: Props) => {
+    const textProps = { color, size, smallcaps, weight }
     return (
-      styledHeadings[Math.max(1, Math.min(6, parseInt(heading, 10))) - 1] ||
```
sohkai: Note that this was actually causing all `Text`s to be rendered as `h1`s by default.

This shouldn't have had an impact on the styling though, since both size and weight are set by props and there's nothing terribly special for `h1`s in base styles.

src/components/Text/Text.js
```diff
@@ -16,96 +28,39 @@ const StyledText = styled.p`
...
-      color={color}
-      children={children}
-    />
-  )
-}
+Text.Block = createTextContainer('div')
+Text.Paragraph = createTextContainer('p')
 
-Text.defaultProps = DefaultProps
+const TypedText: ComponentType<Props> & TextComponentType = Text
```
sohkai: This is kind of a black magic flow hack, but it seems to work.

`TextComponentType` must declare `Block` and `Paragraph` as existing (rather than as `Maybe`s / optionals) if they're to be used to render other components (otherwise flow will complain they might not exist).

PR #18, opened 2017-11-27
publicURL: add injector component

sohkai:
Adds a new `publicUrlInjector()` HOC to decouple the publicURL context injector from AragonApp.

This allows consumers of the library to choose their root element, rather than having to always wrap an `AragonApp` around it.

Thoughts @bpierre?

Todo:
* [ ] Fix docs
* [ ] Fix other repos relying on previous behaviour

### DIFF
--- a/src/public-url.js
+++ b/src/public-url.js
@@ -1,7 +1,8 @@
+/* @flow */
 import React from 'react'
 import PropTypes from 'prop-types'
 
-// High order component wrapper
+// High order component "get" wrapper
 const getPublicUrl = Component => {
   const highOrderComponent = (baseProps, context) => {
     const { publicUrl = '' } = context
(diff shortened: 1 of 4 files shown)


src/public-url.js
```diff
@@ -14,12 +15,32 @@ const getPublicUrl = Component => {
...
 
+type UrlInjectorProps = {
+  publicUrl: string,
+}
+
+// High order component "set" wrapper
+const publicUrlInjector = Component => {
+  class PublicUrlContextInjector extends React.Component<UrlInjectorProps> {
+    getChildContext() {
+      return { publicUrl: this.props.publicUrl }
```
sohkai: Had to fiddle with `publicUrl` today because I didn't realize it always needed a trailing `/`. We could use something like `_.memoize()` to always stuff a trailing `/` (I find this less surprising behaviour, especially since you don't always get the trailing `/`, e.g. `pwd`).

sohkai:
Replaced by #158.

PR #20, opened 2017-11-27
Add DefaultAragonApp

sohkai:
Requires #18.

Adds a `DefaultAragonApp` that sets up the default styling (with pluggable appbar and empty state card) for all Aragon apps.


sohkai:
Replaced by #164.

Issue #13, opened 2017-11-27
Use `create-index`


sohkai:
Add https://github.com/gajus/create-index and autogenerate `index.js` files for each component folder.

------------

@bpierre and I were also discussing what we should do when we need to use "internal" components in more than one place. One idea would be to allow named exports (i.e. `export { ... }`) from a component so that internal components could access them, but then to not export them to consumers by only re-exporting the default export from each component.

An example, for better clarity:
```js
// in component.js
const InternalComponent = {}
const Component = {} // requires InternalComponent

export { InternalComponent } // this export would be accessible to components inside the library, in case they wanted to restyle it slightly or reuse it
export default Component // this export would be the only component available to outside consumers of the library
```

```js
// in index.js
export { default as Component } from "component.js"
```

bpierre:
Agree, we could also only generate the file `src/components/index.js`, and keep `src/index.js` as a non-generated file to have some control over it:

```js
// src/components/index.js

// DO NOT EDIT: file generated by “npm run update-index”; any edit will be lost.
export { default as ComponentA } from "./ComponentA"
export { default as ComponentB } from "./ComponentB"
```

```js
// src/index.js
export * from './components'
export { something, somethingElse } from './some-module'
export { anotherThing } from './another-module'
```

Also, if we strictly follow the rule of one default export per component, we probably don’t even need `create-index`, we could do something like this instead:

```sh
#!/bin/sh
# scripts/update-index.sh

# Generates components/index.js and components/*/index.js files.

ls -d src/components/*/ \
  | xargs basename \
  | awk '{
      print "export * from \"./"$1"\"\nexport default from \"./"$1"\"" > "src/components/"$1"/index.js"
      print "export { default as "$1" } from \"./"$1"\""
    }' \
  > src/components/index.js

```

onbjerg:
Won't this sort of obfuscate what is exported for end-users?

I personally use the source code to navigate around and if I had no idea you were auto-generating the exports (in the future) then I'd be somewhat confused as to what I can import.

Thoughts?

bpierre:
@onbjerg would that be a problem if we leave the generated code in the repository, like we do for icon components?

onbjerg:
No 😊

sohkai:
Closing; moving to this way of exporting components does not seem very valuable.

PR #21, opened 2017-11-27
Add AppBar

sohkai:
Adds default Aragon app bar with pluggable child elements (some apps only have a title in the app bar; the chevron will be hidden in these cases).


src/components/Text/Text.js
```diff
@@ -23,13 +23,19 @@ const StyledText = styled.p`
...
-  StyledText.withComponent('h3'),
-  StyledText.withComponent('h4'),
-  StyledText.withComponent('h5'),
-  StyledText.withComponent('h6'),
+  StyledH1,
+  StyledH2,
+  StyledH3,
+  StyledH4,
+  StyledH5,
+  StyledH6,
```
sohkai: Follows export convention suggested in #13.

src/components/Text/Text.js
```diff
@@ -87,5 +93,15 @@ const Text = ({
...
+  StyledBlock,
+  StyledInline,
+  StyledH1,
+  StyledH2,
+  StyledH3,
+  StyledH4,
+  StyledH5,
+  StyledH6,
+  StyledText,
+}
```
sohkai: Follows export convention suggested in #13 (exporting internal components).

PR #22, opened 2017-11-27
Add EmptyStateCard

sohkai:


src/components/Card/EmptyStateCard.js
```diff
@@ -0,0 +1,64 @@
...
+  display: flex;
+  padding: 40px 60px;
+  align-items: center;
+  text-align: center;
+  section {
+    padding-top: 20px;
+  }
+`
+
+const StyledHeading = StyledH1.extend`
```
sohkai: This could also be

```js
const StyledHeading = styled(Text).attrs({ heading: 1 })`
...
`
```

But that seems unnecessary since `.extend()` was pretty much built to handle this use case.

bpierre: I agree, `styled(Text).attrs({ heading: 1 })`` ` should be used from outside, but we can safely use the styled components from inside the toolkit.

PR #24, opened 2017-11-27
Add SidePanel

sohkai:


src/components/SidePanel/assets/close.svg
```diff
@@ -0,0 +1,14 @@
...
+    <desc>Created with Sketch Beta.</desc>
+    <defs></defs>
+    <g id="WebApp-1366px---Groups---Add" stroke="none" stroke-width="1" fill="none" fill-rule="evenodd" transform="translate(-1336.000000, -20.000000)" opacity="0.699999988">
+        <g id="Modal" transform="translate(890.000000, 0.000000)" fill="#6D777B">
+            <g id="Close" transform="translate(446.000000, 20.000000)">
+                <polygon id="Close_Icon_dark" points="10 1.0137457 6.0137457 5 10 8.9862543 8.9862543 10 5 6.0137457 1.0137457 10 0 8.9862543 3.9862543 5 0 1.0137457 1.0137457 0 5 3.9862543 8.9862543 0"></polygon>
+            </g>
+        </g>
+    </g>
+</svg>
```
sohkai: We should optimize this. @bpierre it'd be nice to document how and which settings we use to optimize our svgs.

bpierre: PR: https://github.com/aragon/aragon-ui/pull/26

src/components/SidePanel/SidePanel.js
```diff
@@ -0,0 +1,110 @@
...
+import React from 'react'
+import styled from 'styled-components'
+import { Motion, spring } from 'react-motion'
+import Text from '../Text/Text'
+import { lerp } from '../../math-utils'
+import getPublicUrl, { styledPublicUrl as asset } from '../../public-url'
+
+import close from './assets/close.svg'
+
+const PANEL_SPRING = { stiffness: 150, damping: 18, precision: 0.001 }
```
bpierre: This should be moved here (probably as `"slow"`) https://github.com/aragon/aragon-ui/blob/master/src/shared-styles.js#L27

PR #25, opened 2017-11-27
Add Table, TableCell, TableHeader, and TableRow

sohkai:
Adds table components meant to be used together.

All tables are expected to be eventually rendered into the following structure, even if intermediate custom components are supplied:

```jsx
<Table>
  <TableRow>
    <TableCell />
    ...
  </TableRow>
  ...
</Table>
```


src/components/Table/TableRow.js
```diff
@@ -0,0 +1,6 @@
+/* @flow */
+import styled from 'styled-components'
+
+const StyledTableRow = styled.tr``
+
+export default StyledTableRow
```
sohkai: This component is kind of useless as-is, but I wanted to add this in case we decide to make `<tr>`s more complex in the future. It also gives consumers a component to subclass their own custom table rows from.

PR #28, opened 2017-11-30
Optimize SVG files via svgo

sohkai:
Generated via `npm run optimize:svg`.

A quick run through the gallery seems fine, although I don't know if we can just take out `viewBox` and have everything still work as it did before.


bpierre:
> I don't know if we can just take out viewBox and have everything still work as it did before.

It should be fine, they are strictly identical than `width` and `height`.

bpierre:
Ah I just noticed that it prevents scaling on resize, we should probably disable this in the svgo config.

bpierre:
I added some commits, it’s now ready to be squashed and merged:

- Revert the optimization commit
- Merge `master` (which contains https://github.com/aragon/aragon-ui/pull/31)
- Optimize SVG files again

PR #30, opened 2017-11-30
Icons

bpierre:
This PR adds a set of icons:

image

They are SVG images, converted to components using [svgr](https://github.com/smooth-code/svgr) (to have them as inline SVGs and make it possible to change their color).

To prevent having all the icons in the final build, they are exported individually (prefixed with `Icon`):

```jsx
import { IconHome } from '@aragon/ui'

const App = () => <IconHome />
```

We could also have an `<Icon />` component for convenience, with an explanatory warning about [tree shaking optimization](https://webpack.js.org/guides/tree-shaking/):

```jsx
import { Icon } from '@aragon/ui'

const App = () => <Icon name="home" />
```

Another strategy could be to delay exporting the individual icons for now. These icon components are relatively small at the moment (17kb more, uncompressed), and it shouldn’t really be a problem until we have lots more.


src/icons/index.js
```diff
@@ -0,0 +1,11 @@
...
+export {default as IconFundraising} from './components/Fundraising.js'
+export {default as IconGroups} from './components/Groups.js'
+export {default as IconHome} from './components/Home.js'
+export {default as IconIdentity} from './components/Identity.js'
+export {default as IconNotifications} from './components/Notifications.js'
+export {default as IconPermissions} from './components/Permissions.js'
+export {default as IconSettings} from './components/Settings.js'
+export {default as IconTokens} from './components/Tokens.js'
+export {default as IconVoting} from './components/Voting.js'
+export {default as IconWallet} from './components/Wallet.js'
```
sohkai: Not sure if this is auto generated by svgr, and if we can change it or not, but it'd be nice if it matched our other exports as `export { default ... } from '...'`

src/index.js
```diff
@@ -1,3 +1,4 @@
+export * from './icons'
```
sohkai: Re-exporting it like this shouldn't be a problem for tree-shaking, yeah? It's only if you re-export it like `export * as Icons from './icons'` that would break tree-shaking?

bpierre: Exactly, as long as you don’t create an object, exports are static and can be optimized by tree shaking.

src/icons/index.js
bpierre: It’s not generated by svgr but I agree, we should generate these too (#13).

PR #34, opened 2017-12-01
BadgeNumber

bpierre:
Depends on https://github.com/aragon/aragon-ui/pull/33.

This PR adds a `<BadgeNumber />` component:

image

It can be used to indicate the available notifications and other things (updates, messages, etc.). Its color is `theme.positive` for now, that could evolve in the future.

### Positioning

Maybe we could also provide something to help positioning it on top of an icon, like this:

image

We could add a property to icon components?

```jsx
<IconWallet badgeNumber={7} />
```

But we might want to position it on top of other things than icons.

Or by having an wrapping component:

```jsx
<BadgeNumber.Attacher>
  <IconWallet />
  <BadgeNumber number={7} />
</BadgeNumber.Attacher>
```

Or by considering that app authors can position it themselves.

Any thoughts?

### DIFF
--- a/src/components/BadgeNumber/BadgeNumber.js
+++ b/src/components/BadgeNumber/BadgeNumber.js
@@ -1,6 +1,7 @@
 import React from 'react'
-
-import { styled, theme, font } from '../..'
+import styled from 'styled-components'
+import theme from '../../theme'
+import { font } from '../../shared-styles'
 
 const BadgeNumber = ({ number = 0, small = false, ...props }) => (
   <Main small={small} {...props}>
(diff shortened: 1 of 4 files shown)


src/components/BadgeNumber/BadgeNumber.js
```diff
@@ -0,0 +1,27 @@
+import React from 'react'
+
+import { styled, theme, font } from '../..'
```
sohkai: 🌶 ♨️ Saucy; I think I actually prefer the more verbose single imports from the actual packages (for `styled`), but for files an argument could be made either way.

bpierre: You’re right! It’s this way because this component started its life in https://github.com/aragon/aragon/pull/68, and I forgot to update this. I agree it’s better to import the package directly, if only for consistency. I’ll change it, thanks.

PR #38, opened 2017-12-04
Add syncing instructions for create-react-app or react-scripts users

sohkai:
As discovered in https://github.com/aragon/aragon-apps/pull/42

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,10 +23,30 @@ const App = () => (
 To get the path of the directory from where the assets need to be copied, use
 `path.dirname(require.resolve('@aragon/ui'))`.
 
+### `create-react-app`
+
+If you're using [`create-react-app`](https://github.com/facebookincubator/create-react-app)
+or [`react-scripts`](https://github.com/facebookincubator/create-react-app/tree/master/packages/react-scripts),
+you can copy over the contents of module's `dist/` folder (i.e.
+`path.dirname(require.resolve('@aragon/ui'))`) to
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -23,10 +23,30 @@ const App = () => (
...
+`path.dirname(require.resolve('@aragon/ui'))`) to
+a [`public/aragon-ui`](https://github.com/facebookincubator/create-react-app/blob/master/packages/react-scripts/template/README.md#adding-assets-outside-of-the-module-system)
+folder in your app and then use `publicUrl="/aragon-ui/`.
+
+You may also want to add this sync step to your build scripts, in case you
+later upgrade this package:
+
+```json
+"scripts": {
+    "sync-assets": "sync-files $(node -e 'console.log(path.dirname(require.resolve(\"@aragon/ui\")))') public/aragon-ui",
```
sohkai: 🙈 I didn't want to add a script like I did in the [Groups app](https://github.com/aragon/aragon-apps/pull/42/commits/febedfe21aa4e4d4afe0721ec823778a1b762348#diff-23b377c1d6bd047480a57649fa235b14), so I came up with this one-liner...

bpierre: Neat! I can’t see `sync-files` in the `react-scripts` deps, so we should probably add to the instructions that it need to be installed.

Or we could tell people to use rsync? The example is *nix anyway:

```sh
rsync -rtu "$(dirname $(node -p 'require.resolve("@aragon/ui")'))" public/aragon-ui
```

We can also replace `console.log` with `-p` 🤗

sohkai: Ahh, I'll add a note to install `sync-files`; mind as well keep things cross platform if it's not too difficult :).

Good call on the `-p` and `dirname`!

PR #60, opened 2017-12-06
Dropdown fixes

bpierre:
A few fixes for the DropDown component:

- Use `role` attributes to make the component more accessible by screen readers.
- Revert back to use `<div>` for DropDownItem (the item is also used outside of `<ul>`).
- Default height when the component is empty (so that it doesn’t look broken).
- Fix a typo in the README.
- Add a `wide` property.

### DIFF
--- a/src/components/DropDown/DropDown.js
+++ b/src/components/DropDown/DropDown.js
@@ -22,7 +22,7 @@ const StyledDropDown = styled.div`
   }
 `
 
-const DropDownItems = styled.ul`
+const DropDownItems = styled.div`
   display: ${({ opened }) => (opened ? 'block' : 'none')};
   min-width: 100%;
   padding: 8px 0;
@@ -108,6 +108,7 @@ class DropDown extends React.Component<Props, State> {
               const scale = opened ? lerp(openProgress, 0.95, 1) : 1
               return (
                 <DropDownItems
+                  role="listbox"
                   opened={openProgress > 0}
                   style={{
                     transform: `scale(${scale},${scale})`,
@@ -116,6 +117,7 @@ class DropDown extends React.Component<Props, State> {
                 >
                   {items.map((item, i) => (
                     <DropDownItem
+                      role="option"
                       key={i}
                       index={i}
                       active={i === active}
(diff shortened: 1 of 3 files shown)


src/components/DropDown/DropDownItem.js
```diff
@@ -74,7 +76,12 @@ class DropDownItem extends React.Component<Props, State> {
     this.setState({ displayFocus: !this.state.pressed })
   }
   render() {
-    const { children, className, mainRef, active } = this.props
+    const {
+      children = NON_BREAKING_SPACE,
```
sohkai: Should this be a `defaultProp` instead?

src/components/DropDown/DropDown.js
```diff
@@ -10,6 +10,8 @@ import getPublicUrl, { styledPublicUrl as asset } from '../../public-url'
 import DropDownItem from './DropDownItem'
 import arrow from './assets/arrow-down.svg'
 
+const NON_BREAKING_SPACE = '\xa0'
```
sohkai: Interesting, never seen this before. Is this better than `&nbsp;` ([is it the same but just a different encoding?](https://stackoverflow.com/questions/7511214/what-is-the-difference-between-x00a0-and-160))?

bpierre: Exactly. `&nbsp;` is an [HTML entity](https://developer.mozilla.org/en-US/docs/Glossary/Entity), while `\xa0` is a codepoint in JavaScript. I could have left it as a non breaking space character (` `), but it would look like a normal space in most editors, and could have been mistakingly replaced by a space (which would break its behavior, because spaces are ignored in HTML).

src/components/DropDown/DropDownItem.js
bpierre: Done, thanks!

bpierre:
> Do these components make sense when they don't have children? Just wondering if we should make children a required prop instead of adding the non breaking space.

I’m not sure about the idea of having UI components that break instead of providing a default state, even if it means having empty components.

It can be useful when prototyping, so people can drop a component in an app, see how it looks and position it, then assign a value later.

It could also be empty for other reasons, for example if the data needed to populate a DropDown is loaded asynchronously, and still want to have it, empty and disabled, during the request.

What do you think?

Note that the breaking space is only used because `line-height` is not available as a unit in CSS, and using a space is an easy way to get it.

sohkai:
> I’m not sure about the idea of having UI components that break instead of providing a default state, even if it means having empty components.

Makes sense. In dev mode, it might be nice to provide a warning for this behaviour, but that can be something we worry about later.

PR #65, opened 2017-12-11
Add Button.Anchor

sohkai:
Required for https://github.com/aragon/aragon/issues/66.

Maybe not the most "component"y (it'd be nice if people didn't have to care about `<a>`s and `Button`s exposed a linking API that didn't reinvent the wheel), but certainly the simplest.

Adds a demo to the gallery:
image

### DIFF
--- a/src/components/Button/Button.js
+++ b/src/components/Button/Button.js
@@ -116,8 +116,7 @@ type Props = {
   wide?: boolean,
 }
 
-// Flow declaration: see https://github.com/styled-components/styled-components/issues/570#issuecomment-332087358
-const StyledButton: ComponentType<Props> = getPublicUrl(styled.button`
+const StyledButton = styled.button`
   width: ${({ wide }) => (wide ? '100%' : 'auto')};
   padding: 10px 15px;
   white-space: nowrap;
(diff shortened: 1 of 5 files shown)


bpierre:
> Maybe not the most "component"y (it'd be nice if people didn't have to care about `<a>`s and `Button`s exposed a linking API that didn't reinvent the wheel), but certainly the simplest.

Agree. Ideally, I would like to not expose it at all, e.g. by switching to an `<a>` when `href` is set (that’s what I was doing on a previous project), but I agree it can be confusing to implicitly swap the HTML element.

sohkai:
Agreed, and I think our philosophy now is to avoid implicitly swapping the underlying HTML element.

src/components/Button/Button.js
```diff
@@ -152,7 +151,24 @@ const StyledButton: ComponentType<Props> = getPublicUrl(styled.button`
...
   }};
-`)
+`
+
+type ButtonComponentType = {
+  Anchor: ComponentType<Props>,
+}
+
+// Flow declaration: see https://github.com/styled-components/styled-components/issues/570#issuecomment-332087358
+// Currently throwing errors: https://github.com/styled-components/styled-components/issues/1350
```
sohkai: @bpierre Note the [`// $FlowFixMe`](https://github.com/aragon/aragon-ui/pull/65/files#diff-63c70416212465830688a856b92eb472R23) later on.

This is necessary for now, but hopefully they'll figure something out :).

Issue #58, opened 2017-12-06
Add Form


sohkai:
See most app side panels.

Location: `src/Form/Form.js`

sohkai:
After taking a look around, I don't think we need to provide our own Form component.

AFAIK, the design language we're leading with in our own apps has all forms inside the sidepanel (as forms are "actionable"). In these sidebars, there's no additional styling needed to signify there's a form; placing input fields is enough. All a form provides is a state container and some validation logic for these inputs.

I'm a fan of [formik](https://github.com/jaredpalmer/formik), in comparison to other form helpers like [react-form](https://react-form.js.org/) and [formsy](https://github.com/formsy/formsy-react/), because of its simplicity and familiar API. It's closely based on the native `form` and `input` components and allows you to easily [plug in your own custom `input`s](https://github.com/jaredpalmer/formik#field-) (rather than forcing you to adhere to their own API or make vendor-specific custom `inputs`). Our own input components, like those in #67, will work transparently with `formik`.

I don't see much point in re-exporting it from this package; we can just include `formik` (or even another form implementation) whenever an app needs a form in its sidepanel. For other app developers, we can add a suggestion in a "tips and tricks" wiki or app tutorial later.

Thoughts @bpierre?

bpierre:
Agree, `<form>` is behavioral and shouldn’t be provided by Aragon UI if possible.

> Our own input components, like those in #67, will work transparently with formik.

formik sounds good! I never tried it, but I remember having some difficulties to make styled components work with [react-formal](http://jquense.github.io/react-formal/) on a project.

If formik becomes our recommended solution to manage forms, maybe we should rename our `Field` component to something else (e.g. `FieldBlock`, `FieldSection`)?

sohkai:
> If formik becomes our recommended solution to manage forms, maybe we should rename our `Field` component to something else (e.g. `FieldBlock`, `FieldSection`)?

Maybe `LabelledField`? The other two sound too "style-derived".

sohkai:
Closing this, as we're not providing our own form element. Renaming can happen later, if we see the need to recommend formik.

PR #67, opened 2017-12-12
Add TextInput (and TextInput.Number)

### LINKED ISSUE #49: Add Text/Number Inputs
Number:

image
--------------
Text:

image
---------------
Disabled:

image

Location: `src/Form/Input.js`

sohkai:
Adds a basic, styled `<input />` box.

I've added the README as `src/Form/Input.md`, since I can see `src/Form` containing multiple components in the future (all the other input types, etc). In general, it might be interesting to have folders under `src/` be _groups_ holding actual components, with each component's documentation living as `<component name>.md`.

Fixes #49.


src/components/Form/Input.js
```diff
@@ -0,0 +1,31 @@
...
+  box-shadow: inset 0 1px 2px rgba(0, 0, 0, 0.06);
+  color: ${theme.textPrimary};
+  appearance: none;
+  &:focus {
+    outline: none;
+    border-color: ${theme.contentBorderActive};
+  }
+  &:read-only {
+    color: transparent;
+    text-shadow: 0 0 0 ${theme.textSecondary};
```
sohkai: Mostly for Firefox; Firefox does this weird thing where `readonly` inputs still show a flashing cursor.

An interesting fix for this is to make the text color transparent, and then use a text-shadow to display text given through `value`.

bpierre:
Shall we name this component `TextInput`? The HTML name is super generic, but we probably want something a bit more specific with a limited set of allowed `type` values.

For example, I think our implementation of `<input type=checkbox />` should be `CheckBox`, `<input type=range />` should be `<Slider />`, etc. What do you think?

> I can see src/Form containing multiple components in the future (all the other input types, etc). In general, it might be interesting to have folders under src/ be groups holding actual components

I agree, having components groups will make things clearer, great idea! About `Form/` specifically, what would be the criteria to put a component in `Form/`? In a classic HTML + HTTP app, some elements like `<input/>` have to be placed in a `<form>` so they can be submitted, but it’s not true for Web apps, where we can have these components inside or outside a form. For example, a `Button` can be used anywhere, a `CheckBox` could be used to filter a list, etc.

I checked what others are doing: [iOS uses the name “Controls”](https://developer.apple.com/ios/human-interface-guidelines/controls/) to group some of its interactive components, and [Android uses the name “Input Controls”](https://developer.android.com/guide/topics/ui/controls.html). What would you think of using a similar name for this group, instead of Form?

src/components/Form/Input.js
bpierre: Nice fix :-)

Do you if this is interpreted correctly by screen readers?

sohkai: That's a good question, a quick investigation doesn't bring up anything too conclusive but a test using voice over on OSX does read the read only value.

sohkai:
> Shall we name this component TextInput?

👍. I'll make it so `TextInput` defaults to `type="text"` and add `TextInput.Number` for `type="number"`.

> What would you think of using a similar name for this group, instead of Form?

Not sure I like `Controls/` since it's pretty ambiguous, but how does `Input/` (or `UserInput/`) sound? Originally I used `/Form` because I thought we were going to need a form element, but [we can scrap that now](https://github.com/aragon/aragon-ui/issues/58).

bpierre:
> I'll make it so TextInput defaults to type="text" and add TextInput.Number for type="number".

👍

> Not sure I like Controls/ since it's pretty ambiguous, but how does Field/ or Input/ (or UserInput/) sound? Originally I used /Form because I thought we were going to need a form element, but we can scrap that now.

`Field/` sounds form related, `UserInput/` sounds the most “correct”, but `Input/` is shorter. So `Input/` is my favorite :-)

PR #71, opened 2017-12-18
[DON'T MERGE] Add babel-plugin-flow-react-proptypes

sohkai:
Unfortunately, this wasn't as easy as just dropping in the plugin, because of [this issue](https://github.com/brigand/babel-plugin-flow-react-proptypes/issues/157)—[MenuItem's type export](https://github.com/aragon/aragon-ui/blob/master/src/components/Header/MenuItem.js#L7) breaks the bundled rollup build D:

This looks like it might be more of a rollup issue than the plugin's, and that it might have a long lead time to fix. It's not ideal, but one solution is to avoid local exports of types altogether given that we're not using them that much anyway.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -3,7 +3,8 @@
     ["inline-json-import", {}],
     ["styled-components", {
       "displayName": true
-    }]
+    }],
+    ["flow-react-proptypes"]
   ],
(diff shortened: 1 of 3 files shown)


sohkai:
Closing this since we're removing flow for now.

PR #72, opened 2017-12-18
Fix installation of markdown-it-anchor

sohkai:
I have no idea how this was working locally before... but now `markdown-it-anchor` is properly installed in the gallery.


bpierre:
> I have no idea how this was working locally before

That’s how the [modules resolution](https://nodejs.org/api/modules.html#modules_loading_from_node_modules_folders) work: if a module can’t be found, it keeps checking `node_modules` in every parent directory until the module is found or the root directory is reached.

PR #75, opened 2017-12-21
EmptyStateCard: generalize input components

sohkai:
Makes it easier to use `EmptyStateCard` in apps, who might want to supply their own button (e.g. an anchor) and a non-local static asset for the icon.

### DIFF
--- a/src/components/Card/EmptyStateCard.js
+++ b/src/components/Card/EmptyStateCard.js
@@ -1,4 +1,5 @@
 /* @flow */
+import type { Node } from 'react'
 import React from 'react'
 import styled from 'styled-components'
 import theme from '../../theme'
@@ -27,6 +28,7 @@ const StyledActionButton = styled(Button)`
 `
 
 type Props = {
+  actionButton: Node,
   actionText: string,
   icon: string,
   onActivate: mixed,
@@ -35,6 +37,7 @@ type Props = {
 }
 
 const DefaultProps = {
+  actionButton: StyledActionButton,
   title: 'Nothing here.',
 }
 
@@ -44,6 +47,7 @@ const EmptyStateCard = ({
   onActivate,
   text,
   title,
+  actionButton: ActionButton,
   ...props
 }: Props) => (
   <StyledCard {...props}>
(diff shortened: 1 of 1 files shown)


src/components/Card/EmptyStateCard.js
```diff
@@ -26,35 +27,38 @@ const StyledActionButton = styled(Button)`
...
   <StyledCard {...props}>
     <section>
-      <img src={icon} alt="" />
+      <Icon />
       <StyledHeading color={theme.accent} weight="bold" size="large">
         {title}
       </StyledHeading>
       <Text>{text}</Text>
-      <StyledActionButton mode="strong" onClick={onActivate}>
+      <ActionButton mode="strong" onClick={onActivate}>
```
bpierre: I wonder if it shouldn’t be a decision made by the card itself, since the button event is already managed by `onActivate`? The card could have a `href` parameter that would trigger this choice.

We also decide to go the other way and give more want to give more control, accepting an element rather than a component (`<Button mode="strong" onClick={…}>` rather than `Button`). I’m not sure I like this idea though, because I see `EmptyStateCard` as a high-level component that should stay simple (we have `Card` for flexibility).

sohkai: > We also decide to go the other way and give more want to give more control, accepting an element rather than a component (<Button mode="strong" onClick={…}> rather than Button)

Actually this should be the case, unless we're misunderstanding each other 😅. The component accepts an element, but `StyledActionButton` is the default element.

I decided against just an `href` since the added flexibility was so easy to get here and we might need it in the future for additional styling / etc.

bpierre: By “element” I mean the object returned by `React.createElement` when a component (`StyledActionButton`) is passed to it. For example, `<StyledActionButton />` or `React.createElement(StyledActionButton, null)`.

That would give more control to the parent by letting it pass the element, rather than creating it from a component. That way, other components could be used if needed (not only the ones implementing the `mode` property). But it also makes it less easy to use, so I don’t like this solution much.

> I decided against just an href since the added flexibility was so easy to get here and we might need it in the future for additional styling / etc.

OK, let’s keep it this way 👍

sohkai: Ahh gotcha. Not sure why I was confused, your first example was pretty clear 😅.

PR #78, opened 2018-01-10
Button: add secondary mode

sohkai:
screen shot 2018-01-09 at 7 44 12 pm

Required for the [voting app (see bottom button)](https://scene.zeplin.io/project/59a827960d4c4cb2274007f5/screen/59a82800d913ba794a4a5b4a).

### DIFF
--- a/src/components/Button/Button.js
+++ b/src/components/Button/Button.js
@@ -16,6 +16,7 @@ const {
   contentBackground,
   contentBorder,
   contentBorderActive,
+  secondaryBackground,
   textPrimary,
   textSecondary,
 } = theme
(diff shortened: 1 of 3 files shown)


src/components/Button/Button.js
```diff
@@ -55,6 +56,15 @@ const modeNormal = css`
   }
 `
 
+const modeSecondary = css`
+  ${plainButtonStyles};
+  background: ${secondaryBackground};
+  &:hover,
+  &:focus {
```
sohkai: @bpierre I removed the hover and focus states from the mode because they seemed out of place.

bpierre: I agree, we’ll add a focus ring later when working on it for all the components.

Also pressing the button feels right, even without the shadow to express the layer. We should probably try to remove it from the other buttons at some point, to align everything and make the hover state lighter (or try something more subtle, like a thin solid shadow).

PR #79, opened 2018-01-11
SidePanel: remove outline and selectable from close button

sohkai:
Avoids the outline when focused:
image

### DIFF
--- a/src/components/SidePanel/SidePanel.js
+++ b/src/components/SidePanel/SidePanel.js
@@ -5,7 +5,7 @@ import styled from 'styled-components'
 import { Motion, spring } from 'react-motion'
 import Text from '../Text/Text'
 import { lerp } from '../../math-utils'
-import { springConf } from '../../shared-styles'
+import { springConf, unselectable } from '../../shared-styles'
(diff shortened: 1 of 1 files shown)


src/components/SidePanel/SidePanel.js
```diff
@@ -62,9 +62,12 @@ const StyledPanelCloseButton = styled.button`
     cursor: pointer;
     background: none;
     border: 0;
+    outline: 0;
     &::-moz-focus-inner {
       border: 0;
     }
+
+    ${unselectable()};
```
sohkai: @bpierre I must've missed something in the docs, but I'm a little confused by how `styled-components` interpolates functions like this. I notice you've used `${unselectable};` by itself before, and that seems to be working fine. Does `styled-components` just execute the function with no args in that case?

bpierre: Yes exactly: if the value passed is a function, `styled-components` will execute it, passing the props as a parameter.

Executing like you did is probably a better idea, in case we want to add arguments later. Let’s do that :-)

PR #82, opened 2018-01-12
Refactor utils structure

### LINKED ISSUE #29: Create a better structure for utils
`shared-utils` and `math-utils` should live under a `utils/` directory.

It would also be nice to move some of the related code in `shared-utils` to their own file, e.g. `a utils/styled/breakpoints.js`.

monokh:
Fixes #29


src/shared-styles.js
```diff
@@ -1,126 +0,0 @@
...
-
-export const unselectable = () => `
-  -webkit-touch-callout: none;
-  -webkit-user-select: none;
-  -moz-user-select: none;
-  -ms-user-select: none;
-  user-select: none;
-`
-
-// TODO: deprecate the previous names
```
monokh: Thought i might as well do this while i'm here. Let me know if it impacts something externally

src/utils/styles/index.js
```diff
@@ -0,0 +1,8 @@
+/* @flow */
+export const unselectable = () => `
+  -webkit-touch-callout: none;
+  -webkit-user-select: none;
+  -moz-user-select: none;
+  -ms-user-select: none;
+  user-select: none;
+`
```
sohkai: It'd be nice if `index.js` here re-exported all the other style utils' exports. That way we could still just import everything we needed from `src/utils/styles` similar to before.

What do you think @bpierre?

monokh: I personally think this is a good idea so i've made the change. Let me know if you don't agree  🙂

src/components/DropDown/DropDown.js
```diff
@@ -4,8 +4,9 @@ import styled from 'styled-components'
 import { Motion, spring } from 'react-motion'
 import ClickOutHandler from 'react-onclickout'
 import theme from '../../theme'
-import { springConf, unselectable } from '../../shared-styles'
-import { lerp } from '../../math-utils'
+import { spring as springConf } from '../../utils/styles'
+import { unselectable } from '../../utils/styles'
```
sohkai: Could you collapse these imports into a single one from `utils/styles` (I think a few other files could also have this applied)?

monokh: Ah good catch :) fixed

monokh:
Hey @sohkai, is there anything missing from this one?

src/utils/styles/index.js
bpierre: > What do you think @bpierre?

Sorry I missed this one: agreed!

Issue #87, opened 2018-01-17
Add ability to detect required input in Field


sohkai:
![image](https://user-images.githubusercontent.com/4166642/33648461-fa37dbc0-da27-11e7-8f36-98036749c728.png)

Should show a highlighted `*` if the child has `required`.

kdichev:
<img width="348" alt="screen shot 2018-01-19 at 15 44 03" src="https://user-images.githubusercontent.com/12297753/35155989-b808eb74-fd2f-11e7-8df8-ed13977c96d9.png">

Is this what's needed if yes I can pull request the feature
@sohkai

sohkai:
@kdichev Yup, you've got it :).

kdichev:
@sohkai have pull requested the feature :)

PR #88, opened 2018-01-19
add required asterix in Field component

kdichev:
I am not sure about how to show the new feature in gallery but you guys can give me a hint.
I am also a bit uneasy with the float rule that I wrote, open to suggestions on how to make this better.

resolves: #87

### DIFF
--- a/gallery/src/pages/PageField.js
+++ b/gallery/src/pages/PageField.js
@@ -18,7 +18,7 @@ const PageField = ({ title }) => (
       <Container>
         <div>
           <Field label="Enter name here:">
-            <TextInput required/>
+            <TextInput required />
           </Field>
         </div>
       </Container>
--- a/gallery/src/pages/PageField.js
+++ b/gallery/src/pages/PageField.js
@@ -11,7 +11,7 @@ const PageField = ({ title }) => (
       <Container>
         <div>
           <Field label="Enter name here:">
-            <TextInput />
+            <TextInput required/>
           </Field>
         </div>
       </Container>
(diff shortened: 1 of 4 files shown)


gallery/src/pages/PageField.js
```diff
@@ -11,7 +11,7 @@ const PageField = ({ title }) => (
       <Container>
         <div>
           <Field label="Enter name here:">
-            <TextInput />
+            <TextInput required/>
```
bpierre: Could you add this as a second example? So we can see the `Field` when required and not required.

kdichev:
@bpierre Viable requests 👍 I will add the changes later today 🥇

src/components/Input/TextInput.md
```diff
@@ -22,6 +22,21 @@ const App = () => (
 
 See [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#Form_<input>_types).
 
+### `required`
```
sohkai: I think it's OK to leave this out since `required` is covered by the catch-all at the bottom (native `<input />` props).

gallery/src/pages/PageField.js
```diff
@@ -15,6 +15,13 @@ const PageField = ({ title }) => (
           </Field>
         </div>
       </Container>
+      <Container>
+        <div>
+          <Field label="Enter name here:">
+            <TextInput required/>
```
sohkai: Needs a space before `/>`

bpierre:
Thanks again @kdichev! 💯💯💯

PR #89, opened 2018-01-21
New icons: Check, Cross, Time

bpierre:
image


src/icons/components/Time.js
```diff
@@ -0,0 +1,14 @@
+import React from "react"
+
+const Time = props => (
+  <svg width={13} height={13} viewBox="0 0 13 13" {...props}>
+    <path
+      d="M6.5 11.76c.8 0 1.535-.2 2.205-.6.66-.39 1.185-.92 1.575-1.59.39-.67.585-1.405.585-2.205S10.67 5.83 10.28 5.16a4.403 4.403 0 0 0-1.575-1.575A4.305 4.305 0 0 0 6.5 3c-.8 0-1.535.195-2.205.585-.66.39-1.185.915-1.575 1.575a4.305 4.305 0 0 0-.585 2.205c0 .8.195 1.535.585 2.205.39.67.915 1.2 1.575 1.59.67.4 1.405.6 2.205.6zm0-10.02c1.03 0 1.98.255 2.85.765.85.49 1.52 1.16 2.01 2.01.51.87.765 1.82.765 2.85s-.255 1.98-.765 2.85c-.49.85-1.16 1.52-2.01 2.01-.87.51-1.82.765-2.85.765s-1.98-.255-2.85-.765a5.386 5.386 0 0 1-2.01-2.01 5.535 5.535 0 0 1-.765-2.85c0-1.03.255-1.98.765-2.85.49-.85 1.16-1.52 2.01-2.01.87-.51 1.82-.765 2.85-.765zm.33 2.52v3.315L9.32 9.06l-.51.765L5.885 8.01V4.26h.945zM3.95 1.395l-2.895 2.37L.26 2.82 3.125.45l.825.945zm8.79 1.425l-.795.945-2.895-2.46.825-.945 2.865 2.46z"
+      fill="#6D777B"
+      fillRule="evenodd"
+      opacity={0.75}
```
sohkai: Wondering if this should be a passable prop? Or if I wanted to change the opacity, I could pass in `style` or a css class to the entire svg?

bpierre: Ah I didn’t see this one, good catch! I think it should be converted to a plain color, and we can change the opacity using `style` if needed.

A prop that I would like to have would be `color`: all our icons are monochromatic so far, and it can be configured with svgr. What do you think?

sohkai: > `color` prop

That would be super cool!

It looks like with `currentColor`, you can get the svg to inherit from the text colour. Maybe we can use that and apply a default text colour in all our svgs?

bpierre: Yes I agree 👍

In specific cases like the checkbox and the error cross, we could simply have `theme.positive` / `theme.negative` as a default color instead of `currentColor`.

PR #90, opened 2018-01-21
New component: Countdown

### LINKED ISSUE #55: Add TimeText
image
-----------
image

Location: `src/Text/TimeText.js`

bpierre:
Requires https://github.com/aragon/aragon-ui/pull/89, closes #55.

Displays a countdown:

image

### DIFF
--- a/src/components/Countdown/Countdown.js
+++ b/src/components/Countdown/Countdown.js
@@ -1,0 +1,108 @@
+import React from 'react'
+import styled from 'styled-components'
+import PropTypes from 'prop-types'
+import IconTime from '../../icons/components/Time'
+import { difference } from '../../date-utils'
+import { unselectable } from '../../shared-styles'
+import theme from '../../theme'
+
+const formatUnit = v => String(v).padStart(2, '0')
+
+class Countdown extends React.Component {
+  componentDidMount() {
+    this.startTimer()
+  }
+  componentWillUnmount() {
+    clearTimeout(this.timeout)
+  }
+  startTimer = () => {
+    this.timeout = setTimeout(() => {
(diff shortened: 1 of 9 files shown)


src/components/Countdown/Countdown.js
```diff
@@ -0,0 +1,108 @@
...
+`
+
+const Part = styled.span`
+  font-size: 15px;
+  font-weight: 600;
+  color: ${theme.textPrimary};
+  & + & {
+    margin-left: 7px;
+  }
+  &:nth-child(n + 2) + & {
```
sohkai: What do you think of making the `:` its own styled component instead? In this case I think I prefer the explicitness rather than a pseudoelement.

src/components/Countdown/Countdown.js
```diff
@@ -0,0 +1,108 @@
...
+  }
+  render() {
+    const { end } = this.props
+    const { days, hours, minutes, seconds, totalInSeconds } = difference(
+      end,
+      new Date()
+    )
+
+    return (
+      <Main dateTime="">
```
sohkai: Should `dateTime` be set to `end.toISOString()`?

I think it makes sense to set `end` as the timestamp, since that's what we're counting down to, but I'm not super familiar with the semantics of `<time>`.

bpierre: Thanks, I forgot to add this. I agree, the countdown represents the `end` date.

src/components/Countdown/Countdown.js
bpierre: Agree, it will make it more readable. 👌

PR #93, opened 2018-01-21
Add SidePanelSeparator and SidePanelSplit

bpierre:
These components can be used inside of `<SidePanel />`.

SidePanelSplit creates a vertical line between its two children:

image

SidePanelSeparator creates an horizontal line:

image


src/components/SidePanel/SidePanelSplit.js
```diff
@@ -0,0 +1,29 @@
...
+const SidePanelSplit = ({ children, ...props }) => (
+  <Main {...props}>
+    <Part>{children[0]}</Part>
+    <Part>{children[1]}</Part>
+  </Main>
+)
+
+const Main = styled.div`
+  display: flex;
+  width: calc(100% + ${HORIZONTAL_PADDING * 2}px);
```
sohkai: Just wondering if using `PANEL_WIDTH` would be easier here (and in `SidePanelSeparator`). Maybe then we wouldn't have to adjust for `HORIZONTAL_PADDING` everywhere?

bpierre: Yes good idea 👍

Issue #10, opened 2017-11-20
Gallery: add an iframe mode for demos


bpierre:
Some full-width components are now responsive and should be rendered in an iframe to make the media queries work properly.

status-open-bounty:
![Contract deploying](https://openbounty.status.im/img/deploying_contract.png)
Transaction: [0x61ca6f1dbb747fc4f8f18b14a271dd9eaafebc88df66c4eda67f18309da2173d](https://etherscan.io/tx/0x61ca6f1dbb747fc4f8f18b14a271dd9eaafebc88df66c4eda67f18309da2173d)

kdichev:
Can I get a bit more information about this issue? Which components are you refering with "some" in particular.

sohkai:
The best case would be to render every demo in an resizable iframe, kind of like what they have at Shopify's Polaris (see e.g. their [Collapsible](https://polaris.shopify.com/components/behavior/collapsible#navigation) component).

Specifically, we have the `SidePanel` component as well as a few components using `<Breakpoint>` that should be rendered in an iframe.

kdichev:
@sohkai I suspect that we would want to build our own Aragon iframe component and not depend on an external dependency component?

kdichev:
I did a simple implementation using an react component `react-frame-component`.

sohkai:
Awesome, thanks for submitting the PR! We haven't had a lot of time lately to check on PRs, but we'll get to them soon :).

I had no idea portals could be used cross-context like this, that's super neat!

PR #95, opened 2018-01-23
Add Frame and Resizable Component

kdichev:
Fixes: #10 

Frame:
- displays its ```children``` components in an ```iframe``` tag
- Uses React Portals
- No external dependencies
- works with ```styled-components```

Resizable:
- can be resized using mouse movement events
- can get its width set in order to display different device sizes


kdichev:
@sohkai when you have the time you can test this branch. I have resolved all issues with initial implementation, so check it out.

Resize functionality is a css feature, so a React ```ResizableContainer``` might not be a bad idea, let me know what you think.

kdichev:
image

For a quick ui reference. Feel free to comment on that 👍

gallery/src/components/Page/Frame.js
```diff
@@ -0,0 +1,53 @@
...
+    }
+  }
+
+  render() {
+    const { children,  ...rest } = this.props
+    const { iframeHead, iframeBody, iframeHeight } = this.state
+    const { instance: { tags } } = StyleSheet // or passed props
+    const styleTags = tags.map((tag, index) => (
+      <StyleTag el={tag.el} key={index} />
+    ))
```
sohkai: I assume this is because we're using portals, and hence `styled-components` still renders its stylesheets in the parent rather than the child iframe?

Maybe we can add a comment here describing why we need to copy them over :).

kdichev: Yes this is correct. Generated styles from ```styled-components``` are captured and from the tags a ```StyleTag``` component is created for each then teleported to the iframe head.

Portals can only teleport react components hence I needed to create components from the style tags. I am trying to find an api from ```styled-components``` that might make this look cleaner.

They have an active pull request ongoing that adds changes so you can specify the element in which styles are injected.

the way they are doing it now is creating style tags and appending them to ```document.head```.

bpierre:
Awesome work! 🦅 💫 💯 🥇

I added a few changes:

- Visual style alignment.
- Firefox compatibility (had to refactor a bit the Frame component for that).
- Prevent the iframe to swallow mousemove events on resize.
- Add the possibility to have a fixed height on a container (e.g. for the SidePanel demo).
- Adapt the demos to the new container.
- Add an `opaque` prop on `Page.Demo` to alternate between a transparent / white background, depending on the demo.

image

@sohkai @kdichev What do you guys think?

kdichev:
@bpierre Great addition!!!
Going for styles in body makes everything easier  🍾. I started exactly that same way, dunno why I went for the head that much. 😸 
Looks good to me! I will test tomorrow and check out the changes in depth, to learn a thing or two 😊

bpierre:
@kdichev @sohkai Merging this now so we can release a new version of Aragon UI soon! Awesome work 💖

PR #99, opened 2018-01-31
Add the ContextMenu component

bpierre:
image


src/components/ContextMenu/ContextMenu.js
```diff
@@ -0,0 +1,124 @@
...
+            <Main
+              opened={opened}
+              style={{
+                boxShadow: `0 4px 4px rgba(0, 0, 0, ${openProgress * 0.03})`,
+              }}
+            >
+              <BaseButton onClick={this.handleBaseButtonClick} opened={opened}>
+                <img src={opened ? ellipsisActive : ellipsis} alt="" />
+                <img
+                  src={arrow}
```
sohkai: Does this need `prefixUrl`?

src/components/ContextMenu/ContextMenu.js
```diff
@@ -0,0 +1,124 @@
...
+  position: absolute;
+  top: ${BASE_HEIGHT - 1}px;
+  right: 0;
+  padding: 10px 0;
+  background: ${theme.contentBackground};
+  border: 1px solid ${theme.contentBorder};
+  border-radius: 3px 0 3px 3px;
+`
+
+export default ContextMenu
```
sohkai: Does this need `getPublicUrl`?

src/components/ContextMenu/ContextMenu.js
```diff
@@ -0,0 +1,124 @@
...
+    display: ${({ opened }) => (opened ? 'block' : 'none')};
+    content: '';
+    position: absolute;
+    bottom: -1px;
+    right: -1px;
+    width: 1px;
+    height: 2px;
+    background: ${theme.contentBorder};
+  }
+  img:first-child {
```
sohkai: I could see an `EllipsisIcon` component here too, to style this ``:

```js
const EllipsisIcon = styled.attrs({
  src: ({ opened }) => { ellipsisActive : ellipsis })
})`
  margin-right: 5px;
`
```

src/components/ContextMenu/ContextMenuItem.js
```diff
@@ -0,0 +1,21 @@
...
+const ContextMenuItem = styled.div`
+  display: flex;
+  align-items: center;
+  padding: 5px 20px;
+  cursor: pointer;
+  ${unselectable()};
+  &:active {
+    background: ${theme.contentBackgroundActive};
+  }
+  & > svg {
```
sohkai: Hmm, I think this rule makes more sense being applied on the children being passed into ContextMenuItem (e.g. what if I wanted to pass an `` as the arrow icon instead)?

src/components/ContextMenu/assets/ellipsis-active.svg
```diff
@@ -0,0 +1 @@
+<svg width="15" height="4" viewBox="0 0 15 4" xmlns="http://www.w3.org/2000/svg"><path d="M7.5 3.213a1.42 1.42 0 0 1-.974-.37c-.278-.248-.418-.588-.418-1.021 0-.384.135-.71.404-.979S7.11.439 7.5.439s.722.135.997.404c.276.27.413.595.413.979 0 .439-.142.78-.427 1.025a1.465 1.465 0 0 1-.983.366zm-5.327 0c-.371 0-.694-.122-.97-.366C.928 2.603.791 2.26.791 1.822c0-.39.133-.718.398-.984.266-.266.594-.399.984-.399s.722.135.997.404c.275.27.413.595.413.979 0 .439-.142.78-.427 1.025a1.465 1.465 0 0 1-.983.366zm10.654 0c-.365 0-.688-.123-.97-.37-.28-.248-.421-.588-.421-1.021 0-.384.134-.71.403-.979.27-.269.598-.404.988-.404s.722.135.997.404c.276.27.413.595.413.979 0 .433-.14.773-.422 1.02a1.45 1.45 0 0 1-.988.371z" fill="#21AAE7" fill-rule="evenodd"/></svg>
```
sohkai: Is the only difference between these two ellipses the colour? Wondering if it'd be better to inline it and then change the colour with css.

src/components/ContextMenu/ContextMenu.js
```diff
@@ -0,0 +1,123 @@
...
+import ClickOutHandler from 'react-onclickout'
+import theme from '../../theme'
+import { springConf, unselectable } from '../../shared-styles'
+
+import ellipsis from './assets/ellipsis.svg'
+import ellipsisActive from './assets/ellipsis-active.svg'
+import arrow from './assets/arrow.svg'
+
+export const BASE_WIDTH = 46
+export const BASE_HEIGHT = 32
```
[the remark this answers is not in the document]
bpierre: Agree, it could be useful to provide an access to these.

src/components/ContextMenu/ContextMenu.js
bpierre: Indeed! Thanks 😄

PR #101, opened 2018-02-04
Add Badge

### LINKED ISSUE #56: Add BadgeText
image
---------------
Circular border-radius:

image
----------------
Different bg:

image
---------------
Header:

image

Location: `src/Badge/BadgeText.js`

Could probably be integrated with the current `BadgeNumber`?

bpierre:
Badge replaces `BadgeNumber` and support other shapes.

image

Note: `BadgeNumber` is still present and uses `Badge`. It will be removed after migrating Aragon and apps to use `Badge`.

Fixes #56


src/components/Badge/Badge.Info.js
```diff
@@ -0,0 +1,28 @@
...
+  </Badge>
+)
+
+Info.defaultProps = {
+  count: -1,
+  small: false,
+}
+
+Info.propTypes = {
+  count: PropTypes.number,
```
sohkai: Just thinking, if we only ever plan to use strings or numbers for `Badge.Info`s, maybe we could change `count` to `text` (or something) and then use `typeof text` to know when to format it?

Same for `Badge.Notification` below.

src/utils/format.js
```diff
@@ -0,0 +1,10 @@
+export const formatCount = (count = -1, max = 99) => {
+  count = parseInt(count, 10)
+  if (count < 1) {
+    return '0'
+  }
+  if (count > max) {
+    return `${parseInt(max, 10)}`
```
sohkai: Should this add a `+` to the end too?

src/components/Badge/Badge.js
```diff
@@ -0,0 +1,80 @@
...
+import BadgeApp from './Badge.App'
+
+const shapeStyles = shape => {
+  if (shape === 'disc') {
+    return css`
+      overflow: hidden;
+      padding-top: 2px;
+      letter-spacing: -0.5px;
+      justify-content: center;
+      align-items: center;
```
sohkai: Do we need `justify-content` and `align-items` since these components aren't in flexbox? (And for `smalldisc` too)

src/theme/Aragon.oco
```diff
@@ -83,6 +83,14 @@ Aragon UI:
   positiveText: =White.White
   negative: =Red.Salmon Red
   negativeText: =White.White
+  badgeAppBackground: =Purple.Lavender
```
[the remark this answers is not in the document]
bpierre: It’s in the spec, but the official apps don’t support it yet:

> Please note that while the export will honor sub groups in your artboard and create nested OCO palettes that way, both the Companion App and the color picker currently don't support palettes nested deeper than one level, so it's better to keep the structure flat in your "swatchboard".

https://github.com/opencolor-tools/sketch-opencolor/blob/master/MANUAL.md#export-palette

src/utils/format.js
bpierre: It was the original plan but it doesn’t fit. I left it as “99” for now, and will ask @jounih about a possible solution.

src/components/Badge/Badge.js
bpierre: They are! :-) https://github.com/aragon/aragon-ui/pull/101/files#diff-b62a88a1d42347ac37d53ca997bd1c96R55

src/components/Badge/Badge.Info.js
bpierre: Yes it could be a nice feature! And people could still use `children` to disable it if they want to.

sohkai: OK, so any children would override whatever's given in the `text` prop? Sounds good to me.

src/components/Badge/Badge.js
sohkai: Ohh good catch, read it as `inline-display`

src/components/Badge/Badge.Info.js
bpierre: Exactly, and `text` could be named `label` so that it doesn’t sound “string”.

sohkai: See 880186c; I also renamed `formatCount()` to `formatNumber()` and added a parameter for replacement text to make it more generic.

bpierre: Cool! I wonder if `formatNumber()` is not too generic for this?

sohkai: Yeah, thinking about it, renaming it to `formatIntegerRange()` or `formatMaxInteger()` might make more sense.

bpierre: 💯

sohkai: See 60eb09d. Everything looks good to me, so merge at will :).

bpierre: Awesome thanks ✨💯

PR #107, opened 2018-02-04
Linting fixes

sohkai:


sohkai:
Doesn't look like `.prettierignore` is working with eslint, so I've added `/* eslint-disable prettier/prettier */` to the index files.

PR #117, opened 2018-02-16
Add RadioGroup

sohkai:
image

The spacing and colours are eyeballed so feel free to adjust them @bpierre! The designs seem to have slightly different colours than what we currently make available in the theme, but I found close enough approximates that it might not be a big deal.

Currently the only animations are fade-ins for the border colours and input's checked circle. Making the circle grow is problematic because:

1. The switch from the non-checked darker background to white is initially jarring if there's a lot of white; and
1. Browsers don't like to center position the circle except for certain "even" dimensions.


src/components/Radio/RadioItem.js
```diff
@@ -0,0 +1,134 @@
...
+  cursor: pointer;
+
+  &:focus,
+  &:hover,
+  input:focus ~ &,
+  input:hover ~ & {
+    border-color: rgba(33, 183, 196, 0.35);
+  }
+  input:checked ~ & {
+    border-color: rgba(33, 183, 196, 0.7);
```
sohkai: This is `Sea.Light Sea` but I don't know how to turn it into a value with alpha.

src/components/Radio/RadioGroup.js
```diff
@@ -0,0 +1,78 @@
...
+            title={title}
+          />
+        ))}
+      </Group>
+    )
+  }
+}
+
+const Group = styled.div.attrs({
+  role: 'radiogroup',
```
sohkai: I only added this role, since I think the `<input type="radio" />`s should already have the right roles implicitly?

bpierre:
I wonder if we should keep the name `RadioGroup` for grouping checkboxes logically, and separate it from any style? Similar to what Android, iOS and XUL do.

Example with simple horizontal group ( ● Foo ○ Bar ○ Baz ) :

```jsx
<RadioGroup selected={selectedIndex} onChange={handleChange}>
  {['Foo', 'Bar', 'Baz'].map(label => (
    <label>
      <RadioButton /> {label}
    </label>
  ))}
</RadioGroup>
```

But I’m not sure it’s much needed with React, we can let people use controlled components instead:

```jsx
<div>
  {['Foo', 'Bar', 'Baz'].map((label, i) => (
    <label>
      <RadioButton
        selected={selectedIndex === i}
        onChange={() => handleChange(i)}
      />
      {label}
    </label>
  ))}
</div>
```

To reproduce what we have now (a styled radio box for the SidePanel), we could export something like `RadioListItem`:

```jsx
<div>
    <RadioListItem title="A choice" description="A description of the choice" selected />
    <RadioListItem title="Another choice" description="A description of the choice" />
    <RadioListItem title="A third choice" description="A description of the choice" />
</div>
```

Or we could use a “super component”, e.g. `RadioList`, exactly in the same way `RadioGroup` is used at the moment:

https://github.com/aragon/aragon-ui/blob/a1aa910fdc2a84554b0b8a7ee836233e8c43cbda/gallery/src/pages/PageRadioGroup.js#L22-L42

What do you think?

src/components/Radio/RadioItem.js
sohkai: @bpierre Ahh, should I use [`onecolor`](https://github.com/One-com/one-color) (`color(<color>).alpha()`)?

bpierre: I think it would be nice to (lazy) wrap the colors with a color utility when one of its method is used. onecolor looks good, let’s use it if it works well with styled-components!

sohkai:
@bpierre I've modified the structure so that there's a separate `<RadioInput />` and a fairly dumb `<RadioGroup />` container that `<RadioList />` uses to render the original UI.

Probably easiest to re-review (feel free to ignore the gallery docs).

src/components/Input/RadioInput.js
```diff
@@ -0,0 +1,93 @@
...
+    inline: PropTypes.bool,
+    onChange: PropTypes.func,
+    onSelect: PropTypes.func,
+  }
+  static defaultProps = {
+    // By default, prevent the default change event from bubbling up
+    onChange: event => {
+      event.stopPropagation()
+    },
+    onSelect: () => {},
```
sohkai: Note that `onChange()` will propagate the `change` event without the index (as a "normal" `onChange()`), while `onSelect()` will only propagate the index.

I've left both options in `<RadioInput />` in case it's useful in other cases than the `<RadioListItem />` (although it could be moved here).

sohkai: Slept on this and decided this was a bad idea. It's been moved up to `<RadioListItem />`.

src/components/Radio/RadioItem.js
sohkai: onecolor works pretty well; using `.css()` / `.cssa()` to explicitly get a css-compatible string solves the `.toString()` problem :).

src/components/Input/RadioInput.md
```diff
@@ -0,0 +1,32 @@
...
+    <RadioInput checked index={0} value="first" onChange={onChange} />
+    <RadioInput index={1} value="second" onChange={onChange} />
+    <RadioInput index={2} value="third" onChange={onChange} />
+  </RadioGroup>
+)
+```
+
+## Properties
+
+### `inline`
```
bpierre: Shouldn’t we always display it inline?

src/components/Radio/RadioGroup.md
```diff
@@ -0,0 +1,134 @@
...
+
+## Properties
+
+### `className`
+
+- Type: `String`
+
+Sets the class name of the `radiogroup` container.
+
+### `*`
```
bpierre: Shouldn’t we have a dedicated prop to do that? e.g. `childrenProps`

sohkai: Mmmm I was thinking about this but I thought it was more awkward. `RadioGroup` is really simple and any props added directly to it can represent children's.

src/components/Input/RadioInput.md
sohkai: Good point!

src/components/Input/RadioInput.js
```diff
@@ -0,0 +1,60 @@
...
+const radioDimmed = css`
+  ${radioActive};
+  &:after {
+    opacity: 0.5;
+    transition: none;
+  }
+`
+
+// Styled component
+const RadioInput = styled.input.attrs({
```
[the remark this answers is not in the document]
sohkai: I agree, I wanted naming consistency with `*Input` but that was misguided.

PR #118, opened 2018-02-16
Add SafeLink

sohkai:
See https://mathiasbynens.github.io/rel-noopener.

Changes `Button.Anchor` to also use `SafeLink`.


src/components/Link/SafeLink.js
```diff
@@ -0,0 +1,8 @@
+import styled from 'styled-components'
+
+const SafeLink = styled.a.attrs({
+  // See https://mathiasbynens.github.io/rel-noopener
+  rel: 'noopener noreferrer',
+})``
```
sohkai: This is a really funny way of doing it, but it looks like `styled-components` has a whitelist of DOM attributes they pass down, which is good. Passing down all the props gives React errors when we use things like `publicUrl`.

PR #126, opened 2018-02-28
Icon exports: remove the .js in the import paths

bpierre:
Depends on https://github.com/aragon/aragon-ui/pull/125


src/icons/index.js
```diff
@@ -1,18 +1,19 @@
...
-export { default as IconIdentity } from './components/Identity.js'
-export { default as IconNotifications } from './components/Notifications.js'
-export { default as IconPermissions } from './components/Permissions.js'
-export { default as IconSettings } from './components/Settings.js'
-export { default as IconShare } from './components/Share.js'
-export { default as IconTime } from './components/Time.js'
-export { default as IconTokens } from './components/Tokens.js'
-export { default as IconVoting } from './components/Voting.js'
-export { default as IconWallet } from './components/Wallet.js'
+export { default as IconAdd } from './components/Add'
```
sohkai: It looks like this is breaking CI (linter complaining that this import isn't resovable). I think a rebase should fix this?

bpierre: Ah yes, it’s because `Add` doesn’t exist in master. I’ll rebase when https://github.com/aragon/aragon-ui/pull/125 is merged.

PR #128, opened 2018-03-04
Add redraw HOC

sohkai:
Splits out the redrawing logic in `<Countdown />` into a reusable HOC.


src/index.js
```diff
@@ -4,6 +4,7 @@ export * from './icons'
 export { default as theme, themeDark, brand, colors } from './theme'
 export { font, grid, spring, breakpoint, BreakPoint, unselectable } from './utils/styles'
 export { formatIntegerRange } from './utils/format'
+export { default as redraw } from './components/HOC/redraw'
```
sohkai: I'm not sure where's the best place to put these types of HOCs, so I just left it here. Also not sure what's the best way to re-export it, so I'm re-exporting it directly.

PR #129, opened 2018-03-04
Add observer HOC for connecting to RxJS

sohkai:
Adds a convenience HOC for connecting a component to an RxJS observable.

Passes down a transformation of the observable's values as props.

### DIFF
--- a/src/rxjs/observe.js
+++ b/src/rxjs/observe.js
@@ -1,0 +1,62 @@
+import React from 'react'
+
+// Higher-order component for convenient subscriptions to RxJS observables
+const observe = (Component, observe, initialState = {}) => {
+  return class extends React.Component {
+    static displayName = `Observe(${Component.displayName ||
+      Component.name ||
+      'Component'})`
+    static propTypes = {
+      observable: ({ observable }, _, componentName) => {
+        if (observable && typeof observable.subscribe !== 'function') {
+          throw new Error(
+            `Invalid prop \`observable\` supplied to \`${componentName}\` ` +
+              '(wrapped by `observe()`). ' +
+              '`observable` must be an RxJS Observable-like object. ' +
+              `Given ${observable} instead.`
+          )
+        }
+      },
+    }
+    state = initialState
+    componentDidMount() {
+      this.subscribe(this.props.observable)
+    }
+    componentWillReceiveProps({ observable: nextObservable }) {
+      const { observable } = this.props
+      // If a new observable gets passed in, unsubscribe from the old and subscribe to the new
+      if (nextObservable !== observable) {
+        this.unsubscribe()
+        this.subscribe(nextObservable)
+      }
+    }
+    componentWillUnmount() {
+      this.unsubscribe()
+    }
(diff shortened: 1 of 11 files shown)


src/rxjs/observe.js
```diff
@@ -0,0 +1,62 @@
...
+      'Component'})`
+    static propTypes = {
+      observable: ({ observable }, _, componentName) => {
+        if (observable && typeof observable.subscribe !== 'function') {
+          throw new Error(
+            `Invalid prop \`observable\` supplied to \`${componentName}\` ` +
+              '(wrapped by `observe()`). ' +
+              '`observable` must be an RxJS Observable-like object. ' +
+              `Given ${observable} instead.`
+          )
```
sohkai: Originally I wanted to make RxJS an `optionalDependency` so I could import `rxjs.Observable` and do an `instanceof` check here, but setting that up was non-trivial.

Doing so would probably require creating two bundles, and the best DX would be something like `import { observable } from '@aragon/ui/rxjs'`. Not sure how lodash or other projects do it (non-obvious from a few glances), but I didn't get anything working after a few attempts and gave up.

This should be OK until we get the time to figure out bundling.

src/rxjs/observe.js
```diff
@@ -0,0 +1,62 @@
...
+        }
+      },
+    }
+    state = initialState
```
sohkai: The `initialState` is not super useful, since the same effect can be produced by `defaultProps` in the underlying component, but there might be cases where it's needed (e.g. a component getting wrapped multiple times). I also think it's more clear as `initialState` rather than `defaultProps`.

src/rxjs/observe.js
```diff
@@ -0,0 +1,62 @@
...
+      this.subscription && this.subscription.unsubscribe()
+    }
+    render() {
+      const { ...props } = this.props
+      // Don't pass down the given observable
+      delete props.observable
+
+      return (
+        <Component
+          subscription={this.subscription}
```
sohkai: This is passing down an instance variable, which is probably frowned upon.

I did it here because `this.subscription` is always tied to `props.observable` so it's unnecessary to put it in state.

bpierre: I wonder if there is any significant overhead resulting of using `setState()`, as this component is already using a state?

Or we could add a comment to explain that it’s important to keep the `this.subscription` assignment synchronous, or the update will fail?

onbjerg:
Looks pretty good, we could probably work from this and make a more Aragon.js-specific "connector" later on 🤩

src/rxjs/observe.js
sohkai: On second thought, just using state is a better idea. There's no good reason not to, since it doesn't cause a re-render, and it's what I've done in the `cache` HOC as well.

sohkai:
I've changed the subscription logic to be in `componentDidMount()` rather than `componentWillMount()` because this also conforms better to react's best-practices. I remember trying this earlier and something broke, but this should is something that should be fixed in the wrapped component rather than the HOC.

PR #130, opened 2018-03-06
Button (strong): support the disabled attribute

bpierre:


src/components/Button/Button.js
```diff
@@ -41,7 +43,7 @@ const plainButtonStyles = css`
   }
   &:hover,
   &:focus {
-    box-shadow: 0 1px 1px rgba(0, 0, 0, 0.2);
+    ${({ disabled }) => (disabled ? 'none' : '0 1px 1px rgba(0, 0, 0, 0.2)')};
```
sohkai: Does this still need `box-shadow`?

src/components/Button/Button.js
```diff
@@ -70,8 +72,14 @@ const modeSecondary = css`
...
   ${plainButtonStyles};
-  color: ${gradientText};
-  background-image: linear-gradient(130deg, ${gradientStart}, ${gradientEnd});
+  color: ${({ disabled }) => (disabled ? disabledText : gradientText)};
+  background-image: ${({ disabled }) =>
+    disabled
+      ? 'none'
+      : `linear-gradient(130deg, ${gradientStart}, ${gradientEnd})`};
+  background-color: ${({ disabled }) =>
+    disabled ? disabledColor : 'transparent'};
```
sohkai: Wondering if this would look better if we condensed all of the `disabled` switches into a single block, e.g.

```js
css`
  ${({ disabled }) => disabled ? css`
    [disabled styles]
  ` : css`
    [normal styles]
  `
`
```

PR #131, opened 2018-03-06
Add font styles needed for the onboarding

bpierre:


src/utils/styles/font.js
```diff
@@ -12,6 +12,7 @@ const FONT_SIZES: {
   large: '16px',
   xlarge: '20px',
   xxlarge: '22px',
+  great: '37px',
```
sohkai: Haha, no `xxxxxlarge`?

bpierre: I couldn’t 😂

We should spend some time to tidy our text styles with @jounih after 0.5 is released.

PR #135, opened 2018-03-07
License should be MIT, fix 134

gasolin:
According to https://github.com/aragon/aragon-ui/blob/master/LICENSE


bpierre:
Thanks @gasolin! 👌

PR #158, opened 2018-04-26
Move “providers” components to src/providers

bpierre:
Requires https://github.com/aragon/aragon-ui/pull/160 to be merged before merging into `master`.

This includes high-order components and render props components.

Also reorganize the exports so that each directory in src/ re-exports its own names.

Each provider is available as a component expecting a render function as its `children`. They also have an `hocWrap` method that acts as an HOC wrapper (this is very useful to wrap styled-components).

The only exception to this is `observe`, which is only an HOC wrapper at the moment and will be migrated later.

In the case of PublicUrl, it is now using the new [React Context](https://reactjs.org/docs/context.html) API.


gallery/src/pages/PageRedraw.js
```diff
@@ -0,0 +1,66 @@
...
+import Container from '../components/Page/DemoContainer'
+
+const redrawDemo = interval => {
+  return class extends React.Component {
+    render() {
+      return (
+        <Redraw interval={interval}>
+          {() => {
+            const color = Math.floor(Math.random() * 0xffff).toString(16)
+            if (!this.count) this.count = 0
```
sohkai: Small nitpick: we could move this up as a closure variable before the class component since we're cheating the state anyway ;) (also in `PageRedrawFromDate`)

src/components/CircleGraph/CircleGraph.js
```diff
@@ -48,7 +48,7 @@ const CircleGraph = ({ value, label, size }) => {
 CircleGraph.propTypes = {
   value: PropTypes.number,
   size: PropTypes.number,
-  label: PropTypes.string,
+  label: PropTypes.func,
```
sohkai: Should this just be an element to allow for both strings and components?

src/components/Countdown/Countdown.js
```diff
@@ -11,47 +11,54 @@ const FRAME_EVERY = 1000 / 30 // 30 FPS is enough for a ticker
...
+    )
+  }
+  renderTime = () => {
+    const { end } = this.props
+    const { days, hours, minutes, seconds, totalInSeconds } = difference(
+      end,
+      new Date()
+    )
+    if (totalInSeconds <= 0) {
+      return <TimeOut>Time out</TimeOut>
```
sohkai: Not sure: do you think this should say `Timed out` (past tense)?

src/components/SidePanel/SidePanel.js
```diff
@@ -109,43 +110,47 @@ class SidePanel extends React.Component {
...
+                    <StyledPanelHeader>
+                      <h1>
+                        <Text size="xxlarge">{title}</Text>
+                      </h1>
+                      {!blocking && (
+                        <StyledPanelCloseButton
+                          type="button"
+                          onClick={this.handleClose}
+                        >
+                          <img src={prefixUrl(close, publicUrl)} alt="Close" />
```
sohkai: Could this just use `styledUrl()`?

src/providers/Redraw/Redraw.js
```diff
@@ -0,0 +1,49 @@
...
+class Redraw extends React.Component {
+  static propTypes = {
+    interval: PropTypes.number,
+    children: PropTypes.func.isRequired,
+  }
+  static defaultProps = {
+    interval: 1000,
+  }
+  state = {
+    lastDraw: -1,
```
sohkai: To simplify `this.draw()`, what do you think about using `Date.now()` here as the initial state?

src/providers/Redraw/Redraw.js
```diff
@@ -0,0 +1,49 @@
...
+  }
+  componentDidMount() {
+    this.raf = null
```
sohkai: Do we need to set it to `null` here?

src/providers/Redraw/Redraw.js
```diff
@@ -0,0 +1,49 @@
...
+    const { interval } = this.props
+    const { lastDraw } = this.state
+    const now = Date.now()
+    const delta = now - lastDraw
+    if (lastDraw === -1 || delta > interval) {
+      this.setState({ lastDraw: now - delta % interval })
+    }
+  }
+  render() {
+    return this.props.children()
```
sohkai: What do you think about passing down `lastDraw`?

src/providers/RedrawFromDate/RedrawFromDate.js
```diff
@@ -0,0 +1,79 @@
...
+  static propTypes = {
+    fromDate: PropTypes.oneOfType([
+      PropTypes.string,
+      PropTypes.number,
+      PropTypes.instanceOf(Date),
+    ]),
+  }
+  state = {
+    redrawTime: EVERY_HOUR,
+    lastDraw: -1,
```
sohkai: Could this also just use `Date.now()` for the initial state?

src/providers/RedrawFromDate/RedrawFromDate.js
```diff
@@ -0,0 +1,79 @@
...
+    const { fromDate } = this.props
+    if (fromDate) {
+      this.restartDrawInterval(getRedrawTime(fromDate))
+    }
+  }
+  componentWillReceiveProps({ fromDate }) {
+    if (!fromDate && this.props.fromDate) {
+      this.clearInterval()
+    } else if (!isEqual(fromDate, this.props.fromDate)) {
+      this.restartDrawInterval(getRedrawTime(this.props.fromDate))
```
sohkai: I think this should be using the new prop; `getRedrawTime(fromDate)`?

src/providers/RedrawFromDate/RedrawFromDate.js
```diff
@@ -0,0 +1,79 @@
...
+      this.setState({ lastDraw: Date.now() })
+
+      const newRedrawTime = getRedrawTime(this.props.fromDate)
+      if (newRedrawTime !== redrawTime) {
+        this.restartDrawInterval(newRedrawTime)
+      }
+    }, redrawTime)
+  }
+  render() {
+    return this.props.children()
```
sohkai: What do you think about passing down `lastDraw`?

src/utils/index.js
```diff
@@ -0,0 +1,4 @@
+/* eslint-disable prettier/prettier */
+export { difference, formatHtmlDatetime } from './date'
+export { formatIntegerRange } from './format'
+export { font, grid, spring, breakpoint, BreakPoint, unselectable } from './styles'
```
sohkai: Should this also re-export `utils/math` and `utils/url`?

src/utils/url.js
```diff
@@ -0,0 +1,5 @@
+// prefix helper
+const prefixUrl = (url, publicUrl) =>
```
sohkai: Probably better to rename as `prefixPublicUrl`

src/providers/RedrawFromDate/index.js
```diff
@@ -0,0 +1,2 @@
+export { default } from './RedrawFromDate'
```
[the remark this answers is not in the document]
bpierre: > E.g. you can do `import RedrawFromDate from 'providers/RedrawFromDate'`?

Yes that’s what it does. 👍

> And does `export * from ...` not re-export the default?

No, `*` exports everything except `default`.

PR #160, opened 2018-04-27
Gallery: add PageRedraw / PageRedrawFromDate pages

bpierre:
To be merged before https://github.com/aragon/aragon-ui/pull/158

### DIFF
--- a/gallery/src/routes.js
+++ b/gallery/src/routes.js
@@ -30,7 +30,9 @@ import PageCard from './pages/PageCard'
 import PageEmptyStateCard from './pages/PageEmptyStateCard'
 import PageTable from './pages/PageTable'
 
-// RxJS
+// Providers
+import PageRedraw from './pages/PageRedraw'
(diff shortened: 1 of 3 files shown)


gallery/src/routes.js
```diff
@@ -84,8 +86,10 @@ export const PAGE_GROUPS = [
     ].map(preparePage),
   },
   {
-    name: 'RxJS',
```
sohkai: What do you think about keeping the RxJS section? `observe()` is not really a generic provider, as it's only meant to be used with an observable.

sohkai: Nah, if I had to pick between not having the section and `RxJSObserve`, I'd pick no section. Let's merge!

bpierre: I’m not super satisfied with the “Providers” section either: it’s related to their implementation rather than their role, so we can do that for now until we reorganize things a bit better.

PR #175, opened 2018-05-27
RadioButton: styling tweaks

bpierre:
<p align=center>

</p>

### DIFF
--- a/src/components/Radio/RadioListItem.js
+++ b/src/components/Radio/RadioListItem.js
@@ -60,19 +60,12 @@ class RadioListItem extends React.Component {
   }
 }
 
-// Utility styles from RadioButton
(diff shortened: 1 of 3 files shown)


src/components/Radio/RadioListItem.js
```diff
@@ -88,14 +81,9 @@ const LabelBox = styled.div`
   cursor: pointer;
 
   &:focus,
-  &:hover,
-  ${RadioButton}:focus ~ &,
-  ${RadioButton}:hover ~ & {
```
sohkai: I did like having the the dimmed state in the radio button when you hover on it to give some feedback to the user. @bpierre @jounih do you think that looks weird?

PR #181, opened 2018-06-09
SidePanel: move to react-spring

bpierre:
Also export a new `springs` object.


src/utils/styles/spring.js
```diff
@@ -1,18 +1,41 @@
...
+  slow: { tension: 150, friction: 18 },
+  normal: { tension: 190, friction: 22 },
+  fast: { tension: 220, friction: 24 },
 }
 
-export const spring = (name: SpringName) => {
-  return SPRINGS[name] || SPRINGS.normal
+// Deprecated, see above
+export const spring = name => {
+  return springs[name] || springs.normal
```
sohkai: Might be useful to have a depreciation strategy, like using `console.warn()` if it's not in production mode.

src/utils/styles/spring.js
```diff
@@ -1,18 +1,41 @@
...
 }
+
+const reactMotionSprings = Object.entries(springs).reduce(
```
sohkai: Not sure if this needs to be here; doesn't seem to be getting used? Should we export it?

bpierre: Ah yes it should be used by `spring()` for react-motion. Thanks!

bpierre:
> Do you think it'd be nice to have a doc somewhere that lists the things we're starting to depreciate? A full blown changelog would be great but it's probably too much for now.

Yes it’s a good idea, we should start having a CHANGELOG.md in this project. I will add with the next release.

PR #183, opened 2018-06-09
Button.Anchor: extend its styles with styled()

bpierre:
Fix the settings buttons in aragon/aragon.


sohkai:
Ahhh we probably don't need #186 then!

PR #186, opened 2018-06-18
Pin styled-components to be <= 3.2.6

sohkai:
Looks like `styled-components@3.3.x` breaks the way we're styling `Button.Anchor` :(


sohkai:
Closing for #186.

PR #189, opened 2018-06-27
ESLint fixes

bpierre:
Mostly to use named imports when available.

Also move AppBar to PropTypes (from flow).


src/providers/RedrawFromDate/RedrawFromDate.js
```diff
@@ -67,11 +67,17 @@ class RedrawFromDate extends React.Component {
...
-    {() => <Component {...props} />}
-  </RedrawFromDate>
-)
+const hocWrap = Component => {
+  const HOC = props => (
+    <RedrawFromDate fromDate={props.fromDate}>
+      {() => <Component {...props} />}
+    </RedrawFromDate>
+  )
+  HOC.propTypes = {
```
sohkai: In these `hocWrap` components, can we use `HOC.displayName = '<name>(getDisplayName(Component))'`?

Otherwise, we can remove the `react-display-name` dependency entirely.

PR #194, opened 2018-07-15
NavigationBar component

bpierre:
Usage (API not finalized yet):

```jsx
class App extends React.Component {
  state = {
    items: ['Level 1'],
  }
  reset = () => {
    this.setState(({ items }) => ({ items: ['Level 1'] }))
  }
  forward = () => {
    this.setState(({ items }) => ({ items: [ ...items, `Level ${items.length + 1}` ] }))
  }
  backward = () => {
    if (items.length <= 1) {
      return
    }
    this.setState(({ items }) => ({ items: items.slice(0, -1) }))
  }
  render() {
    const { items } = this.state
    return (
      <AragonApp>
        <AppView
          appBar={
            <AppBar>
              <NavigationBar items={items} onBack={this.backward} />
            </AppBar>
          }
        >
          <Main>
            <div>
              <button onClick={this.backward} disabled={items.length < 2}>
                back
              </button>
              <button onClick={this.reset} disabled={items.length < 2}>
                reset
              </button>
              <button onClick={this.forward}>
                forward
              </button>
            </div>
          </Main>
        </AppView>
      </AragonApp>
    )
  }
}

```

The API is inspired from [UIKit / UINavigationBar](https://developer.apple.com/documentation/uikit/uinavigationbar).


derekalia:
Is this going to be tied to routing in any way?

bpierre:
@derekalia probably not directly, to keep Aragon UI separated from anything specific to Aragon apps, but I would like to provide different ways of using it:

1. “Manually” (kind of like it is now, handling the `onBack` event manually).

2. Through a simpler (but more restrictive) “screens” declaration, where you could define components that would displayed for every screen of your app. It would handle the `onBack` event, and there should be an easy way to connect it to the routing API, which might come either from the “react-aragon” library or from optional props so that navigation is handled by receiving / sending messages (i.e. the routing API).

src/components/NavigationBar/NavigationBar.js
```diff
@@ -0,0 +1,138 @@
...
+      direction: state.cachedItems.length > props.items.length ? 1 : -1,
+    }
+  }
+  componentDidMount() {
+    this.setState({ animate: true })
+  }
+  render() {
+    const { onBack, items } = this.props
+    const { animate } = this.state
+    const displayedItems = items
```
sohkai: Since this is always an array of 1, maybe we can just have a single `displayedItem`?

src/components/NavigationBar/NavigationBar.js
```diff
@@ -0,0 +1,138 @@
...
+  </animated.span>
+)
+
+const Container = styled.span`
+  display: flex;
+  position: relative;
+  height: 100%;
+`
+
+const SpacedBadge = styled(Badge)`
```
sohkai: Not sure if this was meant to be used later

PR #197, opened 2018-08-14
Add util function to ensure a trailing / on publicUrl

Huanzhang89:


src/utils/url.js
```diff
@@ -1,5 +1,8 @@
 // prefix helper
-const prefixUrl = (url, publicUrl) =>
+export const prefixUrl = (url, publicUrl) =>
   url.startsWith('data:') ? url : publicUrl + url
 
-export { prefixUrl }
+// trailing slash helper
+export const ensureSlash = path => {
+  return path.endsWith('/') ? `${path}/` : path
```
bpierre: Shouldn’t it be the other way?

bpierre: I wonder if we could name it something like `ensureFinalSlash`, `ensureTrailingSlash` or just `trailingSlash`, to make it clear that it is added at the end of the string?

Huanzhang89: oh wow mistakes were made

Huanzhang89: Yeah agreed, will update the name

PR #209, opened 2018-10-03
Update wrapper (App & View) heights

AquiGorka:
With the addition of a top-level banner [here](https://github.com/aragon/aragon/pull/371) scrollable wrappers using `100vh` were not rendering the correct height. These changes update the rules and the content scrolls correctly.

### DIFF
--- a/src/components/AragonApp/AragonApp.js
+++ b/src/components/AragonApp/AragonApp.js
@@ -8,8 +8,11 @@ import { ensureTrailingSlash } from '../../utils/url'
 import logo from './assets/logo-background.svg'
 
 const StyledAragonApp = styled.main`
+  width: 100%;
+  height: 100vh;
+  display: flex;
+  flex-direction: column;
   min-width: 320px;
-  min-height: 100vh;
   background-color: ${theme.mainBackground};
   background-image: ${({ backgroundLogo }) =>
     backgroundLogo ? css`url(${PublicUrl.styledUrl(logo)})` : 'none'};
--- a/src/components/AragonApp/AppView.js
+++ b/src/components/AragonApp/AppView.js
@@ -29,7 +29,7 @@ class AppView extends React.Component {
 
 const Main = styled.div`
   display: flex;
-  height: 100vh;
+  height: 100%;
   flex-direction: column;
   align-items: stretch;
   justify-content: stretch;


src/components/AragonApp/AragonApp.js
```diff
@@ -8,8 +8,11 @@ import { ensureTrailingSlash } from '../../utils/url'
 import logo from './assets/logo-background.svg'
 
 const StyledAragonApp = styled.main`
+  width: 100%;
+  height: 100vh;
+  display: flex;
```
bpierre: Do we need to have `display: flex`? I fear that it would break the layout of apps / websites already using this component (flex children can behave differently than block elements).

src/components/AragonApp/AragonApp.js
```diff
@@ -8,8 +8,11 @@ import { ensureTrailingSlash } from '../../utils/url'
 import logo from './assets/logo-background.svg'
 
 const StyledAragonApp = styled.main`
+  width: 100%;
+  height: 100vh;
```
bpierre: We should probably use `min-height` here, e.g. for websites that are not using `AppView` and expect this to extend with the content.

bpierre:
Rendering https://github.com/aragon/website breaks the layout: [before](https://user-images.githubusercontent.com/36158/46617720-62f86d80-cb15-11e8-9107-d98888240b23.png) vs. [after](https://user-images.githubusercontent.com/36158/46617728-6a1f7b80-cb15-11e8-8a01-856d26279d8a.png).

I wonder if we shouldn’t move away from this component in aragon/aragon?

We could keep `AragonApp` as it is, and create a new component, that could be named `<Root>` or something similar, and that would only:

- Set the props of the context providers we want to make available to components.
- Render `<BaseStyles />`.

For backward compatibility, `AragonApp` could use this new component and only set the styles on top of it.

Also I don’t think we need a dedicated component for the background color / image: `AppView` could provide that.

@AquiGorka @sohkai what do you think?

sohkai:
Agreed, `AragonApp` is a leftover from the dark days when we didn't have `AppView`.

@bpierre What do you think about putting `AragonApp` in a deprecated folder or similar? It'd be nice to upgrade all the apps to not rely on it, but let's do that as we go. Although they also have `100vh` defined I don't think they experience the same problems due to the iframe?

bpierre:
> @bpierre What do you think about putting AragonApp in a deprecated folder or similar?

Sounds good 👍️ We were also talking about having an `experimental/` or `unstable/` folder, it would make sense to have both.

chris-remus:
@bpierre will close.

bpierre:
Closing this issue since Aragon Core is not using `AragonApp` anymore: https://github.com/aragon/aragon/pull/371/commits/b4e78a04e1cddaecce03c311fd589a4f9002f8ac

I also opened an issue to deprecate AragonApp: https://github.com/aragon/aragon-ui/issues/216

Issue #220, opened 2018-10-15
TransactionProgress component


bpierre:
<img width="599" alt="image" src="https://user-images.githubusercontent.com/36158/46942809-08579800-d066-11e8-959b-52d60c24a001.png">

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__This issue now has a funding of 720.0 ANT (555.98 USD @ $0.77/ANT)  attached to it as part of the Aragon Project fund.__

 * If you would like to work on this issue you can 'start work' [on the Gitcoin Issue Details page](https://gitcoin.co/issue/aragon/aragon-ui/220/1484).
* Want to chip in? Add your own contribution [here](https://gitcoin.co/issue/aragon/aragon-ui/220/1484).
* Questions? Checkout <a href='https://gitcoin.co/help'>Gitcoin Help</a> or the <a href='https://gitcoin.co/slack'>Gitcoin Slack</a>
* $42,365.17 more funded OSS Work available on the [Gitcoin Issue Explorer](https://gitcoin.co/explorer)

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__Workers have applied to start work__.


These users each claimed they can complete the work by 12 months from now.
Please review their action plans below:


**1) [deamme](https://gitcoin.co/profile/deamme) has applied to start work _(Funders only: [approve worker](https://gitcoin.co/issue/aragon/aragon-ui/220/1484?mutate_worker_action=approve&worker=deamme) | [reject worker](https://gitcoin.co/issue/aragon/aragon-ui/220/1484?mutate_worker_action=reject&worker=deamme))_.**

There was a research period since this component initially was going to be using some sort of Popover underneath to for positioning relative to the mouse. The component is now done with documentation included with an example how to use it with the new Popover component.

Learn more [on the Gitcoin Issue Details page](https://gitcoin.co/issue/aragon/aragon-ui/220/1484).

gitcoinbot:
@deamme Hello from Gitcoin Core - are you still working on this issue? Please submit a WIP PR or comment back within the next 3 days or you will be removed from this ticket and it will be returned to an ‘Open’ status. Please let us know if you have questions!
* [x] warning (3 days)
* [ ] escalation to mods (6 days)

[...]

Issue #199, opened 2018-09-03
PopOver component


bpierre:
From this conversation: https://github.com/aragon/aragon/issues/299#issuecomment-413887124

- A box that appear on top of everything, and that moves itself to stay inside the viewport (like the menus behave on most operating systems).
- Defaults to the default rounded corner + shadow, but there should be a way to disable any styling.
- The focus should follow properly when opened / closed, which would probably require a way to pass the “opener” (DOM element where the focus was before).

To be used for:

- [DropDown](https://ui.aragon.org/drop-down/)
- [ContextMenu](https://ui.aragon.org/context-menu/)
- [Tooltip](https://github.com/aragon/aragon-ui/issues/47) (not built)
- [DAO Switcher](https://github.com/aragon/design/issues/9) (not built)
- [Transaction progress component](https://github.com/aragon/aragon-ui/issues/220)
- Address Badge
- [TokenBadge](https://github.com/aragon/aragon-ui/issues/221)

deamme:
Hello! I'll be trying to tackle this issue. I'm a little confused and got some questions:
- What do you mean by: `that moves itself to stay inside the viewport`?
- What's the viewport in this context?

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__This issue now has a funding of 1080.0 ANT (833.98 USD @ $0.77/ANT)  attached to it as part of the Aragon Project fund.__

 * If you would like to work on this issue you can 'start work' [on the Gitcoin Issue Details page](https://gitcoin.co/issue/aragon/aragon-ui/199/1483).
* Want to chip in? Add your own contribution [here](https://gitcoin.co/issue/aragon/aragon-ui/199/1483).
* Questions? Checkout <a href='https://gitcoin.co/help'>Gitcoin Help</a> or the <a href='https://gitcoin.co/slack'>Gitcoin Slack</a>
* $41,809.19 more funded OSS Work available on the [Gitcoin Issue Explorer](https://gitcoin.co/explorer)

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__Workers have applied to start work__.


These users each claimed they can complete the work by 12 months from now.
Please review their action plans below:


[...]

PR #225, opened 2018-10-18
Added Popover and TransactionProgress component

deamme:
Closes https://github.com/aragon/aragon-ui/issues/220 and https://github.com/aragon/aragon-ui/issues/199

This should fix Gitcoin warnings.


src/components/Popover/Popover.js
```diff
@@ -0,0 +1,91 @@
...
+      <Provider value={el}>
+        <div ref={this.handleRef}>{children}</div>
+      </Provider>
+    )
+  }
+}
+
+class PopoverWithProvider extends React.Component {
+  static propTypes = {
+    openerRef: PropTypes.node,
```
bpierre: `PropTypes.node` is to check React nodes, but we want to check for a (required) DOM node here:

```suggestion
    openerRef: PropTypes.instanceOf(Element).isRequired,
```

src/components/Popover/Popover.js
```diff
@@ -0,0 +1,91 @@
...
+    containerRef: PropTypes.node,
```
bpierre:
```suggestion
    containerRef: PropTypes.instanceOf(Element),
```

src/components/Popover/Popover.js
```diff
@@ -0,0 +1,91 @@
...
+    placement: PropTypes.string,
+    gutter: PropTypes.string,
```
bpierre: What would you think about setting a default value like `20px`, so that users don’t have to set it?

src/components/Popover/README.md
```diff
@@ -0,0 +1,98 @@
...
+
+## Properties
+
+### `openerRef`
+
+- Type: `DOMNode`
+
+Reference to the "opener"which is what the Popover is positioning relative to.
+
+### `placement`
```
bpierre: Maybe we could add a link to Popper.js here.

gallery/webpack.config.js
```diff
@@ -65,13 +65,24 @@ module.exports = (env, argv) => {
...
+            presets: [
+              [
+                '@babel/preset-env',
+                {
+                  targets: { browsers: 'last 2 versions, > 2%' },
+                  modules: false,
+                },
+              ],
+              '@babel/preset-react',
+            ],
```
bpierre: We should have these changes in the `.babelrc` already: https://github.com/aragon/aragon-ui/blob/e66810f63bc0695d8c0491a82e6be6f081af77f3/gallery/.babelrc

deamme: Yeah but didn't seem to work for me unfortunately and that fixed it

src/components/Popover/Popover.js
deamme: It's not required when doing positioning relative to the mouse

src/components/Popover/Popover.js
deamme: This should be required though

deamme:
@bpierre I've tried to do your suggested API but can't really seem to do it since it's a bit complicated when using Portals (they don't work since things are executed inside the onClick event closure I think..). I'm now looking into how to close the Popover when clicking outside of it (apparently more complicated than I initially thought because of the current API). I don't understand how I need to fix the Icon issue..

src/components/Popover/Popover.js
```diff
@@ -0,0 +1,91 @@
...
+    return ReactDOM.createPortal(
+      <div style={{ position: 'absolute', top, left, zIndex }}>{children}</div>,
+      containerRef
+    )
+  }
+}
+
+const { Consumer, Provider } = React.createContext('Popover')
+
+class ContainerProvider extends React.Component {
```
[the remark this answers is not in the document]
deamme: Yes, this also seems to work fine

src/icons/components/Attention.js
```diff
@@ -3,10 +3,10 @@ import React from "react"
 const Attention = props => (
   <svg width={22} height={22} viewBox="0 0 22 22" {...props}>
     <g transform="translate(4 4)" fill="none" fillRule="evenodd">
-      <rect fill="#DAEAEF" width={14} height={14} rx={7} />
+      <rect fill={props.bg || "#DAEAEF"} width={14} height={14} rx={7} />
       <path
         d="M6.155 8.547h1.298V3.3H6.155v5.247zM6.045 11h1.529V9.537H6.045V11z"
-        fill="#6D777B"
+        fill={props.color || "#6D777B"}
```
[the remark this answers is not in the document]
deamme: I don't understand how to fix this..

src/components/index.js
```diff
@@ -15,6 +15,7 @@ export { default as EmptyStateCard } from './Card/EmptyStateCard'
 export { default as Field } from './Field/Field'
```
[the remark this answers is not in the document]
deamme: Haha, yes!

src/components/Popover/Popover.js
```diff
@@ -0,0 +1,91 @@
...
+      prevProps.visible !== this.props.visible ||
+      prevProps.placement !== this.props.placement ||
+      prevProps.children !== this.props.children
+    ) {
+      this.destroyPopper()
+      this.initPopper()
+    }
+  }
+
+  initPopper = () => {
```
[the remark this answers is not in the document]
deamme: I thought that if I use `this` inside the method, then I need to rebind it?

src/components/Popover/Popover.js
```diff
@@ -0,0 +1,91 @@
...
+    const { openerRef, placement, gutter } = this.props
+    if (placement && !this.popper) {
+      this.popper = new Popper(openerRef, findDOMNode(this), {
```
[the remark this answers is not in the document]
deamme: Nope.. You need to pass the opener element. The way we are doing mouse positioning is just capturing the mouse position and passing coordinates to the component `top` and `left` styles.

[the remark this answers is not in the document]
deamme: What do you mean by center it? Do you want to do another centering than provided Popper.js placement options?

src/components/Popover/Popover.js
deamme: OpenerRef is not always required..

src/components/Popover/Popover.js
deamme: Ah okay, this seems fine

src/components/Popover/Popover.js
deamme: That's fine, for the gutter I presume?

src/components/Popover/README.md
deamme: Yes

src/components/Popover/Popover.js
deamme: Made it required anyways, since I ran into a problem that requires it.

src/icons/components/Attention.js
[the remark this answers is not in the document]
AquiGorka: `npm run icons:build` uses [`svgr`](https://github.com/smooth-code/svgr) to generate the react components every time so, I'm thinking we could use a _styled-wrapper_ that defines the default color and allows overrides:

```js
// remove fills from svg
// keep src/components/Attention.js intact
// add StyledAttention.js and use this as the icon default export

const StyledAttention = styled(Attention)`
  svg path{
    fill: ${({ color = '#DAEAEF' }) => color}
  }
`
```

As the comments suggest, we'll need to remove the fills from the `svg`s and add new styled wrappers which will become the exports for the react-svg-components.

Thoughts?

src/components/TransactionProgress/TransactionProgress.js
```diff
@@ -0,0 +1,116 @@
...
+      <Transition
+        config={springs.swift}
+        from={{ scale: 0.9, opacity: 0 }}
+        to={{ scale: 1, opacity: 1 }}
+        native
+      >
+        {({ scale, opacity }) => (
+          <Card
+            style={{
+              opacity,
```
AquiGorka: Not sure why but when testing the component page locally, the component would not show (kept `opacity: 0`).
Fixed by updating to: `opacity: `${opacity}`

Did this happen to you @deamme?  Maybe @bpierre can shed some light here?

src/icons/components/Attention.js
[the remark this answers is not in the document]
bpierre: I’m going to work on this part, I’m not sure styled-components is needed here but it’s an interesting idea 🤔

I’ll give it a go!

src/components/ProgressBar/ProgressBar.js
```diff
@@ -0,0 +1,50 @@
...
+}
+
+const Main = styled.div`
+  width: 100%;
+  align-items: center;
+`
+const Base = styled.div`
+  width: 100%;
+  height: 8px;
+  background: #edf3f6;
```
AquiGorka: This is used in the [`Slider`](https://github.com/aragon/aragon-ui/blob/master/src/components/Slider/Slider.js#L186) component too, should we add it to the [`theme`](https://github.com/aragon/aragon-ui/blob/master/src/theme/aragon.json)?

src/components/TransactionProgress/TransactionProgress.js
deamme: I think this has happened to @sohkai too but never happened to me

bpierre:
@deamme I started adding some changes, I hope you don’t mind!

- Renamed `handleClose` to `onClose` (this is the convention we use: `onFoo` when it’s a prop, `handleX` when it’s a method).
- `RootProvider` is now `Root` (+ `Root.Provider`), and is exported like the other providers.
- `Popover` now manages the card style + transition, and `TransactionProgress` is using it directly, so that users can use `TransactionProgress` without having to worry about `Popover`. I think we might want to separate it in the future (e.g. `Popover` for positioning, and `CardPopover` for the style), or have a prop, but we can do this later.
- In the gallery, import the components using `@aragon/ui` (compiled) rather than `ui-src` (source), so that we can remove the specific babel configuration.

src/components/ProgressBar/ProgressBar.js
bpierre: There is some work to do on the theme with the move to Lorikeet, so let’s fix the two together, when working on the new palette.

src/components/TransactionProgress/TransactionProgress.js
bpierre: It happened to me too, not sure why, but it’s fixed since Popover is managing it.

src/providers/Root/Root.js
```diff
@@ -0,0 +1,26 @@
...
+  state = { element: null }
+  handleRef = element => {
+    this.setState({ element })
+  }
+  render() {
+    const { element } = this.state
+    const { children } = this.props
+    return (
+      <Provider value={element}>
+        <div ref={this.handleRef}>{children}</div>
```
AquiGorka: This provider adds a wrapper `div`, maybe this will bring issues with styling (we'll need to pay attention to heights).

deamme:
@bpierre Sounds good! I don't mind at all 😄

src/providers/Root/Root.js
bpierre: Good point, it must be the one rare case where we need to create an element, so that we can refer to it. Tested on aragon/aragon and aragon-website, both are working with it since they don’t set any height on `body`.

bpierre:
@deamme I just merged it! Thank you for this big piece of work 🎉✨💯

PR #226, opened 2018-10-21
EmptyStateCard: deprecates passing components for icon and actionButton

bpierre:
- `action` is the new prop that allow to override the button.
- `icon` now accepts elements (passing a component still works but is deprecated).
- `actionButton` still works but is deprecated.

Some warnings will be displayed in the console when deprecated props (or prop types) are used.

This commit also adds:

- Custom prop types in `src/proptypes.js`.
- A `noop()` function in utils.
- lodash-es as a dependency.
- A devbox app for `EmptyStateCard`.
- `log()`, `warn()` and `devOnly` utils.

### DIFF
--- a/src/components/Card/EmptyStateCard.js
+++ b/src/components/Card/EmptyStateCard.js
@@ -2,7 +2,7 @@ import React from 'react'
 import PropTypes from '../../proptypes'
 import styled from 'styled-components'
 import { theme } from '../../theme'
-import { noop } from '../../utils'
+import { noop, warn } from '../../utils'
 import Button from '../Button/Button'
 import Text from '../Text/Text'
 import Card from './Card'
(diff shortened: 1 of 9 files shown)


src/components/Card/EmptyStateCard.js
```diff
@@ -1,82 +1,150 @@
...
+    this.displayWarnings({}, props)
+  }
+
+  componentDidUpdate(prevProps) {
+    this.displayWarnings(prevProps, this.props)
+  }
+
+  displayWarnings(prevProps, props) {
+    // Deprecated warnings
+    if (prevProps.actionButton !== props.actionButton && props.actionButton) {
```
sohkai: It'd be nice to preface this with `if (process.env.NODE_ENV === 'development')`

PR #231, opened 2018-10-24
Add the IdentityBadge component

bpierre:
Only the badge for now, the popover will be added later.

image


package.json
```diff
@@ -83,9 +83,11 @@
     "lodash-es": "^4.17.11",
     "onecolor": "^3.0.5",
     "prop-types": "^15.6.0",
+    "react-blockies": "^1.4.0",
     "react-display-name": "^0.2.3",
     "react-onclickout": "^2.0.8",
     "react-spring": "^5.6.8",
-    "recursive-copy": "^2.0.9"
+    "recursive-copy": "^2.0.9",
+    "web3-utils": "^1.0.0-beta.36"
```
bpierre: web3-utils adds about 200k (it doesn’t seem to be any version using ESM). I will investigate on this later, even if it means having to copy the functions we use here.

src/utils/index.js
```diff
@@ -1,15 +1,8 @@
...
-  spring,
-  springs,
-  breakpoint,
-  BreakPoint,
-  unselectable,
-} from './styles'
+export * from './date'
+export * from './format'
+export * from './styles'
+export * from './web3'
```
bpierre: I moved to export everything here, since we are re exporting everything and to make it similar to [src/index.js](https://github.com/aragon/aragon-ui/blob/759f51a9175d93d67f194654ea47eecaf5cbc4c7/src/index.js)

package.json
sohkai: I wonder if we can get away with not having it. AFAIK we only use it right now for the `isAddress()` check, and I think this library shouldn't concern itself with too much of the data's consistency.

What do you think about providing a `validateAddress` prop to `IdentityBadge`, and we can default it be a simple regex that doesn't check for checksums (e.g. `/^0x[0-9a-fA-F]{40}$/`)? If the user wants to change the validation, they can provide their own.

At some point this component will also have to handle ENS names, and it's unlikely we'll want to handle the check that the address and ENS name match up correctly in this component (it should just render what it gets).

bpierre: Cool yes let’s do that. 👍

For ENS, we might want the badge to resolve the Ethereum address from an ENS domain… let’s discuss this later :D

sohkai: I'd suggest we make a different package, or extension, with some of these web3 goodies. It's not particularly difficult for a developer to resolve and reverse resolve an ENS name themselves, but we could provide HOCs separately with this functionality.

PR #235, opened 2018-10-28
IdentityBadge: open etherscan when clicked

bpierre:
This is temporary (so we can use the badge everywhere) and will be replaced by the Popover when ready.

### DIFF
--- a/src/components/IdentityBadge/IdentityBadge.js
+++ b/src/components/IdentityBadge/IdentityBadge.js
@@ -3,7 +3,12 @@ import PropTypes from 'prop-types'
 import styled from 'styled-components'
 import Blockies from 'react-blockies'
 import { Text, SafeLink } from '..'
-import { isAddress, shortenAddress, blockExplorerUrl } from '../../utils'
+import {
+  isAddress,
+  shortenAddress,
+  blockExplorerUrl,
+  stylingProps,
+} from '../../utils'
 
 const IDENTICON_SCALE = 3
 const IDENTICON_SQUARES = 8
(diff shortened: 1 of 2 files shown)


src/utils/web3.js
```diff
@@ -45,9 +47,40 @@ export function shortenAddress(address, charsLength = 4) {
...
+  if (provider !== 'etherscan') {
+    return ''
+  }
+
+  if (!ETHERSCAN_NETWORKS.includes(network)) {
+    return ''
+  }
+
+  if (!ETHERSCAN_TYPES.includes(type)) {
+    return ''
```
sohkai: It'd be nice to warn in all of these cases where we return an empty string

src/components/IdentityBadge/IdentityBadge.js
```diff
@@ -53,6 +59,13 @@ const Main = styled.div`
   cursor: default;
 `
 
+const MainAsLink = styled(Main.withComponent(SafeLink)).attrs({
```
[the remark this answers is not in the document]
sohkai: Could we add a comment for this, or immediately merge master with `next` and add a commit to fix this?

src/components/IdentityBadge/IdentityBadge.js
```diff
@@ -24,8 +24,14 @@ class IdentityBadge extends React.PureComponent {
   render() {
     const { entity, shorten, fontSize, ...props } = this.props
     const address = isAddress(entity) ? entity : null
+    const MainComponent = address ? MainAsLink : Main
     return (
-      <Main title={address} onClick={this.handleClick} {...props}>
+      <MainComponent
+        title={address}
+        onClick={this.handleClick}
+        {...(address ? { href: blockExplorerUrl('address', address) } : {})}
```
sohkai: It'd be nice to have some way to provide the `network` into this component, so it makes the right URL.

src/utils/web3.js
```diff
@@ -45,9 +47,40 @@ export function shortenAddress(address, charsLength = 4) {
...
 export function isAddress(address) {
   return ADDRESS_REGEX.test(address)
 }
+
+/**
+ * Generates an etherscan URL
+ *
+ * @param {string} type The type of URL (block, transaction, address or token).
+ * @param {string} value Identifier of the object, depending on the type (block number, transaction hash, …).
+ * @param {string} network The Ethereum network (mainnet, kovan, rinkeby or ropsten).
```
sohkai: I'd call this the `networkType`, for more clarity (on web3 the API's `web3.eth.net.getNetworkType()`), since usually the network is actually canonically specified by a numeric ID (the names are just sugar).

PR #240, opened 2018-11-02
Feature move Modal to Aragon UI

AquiGorka:
These changes include:

- `ModalProvider`, `ModalConsumer`, `ModalView`, `Modal` and `EscapeOutside` components.
- `PropTypes` for the components

**Update**

- Pages for `ModalManager` & `Modal`


AquiGorka:
> What would you think about using something like focus-trap-react to prevent the focus to move outside until the modal gets closed?

I like the idea of _focus-trapping_ modals, what were you thinking? A `<FoucsTrap>{children}</FocustTrap>` that we could reuse or simply add the functionality to `Modal`?

> Also, I wonder if we shouldn’t make it more generic, removing title, body (to use children instead) and More? Then we could move these in a specialized modal (e.g. Dialog, ModalInfo, …).

This implies changing current implementations, specifically `DeperecatedDAO`. I do prefer the interface you suggest (and if I recall correctly I would've implemented it this way but we had the specific use case of the _deprecated dao modal_ and I we agreed on using `title`, `body` and `more` instead of leaving the content open).

> Still about the API, I wonder if we could get rid of ModalProvider entirely? By rendering it on Root like Popover does

That means only one `Modal` at root level. With `Consumer/Provider` the user could have modals within other parts of their apps - in that sense, how could an app display a Modal inside the client view only for itself and non-blocking for the let's say the `MenuPanel`?

If possible I'd like to merge this as is and then open another issue to:

1. Update api and interface (if we decide to go down the "portal" way)
1. Update current implementations of Modal

Let me know your thoughts @bpierre

bpierre:
> I like the idea of focus-trapping modals, what were you thinking? A `<FoucsTrap>{children}</FocustTrap>` that we could reuse or simply add the functionality to Modal?

I would simply add the feature to `Modal` for now, I don’t see anything else that would benefit from it, and if `Modal` is generic enough, there would be no need to create another modal implementation.

> That means only one Modal at root level.

I’m not sure I get this? Here is what I have in mind with two modals at the root level: https://codesandbox.io/s/onxo8x4n6

> With Consumer/Provider the user could have modals within other parts of their apps - in that sense, how could an app display a Modal inside the client view only for itself and non-blocking for the let's say the MenuPanel?

I think a modal should always be displayed at the root level, which can be decided by the user, but we could let people override it if they really need something else (by providing another DOM element to host the portal). In the case of Aragon apps running in an iframe, they will be contained inside the iframe like everything else, so the MenuPanel won’t be blocked by it.

> If possible I'd like to merge this as is and then open another issue to:
> - Update api and interface (if we decide to go down the "portal" way)
> - Update current implementations of Modal

If we plan to change the external API of `<Modal />`, I think we should do it before releasing and having users relying on it, so we don’t have to release a breaking change right after its introduction. Do you see any issue with updating this PR until we are happy with the API?

AquiGorka:
@bpierre added the `Root` provider as we discussed offline.

src/components/Modal/README.md
```diff
@@ -0,0 +1,65 @@
...
+class App extends React.Component {
+  state = {
+    openModal: false,
+  }
+
+  render() {
+    const { openModal } = this.state
+
+    return (
+      <RootProvider>
```
bpierre: Can we use `Main` here?

AquiGorka: Done

PR #241, opened 2018-11-06
Countdown: fix flickr when counting down

AquiGorka:
While reviewing #225 I ran into this:

flickr


bpierre:
I agree that it jumps a bit, but having a fixed width creates inconsistent spacing:

image

What would you think about doing this, but only for the seconds?

AquiGorka:
Latest update (adjusts the position of the `S` to keep consistent spacing):

seconds

Issue #215, opened 2018-10-11
Toast component


bpierre:
It could be useful to have a notification “toast” component, similar to [the notify feature of `UIkit`](https://getuikit.com/v2/docs/notify.html).

Two components are needed:

- The toast component itself: a styled box that renders its `children`.
- A manager that would position the boxes and animate their enter / leave transitions.

Questions / thoughts:

Do we want to allow users to discard them by clicking on a close button? Or by clicking directly on the message?
=> no behavior on press for now.

Instead of closing a message being clicked, we could emit an event so that it could do something (like macOS notifications).
=> no behavior on press for now.

How does it translate on small screens? Apart from the appearance, we should probably only display one at a time.
=> see [Jouni’s comment](https://github.com/aragon/aragon-ui/issues/215#issuecomment-429263808) below

About the API, an approach could be to use a provider + a function like [React toastify](https://fkhadra.github.io/react-toastify/) does.

To do:

- [x] Visual design (@jounih).
- [ ] Toast component implementation.
- [ ] Manager implementation.

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__This issue now has a funding of 500.0 ANT (430.6 USD @ $0.86/ANT)  attached to it__ as part of the Aragon Project fund__.__

 * If you would like to work on this issue you can 'start work' [on the Gitcoin Issue Details page](https://gitcoin.co/issue/aragon/aragon-ui/215/1456).
* Want to chip in? Add your own contribution [here](https://gitcoin.co/issue/aragon/aragon-ui/215/1456).
* Questions? Checkout <a href='https://gitcoin.co/help'>Gitcoin Help</a> or the <a href='https://gitcoin.co/slack'>Gitcoin Slack</a>
* $42,691.04 more funded OSS Work available on the [Gitcoin Issue Explorer](https://gitcoin.co/explorer)

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__Workers have applied to start work__.


These users each claimed they can complete the work by 12 months from now.
Please review their action plans below:


[...]

PR #243, opened 2018-11-06
#215 add MessageHub component (obsolete)

drcmda:
https://github.com/aragon/aragon-ui/issues/215

Adds a MessageHub component to AragonUI, documentation and running example included.

messagehub

### DIFF
--- a/src/components/index.js
+++ b/src/components/index.js
@@ -33,7 +33,7 @@ export { default as TableHeader } from './Table/TableHeader'
 export { default as TableRow } from './Table/TableRow'
 export { default as Text } from './Text/Text'
 export { default as TextInput } from './Input/TextInput'
-export { MessageHub, Toast } from './MessageHub/MessageHub'
+export { ToastHub, Toast } from './ToastHub/ToastHub'
 export {
   default as TransactionProgress,
 } from './TransactionProgress/TransactionProgress'


src/components/MessageHub/MessageHub.js
```diff
@@ -0,0 +1,186 @@
...
+  position: absolute;
+  bottom: ${props => (props.top ? '10px' : '0')};
+  left: 0px;
+  width: auto;
+  background-image: linear-gradient(130deg, #00b4e6, #00f0e0);
+  height: 5px;
+`
+
+let id = 0
+let { Provider, Consumer } = React.createContext()
```
bpierre: I was thinking that maybe it would be nice to have, as a default value, a function that would just throw an error explaining that `MessageHub` need to be declared at an upper level for `Toast` to work.

src/components/MessageHub/MessageHub.js
```diff
@@ -0,0 +1,186 @@
...
+              </Message>
+            )}
+          </Transition>
+        </Container>
+      </React.Fragment>
+    )
+  }
+}
+
+MessageHub.Toast = Consumer
```
bpierre: We are thinking about adding soon a global component that would declare all the common providers, so that people don’t have to do it themselves when they want to use a component that require a provider. Kind of what [`AragonApp`](https://github.com/aragon/aragon-ui/blob/master/src/components/AragonApp/AragonApp.js#L50) does with `PublicUrl` at the moment, but without the specific styles, and with all the common providers (as long as their instantiation is not too slow). It would be optional, but kind of the “easy way” to use Lorikeet. Maybe it could be named `<Lorikeet>`, `<Base>` or something similar.

That’s why I think it would be nice to export `Toast` directly too (maybe not here but in `components/index.js`). That way, people could use `<Toast />` and it would work directly, without having to declare the provider at the top level of their app.

What do you think?

src/components/MessageHub/MessageHub.js
```diff
@@ -0,0 +1,186 @@
...
+    config: PropTypes.object,
+    timeout: PropTypes.number,
+    showIndicator: PropTypes.bool,
+    threshold: PropTypes.number,
+    position: PropTypes.PropTypes.oneOf(['start', 'center', 'end']),
+    top: PropTypes.bool,
+  }
+
+  static defaultProps = {
+    config: { tension: 125, friction: 20, precision: 0.1 },
```
bpierre: Maybe we could use [`springs.lazy`](https://github.com/aragon/aragon-ui/blob/master/src/utils/styles/spring.js#L13) here, only extending its precision?

src/components/MessageHub/MessageHub.js
```diff
@@ -0,0 +1,186 @@
...
+  padding: 12px 22px;
+  font-size: 1em;
+  display: grid;
+  grid-template-columns: 1fr;
+  grid-gap: 10px;
+  border-radius: 3px;
+  overflow: hidden;
+`
+
+export const Button = styled('button')`
```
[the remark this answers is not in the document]
drcmda: Oh my, you're right ... it had a close button, which i've removed for this PR. Will remove it.

src/components/MessageHub/MessageHub.js
```diff
@@ -0,0 +1,186 @@
...
+let id = 0
+let { Provider, Consumer } = React.createContext()
+
+class MessageHub extends React.PureComponent {
+  static propTypes = {
+    config: PropTypes.object,
+    timeout: PropTypes.number,
+    showIndicator: PropTypes.bool,
+    threshold: PropTypes.number,
+    position: PropTypes.PropTypes.oneOf(['start', 'center', 'end']),
```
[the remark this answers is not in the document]
drcmda: This was me being lazy since i just do `flex-${props.position}` later on : P but it would be no problem.

src/components/MessageHub/MessageHub.js
```diff
@@ -0,0 +1,186 @@
...
+    return (
+      <React.Fragment>
+        <Provider value={this.add} children={children} />
+        <Container position={position} top={top}>
+          <Transition
+            native
+            items={this.state.items}
+            keys={item => item.key}
+            from={{ opacity: 0, height: 0, life: 1 }}
+            enter={{ opacity: 1, height: 'auto' }}
```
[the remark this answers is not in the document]
drcmda: We can use a fixed size as well. auto *is* magic. 😁

drcmda:
I'll fix all the review changes tomorrow. As for the 30px margin, no problem i guess.

drcmda:
Found a way using transforms and visible overlays. All changes applied.

src/components/MessageHub/README.md
```diff
@@ -0,0 +1,69 @@
...
+  </MessageHub>
+)
+```
+
+## Properties
+
+### `config`
+
+- Type: `Object`
+- Default value: `{ tension: 125, friction: 20, precision: 0.1 }`
```
bpierre: I forgot that this need to be updated now

src/components/MessageHub/README.md
```diff
@@ -0,0 +1,69 @@
...
+
+- Type: `Number`
+- Default value: `Infinity`
+
+Set this property to change the threshold of toasts being presented at the same time. This will be used as an aproximation or a guideline, but it will still allow bursts to at least show.
+
+### `position`
+
+- Type: one of `'start'`, `'center'`, `'end'`
+- Default value: `end`
```
bpierre: And these too 😊

src/components/MessageHub/README.md
```diff
@@ -0,0 +1,69 @@
...
+- Type: `Boolean`
+- Default value: `false`
+
+Set this property to `true` to make toasts appear from the top.
+
+## Compound components
+
+### `Toast`
+
+The Toast component allows you to consume the Hubs means of creating toasts.
```
sohkai: I'm not quite sure what this sentence means 😅.

src/components/MessageHub/MessageHub.js
```diff
@@ -0,0 +1,170 @@
...
+import Text from '../Text/Text'
+
+let id = 0
+
+const move = pixel => `translate3d(0,${pixel}px,0)`
+const { Provider, Consumer: Toast } = React.createContext(() => {
+  throw "For Toast to work it needs to be part of a MessageHub's tree, which has to be declared at an upper level!"
+})
+
+class MessageHub extends React.PureComponent {
```
sohkai: Given the nomenclature, I wonder if we should name this `ToastHub`?

src/components/MessageHub/MessageHub.js
```diff
@@ -0,0 +1,170 @@
...
+  static propTypes = {
+    config: PropTypes.object,
```
sohkai: Could we name this `transitionConfig` or `animationConfig` instead? I'd assume things like `position` and `threshold` would go into a typical `config`.

gallery/src/pages/PageMessageHub.js
```diff
@@ -0,0 +1,38 @@
...
+  }
+`
+
+const PageMessageHub = ({ title }) => (
+  <MessageHub>
+    <Page title={title} readme={readme}>
+      <Page.Demo opaque>
+        <Container>
+          <div>
+            <Toast>
```
sohkai: Given how this isn't actually a rendered component, would it make sense to name this `ToastController` or `ToastManager`?

drcmda: @sohkai no problem, should it be ToastHub, ToastManager or ToastController?

src/components/MessageHub/MessageHub.js
drcmda: NP. Should position, showIndicator etc remain single props? Would allow flags `<Hub showIndicator canCancel ...>`.

gallery/src/pages/PageMessageHub.js
bpierre: I think we should keep `ToastHub` for the provider. Now for this one, I like that `<Toast>` is short and simple. What about using `<Toasts>`, since you can generate several from the passed function?

src/components/MessageHub/MessageHub.js
```diff
@@ -0,0 +1,170 @@
...
+    switch (props.position) {
+      case 'left':
+        return 'flex-start'
+      case 'right':
+        return 'flex-end'
+      default:
+        return 'center'
+    }
+  }};
+  @media (max-width: 700px) {
```
AquiGorka: To follow a _mobile-first_ approach would you mind setting the breakpoint with `min-width`?
```css
/* mobile first, rules for all elements */
align-items: center;

/* breakpoints for larger screens, overrides what's necesary */
@media (min-width: 700px) {
  align-items: ${props => {
    switch (props.position) {
      case 'left':
        return 'flex-start'
      case 'right':
        return 'flex-end'
      default:
        return 'center'
    }
  }};
}

File: src/components/MessageHub/MessageHub.js [hunk 14]
```diff
@@ -0,0 +1,170 @@
...
+    align-items: center;
+  }
+`
+
+const Message = styled(animated.div)`
+  box-sizing: border-box;
+  position: relative;
+  width: 42ch;
+  @media (max-width: 700px) {
```
AquiGorka: Would you mind updating these rules too to a _mobile-first_ with `min-width` approach?

drcmda:
Sorry, i seem to have a bit of github trouble atm, shouldn't have used the edit feature. It shut me out for some reason, repo above says "unknown" now. Still need to correct the folder src/components/MessageHub into ToastHub. 😣

drcmda:
Really sorry, don't know how this happened ... 

I've created a new PR: https://github.com/aragon/aragon-ui/pull/249 i don't think Github can fix orphaned PR's.

PR #248, opened 2018-11-09
Fix lint and various other cosmetic changes

sohkai:


src/components/TransactionProgress/TransactionProgress.js
```diff
@@ -9,18 +9,24 @@ import {
...
-  font,
   theme,
 } from '../..'
+import { noop } from '../../utils'
 
 class TransactionProgress extends React.Component {
   static propTypes = {
+    openerRef: PropTypes.instanceOf(Element).isRequired,
+    top: PropTypes.string,
+    left: PropTypes.string,
```
sohkai: This really seems like an abstraction leak on the `Popover`, but not sure how best to handle this other than just passing all other props down to the `Popover` or having a dedicated `popover` prop :/

bpierre: Yeah I think we should just pass `openerRef` here. I would also like to move `TransactionProgress` to `TransactionProgressPopover`, and `TransactionProgress` would be a component that renders a badge opening `TransactionProgressPopover`. That way, people wouldn’t have to deal with `openerRef` at all.

src/providers/Root/Root.js
```diff
@@ -1,8 +1,12 @@
 import React from 'react'
+import PropTypes from 'prop-types'
 
 const { Provider, Consumer } = React.createContext(null)
 
 class RootProvider extends React.Component {
+  static propTypes = {
+    children: PropTypes.node,
```
AquiGorka: Shouldn't `children` be of type `func`?

bpierre: For the consumer yes, but this is the provider :-)

AquiGorka: Ahhh, thank you very much for calrifying. :) TIL.

PR #249, opened 2018-11-09
#215 add ToastHub component

drcmda:
Old history: https://github.com/aragon/aragon-ui/pull/243 (243 was orphaned unfortunately) 

Reference: https://github.com/aragon/aragon-ui/issues/215

Adds a MessageHub component to AragonUI, documentation and running example included.

messagehub


drcmda:
@sohkai i've named the spring config "animationConfig", as for the other props, showIndicator, timeout and so on, should they remain single props or join a config object?

bpierre:
@drcmda @sohkai What about `transitionConfig` or `springConfig`  instead?

Also, exposing the spring config is something that we don’t do for the other components. We could decide to start doing it, but I’m not sure we should let people change the spring configurations of “high level” components (as opposed to “low level” components like Popover), even after the move to Lorikeet and a proper theming system. The general feel depends a lot on the transitions, and I think it is important to have a consistent feel across different apps using the toolkit (e.g. Material Design which is quite strict about this).

About the other props, there is no strict rule yet, but we usually keep them as single props.

drcmda:
In that case i think it would be fine to remove the prop and default to springs.lazy. It would be strange if this one does it different than the bunch.

sohkai:
@drcmda I've added two commits to remove `animationConfig` from the README, and to fix the linter :).

src/components/ToastHub/ToastHub.js
```diff
@@ -0,0 +1,171 @@
...
+    // Then fade out
+    await next({ to: { opacity: 0 } })
+    // Then shrink, the last "true" informs Keyframes that is is the last frame
+    await next({ to: { height: 0 } }, true)
+    this.setState(state => ({
+      leaving: state.leaving.filter(i => i.key !== item.key),
+    }))
+  }
+  render() {
+    const { children, showIndicator, position, top, style } = this.props
```
sohkai: Should we use [`className` instead of `style`](https://www.styled-components.com/docs/basics#styling-any-components)?

drcmda: I wanted to but shied away since React will deprecate className in favour of "class" soon. Should i add it? Replace styles, or support both classname and styles? Overriding was a suggestion that @AquiGorka made

sohkai: > React will deprecate className in favour of "class"

👀 

I'd say let's keep to `className` for now since that's what the rest of the codebase is using, and when the upstream changes from React happen we can wholesale move over.

drcmda: Sure thing, and one last question, is it normal for aragon-ui to delegate "style" or should it be removed in favour of className?

bpierre: I think we should pass `style` as well. There is a utility function, [`stylingProps()`](https://github.com/aragon/aragon-ui/blob/master/src/utils/components.js#L16-L21), to pass the two props at the same time:

```jsx
<Container {...stylingProps(this)}>
```

sohkai:
Ahh @drcmda let's use [`stylingProps`](https://github.com/aragon/aragon-ui/pull/249#discussion_r233013892), since that'll also get rid of the linter warnings for the styling props.

Sorry for the churn!

bpierre:
Merged! 💥

Thank you @drcmda, I can’t wait to start using it! ✨🎉

PR #250, opened 2018-11-13
styled-components v4 (migrate the last v3 bits in master)

bpierre:


src/components/IdentityBadge/IdentityBadge.js
```diff
@@ -28,19 +28,25 @@ class IdentityBadge extends React.PureComponent {
     fontSize: 'normal',
     networkType: 'main',
   }
+  getMainProps(address) {
+    const { networkType } = this.props
+    const baseProps = stylingProps(this)
+    return !address
```
AquiGorka: Can this happen? To not get an `address` all the way down to an `IdentityBadge`?
Would you consider this syntax to avoid the ternary:

```js
{
  ...baseProps,
  !!address && ...{
    as: SafeLink,
    target: '_blank',
    href: blockExplorerUrl('address', address, { networkType }),
    style: { ...baseProps.style, cursor: 'pointer' },
  }
}
```

src/components/Button/Button.js
```diff
@@ -232,13 +237,9 @@ const StyledButton = styled.button.attrs({ type: 'button' })`
...
 const Button = PublicUrl.hocWrap(StyledButton)
-const Anchor = PublicUrl.hocWrap(
-  styled(StyledButton.withComponent(SafeLink))`
-    ${unselectable};
-    display: inline-block;
-    text-decoration: none;
-  `
-)
+const Anchor = PublicUrl.hocWrap(props => (
+  <StyledButton as={SafeLink} {...props} />
```
sohkai: 🤘

src/components/IdentityBadge/IdentityBadge.js
sohkai: It's possible to have a null `address` here if the [address check fails](https://github.com/aragon/aragon-ui/blob/09012fa63ee9e6129c6099280829bcd26a62de18/src/components/IdentityBadge/IdentityBadge.js#L46)

bpierre: > Would you consider this syntax to avoid the ternary

@AquiGorka Yes let’s do that, it doesn’t cost much more to spread `baseProps` in both cases. But I would still use a ternary rather than `&&` for the nested object, to avoid spreading a non-iterable object.

bpierre: I made it an early return in the end 👍

PR #251, opened 2018-11-13
devbox: demos index

bpierre:
Not using the UI toolkit on purpose, to differentiate the index from the actual demos.

image

### DIFF
--- a/devbox/index.js
+++ b/devbox/index.js
@@ -12,22 +12,129 @@ import IdentityBadge from './apps/IdentityBadge'
 import Popover from './apps/Popover'
 import TransactionProgress from './apps/TransactionProgress'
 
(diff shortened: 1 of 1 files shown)


devbox/index.js
```diff
@@ -12,22 +12,129 @@ import IdentityBadge from './apps/IdentityBadge'
...
+  })
+)
+
+const STYLES = `
+*, *:before, *:after {
+  box-sizing: border-box;
+}
+body {
+  margin: 0;
+  --background: hsl(48, 94%, 98%);
```
sohkai: Woah, is this some new styled-components voodoo css?

bpierre: [CSS vars 😁](https://developer.mozilla.org/en-US/docs/Web/CSS/Using_CSS_variables)

sohkai: Ohh, gotcha, they're css variables 😄

PR #252, opened 2018-11-13
chore: update development dependencies

sohkai:
Adds:

- Pre-push lint check
- Pre-commit prettier formatting

Updates:

- babel to `babel@7`
- rollup to latest
- ESLint to latest

Removes:

- flow configuration leftovers

Notes:

- With the upgrade to rollup, there's now a number of complaints about circular dependencies. AFAIK there's no good way to turn this off yet, but it's also a strong signal that maybe we should rethink how we export some of the modules so we can avoid these circular dependencies (see https://github.com/aragon/aragon-ui/issues/253).

---------------

Bundle before/after (no idea why gzip is so different):

screen shot 2018-11-13 at 4 49 21 pm

Left is new, right is old.


.babelrc
```diff
@@ -1,9 +1,17 @@
...
+    ],
+    "@babel/preset-react"
+  ],
   "plugins": [
     ["inline-json-import", {}],
     ["styled-components", { "displayName": true }],
-    ["transform-react-jsx", {}],
-    ["transform-react-display-name", {}]
-  ],
-  "presets": [["env", { "modules": false }], "stage-2"]
```
sohkai: I think the only thing left in [stage 2/3](https://babeljs.io/docs/en/babel-preset-stage-2) that we use is the class properties.

package.json
```diff
@@ -35,43 +35,39 @@
     "styled-components": "^3.0.0"
   },
   "devDependencies": {
-    "babel-cli": "^6.26.0",
-    "babel-eslint": "^8.0.1",
-    "babel-plugin-external-helpers": "^6.22.0",
```
sohkai: Apparently [`rollup-plugin-babel`](https://github.com/rollup/rollup-plugin-babel/blob/master/src/index.js#L87) does this deduping automatically now, as of a year ago 🤷‍♂️

package.json
```diff
@@ -35,43 +35,39 @@
...
-    "rollup-plugin-commonjs": "^8.2.6",
-    "rollup-plugin-filesize": "^1.4.2",
-    "rollup-plugin-gzip": "^1.3.0",
+    "rollup": "^0.67.1",
+    "rollup-plugin-babel": "^4.0.3",
+    "rollup-plugin-commonjs": "^9.2.0",
+    "rollup-plugin-filesize": "^5.0.1",
     "rollup-plugin-node-resolve": "^3.0.0",
     "rollup-plugin-progress": "^0.4.0",
-    "rollup-plugin-rebase": "^0.12.0",
```
sohkai: `rollup-plugin-rebase` wasn't being used, and I don't think we need it anyway since we don't need to move files around when building?

package.json
```diff
@@ -90,5 +86,15 @@
     "react-onclickout": "^2.0.8",
     "react-spring": "^6.1.3",
     "recursive-copy": "^2.0.9"
+  },
+  "browserslist": {
+    "development": "> 1%, last 3 versions, ie >= 9, ios >= 8, android >= 4.2",
+    "production": "> 1%, last 3 versions, ie >= 9, ios >= 8, android >= 4.2"
```
sohkai: I took these from [`aragon/aragon`'s (unused) config](https://github.com/aragon/aragon/blob/master/.babelrc#L8). This pretty closely matches the default config for browserslist, and I'm not sure if we had this in this repo before?

package.json
```diff
@@ -90,5 +86,15 @@
...
+  },
+  "husky": {
+    "hooks": {
+      "pre-push": "npm run lint",
+      "pre-commit": "pretty-quick --staged"
```
AquiGorka: ✨ no more rebasing to fix code format :)

rollup.config.js
```diff
@@ -2,6 +2,7 @@ import resolve from 'rollup-plugin-node-resolve'
 import commonjs from 'rollup-plugin-commonjs'
 import babel from 'rollup-plugin-babel'
 import url from 'rollup-plugin-url'
+import filesize from 'rollup-plugin-filesize'
```
AquiGorka: 👍

sohkai:
@AquiGorka Actually I was thinking that this should have its own independent config, since it'll be more community driven than our apps. I wouldn't want to pepper things like aragon's configs here.

package.json
```diff
@@ -35,43 +35,39 @@
...
+    "eslint-plugin-prettier": "^3.0.0",
+    "eslint-plugin-promise": "^4.0.1",
     "eslint-plugin-react": "^7.4.0",
-    "eslint-plugin-standard": "^3.0.1",
+    "eslint-plugin-standard": "^4.0.0",
     "gh-pages": "^1.0.0",
+    "husky": "^1.1.4",
     "opencolor": "^0.2.0",
     "opencolor-converter": "git://github.com/opencolor-tools/opencolor-converter.git#v2.1.21",
     "prettier": "^1.12.0",
```
sohkai: @bpierre @AquiGorka debating if I should pin this to 1.14 like https://github.com/aragon/aragon/pull/472

bpierre: I would say no, let’s go with the flow 🌊

AquiGorka: Is there any difference in the outputs?
I am reading the changelog from `1.12` ->  `1.13` -> `1.14` and I think it'll be helpful to read through it knowing if there's anything affecting the codebase.

AquiGorka:
>  I was thinking that this should have its own independent config, since it'll be more community driven than our apps.

The `eslint config` provides a base that can be extended, and it can be community driven too.

package.json
sohkai: `1.15` introduced a number of cosmetic changes, the most notable of which is https://prettier.io/blog/2018/11/07/1.15.0.html#javascript

AquiGorka: Ahhh, I get it now, please pin it. Now that I'm trying to normalize `dev` dependencies it would've been so much easier - I am having to find a common version for all the apps.

sohkai: :/ Looks like prettier has a policy of avoiding options, so it's unlikely they'll allow us to configure that particular bit.

Rather than find a common version, it's probably best just to keep updated with the latest.

bpierre: Yes my vote would go to keep following the new versions, except if we have a strong reason to reject a new formatting syntax. If we want more control (to upgrade the codebase at once when a new version is introduced), we could also pin it to the latest version, making sure we update it regularly.

package.json
bpierre: Following this conversation we had some time ago: https://github.com/aragon/aragon.js/issues/47

I think we should use`> 1%` for Aragon UI / Lorikeet (its usage is broader than our apps), and have `>2%, last 1 edge versions, not ie > 0, not op_mini all` for our apps (in the config @AquiGorka is building).

sohkai:
@bpierre @AquiGorka added a few changes:

- Upgrade to `prettier@1.15.0`
- Use `lint-staged` instead of `pretty-quick` (uses all of ESLint rather than just prettier)
- Fixed some linting issues

package.json
bpierre: Yes, we are using `rollup-plugin-url` to do that. I’m wondering if the only difference between the two is that `rollup-plugin-url` can inline files? [The README](https://github.com/sebastian-software/rollup-plugin-rebase#comparison) seem to imply that it is more adapted to a library than `rollup-plugin-url`, but I don’t see how.

sohkai: Yeah... seems like it does the same thing except it doesn't inline?

package.json
```diff
@@ -35,44 +35,40 @@
...
-    "eslint-plugin-flowtype": "^2.39.1",
+    "eslint": "^5.9.0",
+    "eslint-config-prettier": "^3.3.0",
+    "eslint-config-standard": "^12.0.0",
+    "eslint-config-standard-react": "^7.0.2",
     "eslint-plugin-import": "^2.8.0",
-    "eslint-plugin-node": "^5.2.1",
-    "eslint-plugin-prettier": "^2.3.1",
-    "eslint-plugin-promise": "^3.6.0",
+    "eslint-plugin-node": "^8.0.0",
```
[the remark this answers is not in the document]
sohkai: Turns out this is needed by `standard`.

PR #257, opened 2018-11-27
Checkbox

### LINKED ISSUE #50: Add CheckBox
image

Location: `src/Form/CheckBox.js`

macor161:
Fixes #50 

image

Let me know if something needs to be changed.


bpierre:
Hi @macor161, thanks for contributing!

I added a few commits to tweak its visual design and some other things. Let me know if that works for you!

Changes:

- Follow the design from the [aragon/design Sketch file](https://github.com/aragon/design/blob/master/sketch/aragon-web-app.sketch).
- Use react-spring for consistent animation.
- Use a `<IconCheck />` for the tick mark instead of CSS.
- Add a focus state.

Using a component + a custom spring makes it impossible to only use CSS, that’s why the `<input type=checkbox>` has been replaced by a button and ARIA attributes.

It follows the required behavior for the [checkbox role](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/checkbox_role#Required_JavaScript):

- A click event changes its state.
- Using the space bar or the enter key while the item is focused changes its state.

None of these events needed to be actually implemented, because a `<button type=button>` is used and already does it.

Missing states: disabled, mixed (`[−]`). /ping @jounih

Demo:

src/components/Input/CheckBox.js
```diff
@@ -0,0 +1,117 @@
...
+  &:active {
+    border-color: #c9d9de;
+  }
+  &:focus ${FocusRing} {
+    display: block;
+  }
+  &:focus:not(:focus-visible) ${FocusRing} {
+    display: none;
+  }
+  &:focus-visible ${FocusRing} {
```
sohkai: Nice!

src/components/Input/CheckBox.js
```diff
@@ -0,0 +1,117 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import styled from 'styled-components'
+import { Spring, animated } from 'react-spring'
+import { IconCheck } from '../../icons'
+import theme from '../../theme'
+import { springs } from '../../utils'
+
+class CheckBox extends React.Component {
```
sohkai: One small nitpick: I think `Checkbox` (lowercase b) might be better, since it's more commonly used.

src/components/Input/CheckBox.md
```diff
@@ -0,0 +1,19 @@
...
+const onChange = checked => { console.log('checked', checked) }
+
+const App = () => (
+  <div>
+    <label><CheckBox onChange={onChange} checked /> First</label>
+    <label><CheckBox onChange={onChange} /> Second</label>
+    <label><CheckBox onChange={onChange} /> Third</label>
+  </div>
+)
+```
```
sohkai: Could we add descriptions for the proptypes here? In particular, I'm not sure what "mixed" means.

src/components/Input/CheckBox.js
bpierre: Ah, I made the change because both [Android](https://developer.android.com/reference/android/widget/CheckBox) and [macOS](https://developer.apple.com/documentation/appkit/nsaccessibility/role/1525866-checkbox) are using “CheckBox”, and that’s also what we were using in the issue https://github.com/aragon/aragon-ui/issues/50.

I don’t really have any opinion on this, but I want to go for the most obvious one for users. What do you think?

bpierre:
@sohkai When clicking on a checkbox? It should only appear if the window is focused out, or when pressing a key.

sohkai:
> @sohkai When clicking on a checkbox? It should only appear if the window is focused out, or when pressing a key.

@bpierre Yes... either on the checkbox itself or on its label:

<a href="https://imgflip.com/gif/2olnwg"></a>

bpierre:
@sohkai oh got it, it’s because I have experimental web features enabled in Chrome… I thought [`:focus-visible`](https://caniuse.com/#search=%3Afocus-visible) was supported but it’s not the case yet. I’ll replace it with another solution.

src/components/Input/CheckBox.js
AquiGorka: I vote for lowercase b as @sohkai suggested ([ref](https://www.w3.org/TR/html52/sec-forms.html#checkbox-state-typecheckbox))

src/utils/styles/spring.js
```diff
@@ -8,13 +8,17 @@ export const springs = {
...
-  smooth: { mass: 0.7, tension: 300, friction: 25, precision: 0.001 },
+  smooth: { mass: 0.7, tension: 300, friction: 25 },
 
-  // Fast speed spring, for actions that need to feel almost instant.
-  swift: { mass: 0.5, tension: 400, friction: 30, precision: 0.001 },
+  // Fast speed spring, for interactions that feel “almost instant”.
+  swift: { mass: 0.5, tension: 800, friction: 30 },
+
+  // Super fast speed spring, for interactions that feel instant (e.g. a
+  // chekbox tick).
```
AquiGorka: Typo here: `chekbox`

src/components/Input/CheckBox.js
sohkai: Let's use `Checkbox` then :). It seems to be the canonical way to spell it.

bpierre: I found at least one toolkit using `Checkbox` ([Flutter](https://docs.flutter.io/flutter/material/Checkbox-class.html)), let’s go for it then.

bpierre:
> Would it make sense to include in the description that we use the React way of controlled components and that's why there is no need for name/value?

Yes totally agree, we should talk about the general principles we follow when designing the toolkit API 👍

sohkai:
@bpierre Oops, forgot to push the changes for the gallery page earlier :). Also fixed a linting warning with using `theme` as the default export.

gallery/src/pages/PageCheckBox.js
```diff
@@ -0,0 +1,63 @@
...
+    return (
+      <Page title={title} readme={readme}>
+        <Page.Demo>
+          <DemoHeader>Checkbox</DemoHeader>
+          <Container>
+            <div>
+              <label>
+                <Checkbox
+                  checked={first && second}
+                  indeterminate={Boolean(first ^ second)}
```
bpierre: 👌

PR #258, opened 2018-11-28
Fix the “Context.Consumer.Provider is not supported” warning

### LINKED ISSUE #232: React doesn't like attaching Context Providers to Consumers
Related to `PublicUrl`, I think React's giving this warning message because of how we're attaching the `Provider` to its `Consumer`:

screen shot 2018-10-24 at 11 13 08 pm

[This is where it's used](https://github.com/aragon/aragon/blob/master/src/index.js#L25)

bpierre:
Fixes #232


bpierre:
🎁 🎁 🎁

PR #259, opened 2018-12-01
Add inertia scrolling for ios devices

AquiGorka:


src/components/BaseStyles/BaseStyles.js
```diff
@@ -68,6 +68,9 @@ const injectStyles = (asset, legacyFonts) => injectGlobal`
   *:after {
     box-sizing: border-box;
   }
+  html, body {
+    -webkit-overflow-scrolling: touch;
+  }
```
bpierre: Is it inherited by children elements?

If yes, maybe we can move this in the `html` selector? And if not, to the `*, *:before, *:after` selector?

AquiGorka: You are right, no need to add it to both `html` and `body` as it is inherited and `html` is enough.

sohkai:
screen shot 2018-12-02 at 3 37 38 pm

Oh CSS 😄

Issue #263, opened 2018-12-04
Stopwatch support for countdown


josx:
I am working on Payroll (an aragon future-app ) and using aragon-ui components (BTW great work!)

1) I have already created a new component called Timer that has the same logic that countdown, but it is a Stopwatch. (We need it because we need to show "Time since last XX" )
https://github.com/protofire/aragon-ui/commit/72d660c884b61de6dccbee635f9c963e7e759baa

2) In the other hand I have changed Countdown to suppor a new prop called: ```start```, if you send it, countdown will work as a stopwatch since date on start prop. (without loosing backwards compatibility)
https://github.com/protofire/aragon-ui/commit/7e8da33ccf52a3efa1f1808bb0e49ce79e24c7ba

### I would make a PR, but what do you prefer 1) or 2)

bpierre:
Hi @josx! That’s great, we were just talking about it earlier this week with @drcmda, who is working on the notifications panel for the Aragon client: https://github.com/aragon/aragon/pull/495

For the PR, I would rather go for 2), renaming Countdown to something else. We could do something like this:

- Create a `Timer` component (or `Duration`?), supporting dates in the past and the future.
- We should keep the `CountDown` component for compatibility, but it would do nothing else than using the new component.

josx:
Ok, I would refactor to have a new component that supports both behaviours and also keep backwards compatibility with Countdown component.

bpierre:
Awesome, thanks! 👍

drcmda:
That would be awesome, i would use it for notifications when it's ready.

josx:
@bpierre Now you can review the PR ;)

sohkai:
Closed by #264.

PR #264, opened 2018-12-05
Add Timer component

josx:
After some chat with @bpierre  on #263, I followed his advice on add a new component called Timer that support both start or end as prop (stopwatch or countdown features). At the same I have kept backwards compat with current Countdown component (that just use Timer internally).

### DIFF
--- a/src/components/Timer/Timer.js
+++ b/src/components/Timer/Timer.js
@@ -1,6 +1,7 @@
 import React from 'react'
 import styled from 'styled-components'
 import PropTypes from 'prop-types'
+import memoize from 'lodash-es/memoize'
 import { Redraw } from '../../providers/Redraw'
 import IconTime from '../../icons/components/Time'
 import { difference, formatHtmlDatetime } from '../../utils/date'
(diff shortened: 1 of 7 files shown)


josx:
I dont know why travis show as pending, and when you enter and view the details **says passed in green**.

src/components/Timer/Timer.js
```diff
@@ -0,0 +1,115 @@
...
+
+const FRAME_EVERY = 1000 / 30 // 30 FPS is enough for a ticker
+
+const formatUnit = v => String(v).padStart(2, '0')
+
+class Timer extends React.Component {
+  static propTypes = {
+    start: PropTypes.instanceOf(Date),
+    end: PropTypes.instanceOf(Date),
+    removeDaysAndHours: PropTypes.bool,
```
bpierre: Since we are moving to a new component, what would you think of having a system a bit more flexible than `removeDaysAndHours`, that would allow to:

- Hide units on the left side (e.g. days + hours).
- Hide units on the right side (e.g. seconds).
- Automatically hide units on the left side (e.g. `00d 04h 07m 11s` would become `04h 07m 11s`).

Maybe using something like `format`, that would allow these kind of examples:

- `"dhms"`: display everything.
- `"dhm"`: hide the seconds.
- `"hm"`: display the hours and minutes, even if there are more hours than 24.
- `"auto"`: automatically hide 0 units on the left side.

bpierre:
@josx I added commits to sync with `master` + add a `devbox/` demo 👍

josx:
@bpierre your feature request about format it is done and working ;)

bpierre:
@josx Thank you, super clean! 💫

I added a few things:

- Add two formats: "m" and "s".
- Add a `showEmpty` prop: when set to `false` (default), the units on the left side of the timer will be automatically hidden if they are equal to zero.
- Adjust the time depending on the format, so that the indicated time is always correct. For example, “1D 3H 10M” becomes “27H 10M” when days are not displayed.

PR #266, opened 2018-12-07
devbox: move from parcel to webpack

bpierre:
On my laptop, it makes the build twice as fast, and live reload works.

### DIFF
--- a/devbox/webpack.config.js
+++ b/devbox/webpack.config.js
@@ -1,0 +1,99 @@
+const path = require('path')
+const HtmlWebpackPlugin = require('html-webpack-plugin')
+const CleanWebpackPlugin = require('clean-webpack-plugin')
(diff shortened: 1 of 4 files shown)


devbox/webpack.config.js
```diff
@@ -0,0 +1,100 @@
+const path = require('path')
+const HtmlWebpackPlugin = require('html-webpack-plugin')
+const CleanWebpackPlugin = require('clean-webpack-plugin')
+const { BundleAnalyzerPlugin } = require('webpack-bundle-analyzer')
+const UglifyJsPlugin = require('uglifyjs-webpack-plugin')
```
sohkai: I think we can remove this if we're using Terser?

bpierre: Done 🙈

PR #269, opened 2018-12-11
Add a utility component to emulate :focus-visible

bpierre:
Implements a behavior similar to :focus-visible for browsers that are not supporting it yet.

It follows the Chrome implementation, checking for a pointer device rather than a keyboard event.

Resources:
- https://caniuse.com/#search=%3Afocus-visible
- https://github.com/WICG/focus-visible/issues/88#issuecomment-363227219
- https://chromium-review.googlesource.com/c/chromium/src/+/897002<Paste>

### DIFF
--- a/src/components/FocusVisible/FocusVisible.js
+++ b/src/components/FocusVisible/FocusVisible.js
@@ -1,0 +1,59 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+
+// Implements a behavior similar to :focus-visible for browsers that are not
+// supporting it yet.
+//
+// It follows the Chrome implementation, checking for a pointer device rather
+// than a keyboard event.
+//
+// Resources:
+//  - https://caniuse.com/#search=%3Afocus-visible
+//  - https://github.com/WICG/focus-visible/issues/88#issuecomment-363227219
+//  - https://chromium-review.googlesource.com/c/chromium/src/+/897002<Paste>
+class FocusVisible extends React.Component {
+  static propTypes = {
+    children: PropTypes.func.isRequired,
+    element: PropTypes.instanceOf(Element).isRequired,
+  }
+  state = {
+    focusVisible: false,
+  }
+  componentDidMount() {
+    document.addEventListener('mousedown', this.handlePointerEvent)
+    document.addEventListener('mouseup', this.handlePointerEvent)
+    document.addEventListener('touchstart', this.handlePointerEvent)
+    document.addEventListener('touchend', this.handlePointerEvent)
+  }
+  componentWillUnmount() {
+    document.removeEventListener('mousedown', this.handlePointerEvent)
+    document.removeEventListener('mouseup', this.handlePointerEvent)
+    document.removeEventListener('touchstart', this.handlePointerEvent)
+    document.removeEventListener('touchend', this.handlePointerEvent)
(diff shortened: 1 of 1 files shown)


src/components/FocusVisible/FocusVisible.js
```diff
@@ -0,0 +1,66 @@
...
+  }
+  handleFocus = () => {
+    this.setState({ focusVisible: !this._pointerActive })
+  }
+  handlePointerEvent = e => {
+    this._pointerActive = true
+    this._timer = setTimeout(() => {
+      this._pointerActive = false
+    }, 0)
+    this.setState({ focusVisible: false })
```
sohkai: Do we know the order that the pointer events and focus events are fired? Reading this I assume it's always pointer then focus (which makes sense). Would be nice to leave a comment :).

src/components/FocusVisible/FocusVisible.js
```diff
@@ -0,0 +1,66 @@
...
+// than a keyboard event.
+//
+// Resources:
+//  - https://caniuse.com/#search=%3Afocus-visible
+//  - https://github.com/WICG/focus-visible/issues/88#issuecomment-363227219
+//  - https://chromium-review.googlesource.com/c/chromium/src/+/897002<Paste>
+class FocusVisible extends React.Component {
+  static propTypes = {
+    children: PropTypes.func.isRequired,
+    element: PropTypes.instanceOf(Element),
```
sohkai: Should we make `element` required? This component doesn't seem useful unless you give it an `element` (and I'd consider renaming it to `focusElement` to be clearer).

bpierre:
> There's also a polyfill that uses class names, which you've probably run into.

Yes, heavily inspired from it!

> However, given how this is meant to be used as part of a library, I think it makes sense to scope it as a component (otherwise we'd have to add a FocusVisibleProvider or something weird).

Yes exactly, I think it would be weird to depend on an external provider just to handle focus states. Firefox supports `:focus-visible`, Chrome is really close (behind a flag), and hopefully Safari will follow. When the support will be good enough, we could decide to remove the component altogether and support older browsers by having a visible focus on click (or more precisely, have the different behaviors browsers have for `:focus`).

src/components/FocusVisible/FocusVisible.js
```diff
@@ -0,0 +1,66 @@
...
+  componentDidUpdate(prevProps) {
+    const { element } = this.props
+    if (prevProps.element === element) {
+      return
+    }
+    if (prevProps.element) {
+      prevProps.element.removeEventListener('focus', this.handleFocus)
+    }
+    if (element) {
+      element.addEventListener('focus', this.handleFocus)
```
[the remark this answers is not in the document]
bpierre: I changed the API a bit, there is no need to pass a ref anymore 👍

src/components/FocusVisible/FocusVisible.js
bpierre: I couldn’t find anything in the UI Events spec, it’s only defined within a same group (e.g. [focus events](https://www.w3.org/TR/uievents/#events-focusevent-event-order)). But that’s how Firefox / Chrome / Safari work (and I guess Edge, since that’s how `WICG/focus-visible` is implementing it). Adding a comment 👍

Issue #45, opened 2017-12-06
Add PartitionBar


sohkai:
![image](https://user-images.githubusercontent.com/4166642/33647991-c8621fc2-da25-11e7-89b7-3b0760a87acc.png)
---------------
![image](https://user-images.githubusercontent.com/4166642/33647994-ce4ef6f8-da25-11e7-8e9b-67a81ffd99a6.png)

Location: `src/Bar/PartitionBar.js`

sohkai:
@sistemico We actually already have this component built, but it's not available in aragon-ui yet.

There's actually two (I've clarified in the initial issue) partition bars:

- `PartitionBar` is available as part of the Token Manager app's `SideBar` (named as `StakeBar`): https://github.com/aragon/aragon-apps/blob/master/apps/token-manager/app/src/components/SideBar.js#L61
- `ThresholdBar` is available as part of the Voting app's `VoteSummary` (named as `QuorumBar`): https://github.com/aragon/aragon-apps/blob/master/apps/voting/app/src/components/VoteSummary.js#L45

It'd be best to take these components, better encapsulate them, and add them to aragon-ui :).

cc @bpierre

bpierre:
I renamed the issue since `PartitionBar` has been added by https://github.com/aragon/aragon-ui/pull/270.

PR #270, opened 2018-12-11
PartitionBar component

lmcorbalan:
This PR adds the **PartitionBar** component requested on #45

It includes a caption list under the bar showing a detail for each partition, which can be turned off. By default it is built using the data passed to the component but it also has the ability to receive a custom component  as description for each partition 

##### No caption
image

##### Default caption
image

##### Custom caption
image


src/components/PartitionBar/PartitionBar.js
```diff
@@ -0,0 +1,132 @@
...
+      name: PropTypes.string.isRequired,
+      percentage: PropTypes.number.isRequired,
+    })
+  ).isRequired,
+  colors: PropTypes.arrayOf(PropTypes.string),
+  caption: PropTypes.func,
+}
+
+PartitionBar.defaultProps = {
+  data: [],
```
sohkai: We should probably remove this as a default prop since it's required.

src/components/PartitionBar/PartitionBar.js
```diff
@@ -0,0 +1,132 @@
...
+  align-items: stretch;
+  width: 100%;
+  height: 10px;
+  margin-top: 20px;
+  border-radius: 4px;
+  overflow: hidden;
+`
+
+const Partition = styled.div`
+  flex: ${({ value }) => value};
```
sohkai: ❤️ how browsers automatically do all the resizing magic for us

PR #276, opened 2018-12-14
Transaction badge

josx:
In payroll future app we need to use a Transaction Badge. 
So following IdentityBadge We have done Transaction Badge.

Closes https://github.com/protofire/aragon-apps/issues/90


src/utils/web3.js
```diff
@@ -71,23 +115,17 @@ export function blockExplorerUrl(
...
-  if (provider !== 'etherscan') {
-    warn('blockExplorerUrl(): provider not supported.')
-    return ''
-  }
+  const explorer = BLOCK_EXPLORERS[provider]
 
-  if (!ETHERSCAN_NETWORK_TYPES.includes(networkType)) {
-    warn('blockExplorerUrl(): network type not supported.')
+  if (!explorer) {
+    warn('blockExplorerUrl(): provider not supported.')
```
sohkai: Could be nice to log which provider failed (easier to catch a typo or etc.)

src/utils/web3.js
```diff
@@ -1,8 +1,35 @@
...
+  ['block', 'block'],
+  ['transaction', 'tx'],
+  ['address', 'address'],
+  ['token', 'token'],
+])
+
+const BLOCK_EXPLORERS = {
+  etherscan: ({ type, value, networkType }) => {
+    if (!ETHERSCAN_NETWORK_TYPES.has(networkType)) {
+      throw new Error('provider not supported.')
```
sohkai: Should be 'network not supported'

PR #284, opened 2019-01-22
Add ButtonBase component

bpierre:
Builds on #283 (note: this PR requires to be merged in `babel-runtime` to facilitate its reviewing: it will be changed to `master` later).

This is a simple button component that does nothing except displaying a focus ring. This is useful as a base to build other button types.

We could also make the focus ring a component that could be overridden with a prop, if it becomes needed at some point in the future.


src/components/Button/ButtonBase.js
```diff
@@ -0,0 +1,53 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import styled from 'styled-components'
+import { FocusVisible } from '..'
+import { theme } from '../../theme'
+import { unselectable, font } from '../../utils/styles'
+
+class ButtonBase extends React.PureComponent {
```
AquiGorka: Curious here: why a `PureComponent` instead of a `function`?

src/components/Button/ButtonBase.js
```diff
@@ -0,0 +1,53 @@
...
+  }
+}
+
+const Main = styled.button.attrs({ type: 'button' })`
+  display: inline-block;
+  padding: 0;
+  white-space: nowrap;
+  ${font({ size: 'small', weight: 'normal' })};
+  ${unselectable};
+  color: ${theme.textSecondary};
```
AquiGorka: I think I've needed to modify the color before, maybe this could take in value from a `prop`?

src/components/Button/ButtonBase.js
```diff
@@ -0,0 +1,53 @@
...
+
+  background: none;
+  border: 0;
+  border-radius: 3px;
```
AquiGorka: Should this be modifiable from a `prop` too?

src/components/Button/ButtonBase.js
bpierre: I think in most cases, button’s props are not updated, so it might be a sane choice to skip its rendering everytime its parent re-renders without changing its props.

I now try to use `PureComponent` ([or, when we’ll move to 16.6+, functional components with `memo`](https://reactjs.org/blog/2018/10/23/react-v-16-6.html)) on components that are usually not updated often, and `Component` (or functional components) with or without `shouldComponentUpdate` on components that are often updated.

AquiGorka: :til-emoji:

src/components/Button/ButtonBase.js
bpierre: I wanted to let users customize anything using styled extends, [like I did in `ButtonIcon`](https://github.com/aragon/aragon-ui/pull/285/files#diff-6528b9de4ae18f16135fa568b855e8e6R8), rather than selecting what can be modified. The reason is that it’s a component people will use to build other components, but rarely directly. What do you think?

bpierre:
> Why does this need async/await?

It doesn’t 😆 I branched from the babel runtime branch by mistake, and I left it like this because it was going to be merged at the same time anyway.

src/components/Button/ButtonBase.js
AquiGorka: 💯 agree on the approach. Let's make it super explicit.

PR #286, opened 2019-01-22
Add a selection style

bpierre:
image

### DIFF
--- a/src/components/BaseStyles/BaseStyles.js
+++ b/src/components/BaseStyles/BaseStyles.js
@@ -136,6 +136,10 @@ const GlobalStyle = createGlobalStyle`
   b {
     font-weight: 600;
   }
+  ::selection {
+    color: {theme.textPrimary};
+    background: #B2D7FF;
+  }
 `
 
 export default PublicUrl.hocWrap(BaseStyles)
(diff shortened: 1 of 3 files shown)


bpierre:
> Do we have selectable text with a dark background?

I don’t think we do, but it would work anyway since both text color and background are defined. The only case would be to have exactly the same color as a background before the selection, making the selection invisible. But this color only exists for text selection so it should be fine.

Also this will be customizable with the Lorikeet’s theming system.

src/components/BaseStyles/BaseStyles.js
```diff
@@ -136,6 +136,10 @@ const GlobalStyle = createGlobalStyle`
   b {
     font-weight: 600;
   }
+  ::selection {
+    color: {theme.textPrimary};
+    background: #B2D7FF;
```
sohkai: In the same vein as @AquiGorka, I wonder if we should make these a theme color?

bpierre:
@jounih before this get merged, could you confirm we can use this color everywhere as a selection style?

src/components/BaseStyles/BaseStyles.js
bpierre: I moved it in the theme palette, but the blue is not defined in the base colors: this color doesn’t exist there, and it would have required to create another group for it, while we already have [several blue groups](https://ui.aragon.org/colors/) that are not really used.

I also think that we shouldn’t spend to much time trying to fit new colors to the palettes we have now, as @jounih is working on a new color system for aragonUI (as a Lorikeet theme) that should include everything needed.

PR #288, opened 2019-01-22
Add IconCopy component

bpierre:
<p align=center>image</p>


src/icons/svg/copy.svg
```diff
@@ -0,0 +1 @@
+<svg width="22" height="22" viewBox="0 0 22 22" xmlns="http://www.w3.org/2000/svg"><path d="M12.263 5.806H6.651A1.66 1.66 0 0 0 5 7.457v9.892A1.66 1.66 0 0 0 6.651 19h5.612a1.66 1.66 0 0 0 1.652-1.651V7.457c0-.923-.746-1.651-1.652-1.651zm.994 11.525c0 .55-.443.994-.994.994H6.651a.993.993 0 0 1-.994-.994V7.457c0-.55.444-.994.994-.994h5.612c.55 0 .994.444.994.994v9.874zM16.72 4.65v9.892a1.66 1.66 0 0 1-1.651 1.651.337.337 0 0 1-.338-.337c0-.196.142-.338.338-.338.55 0 .994-.444.994-.994V4.651a.993.993 0 0 0-.994-.994H9.457a.993.993 0 0 0-.994.994.337.337 0 0 1-.338.338.327.327 0 0 1-.337-.338C7.806 3.746 8.552 3 9.458 3h5.61a1.66 1.66 0 0 1 1.652 1.651z" fill="#8B9396" fill-rule="nonzero"/></svg>
```
AquiGorka: Why do we need an independent `svg`? Why didn't you use the `svg` in the Component?
React will probably complain about `fill-rule` -> `fillRule`

bpierre: The SVG files are only used to generate the icon components using [`npm run icons:build`](https://github.com/aragon/aragon-ui/blob/master/package.json#L25) 👍 see [`icons/`](https://github.com/aragon/aragon-ui/tree/master/src/icons)

bpierre: > Why do we need an independent svg?

So we can update these components by modifying any of these files in a normal SVG editor, then update the components automatically from it.

sohkai: @bpierre what do you think about adding `npm run icons:build && npm run optimize:svg && npm run theme:build` on push or publish?

bpierre: Yes good idea!

bpierre: Thinking more about it, I don’t think we should have it on publish: the generated files are versioned and should be present already. If not, it would mean publishing a version on npm that doesn’t correspond to the git tag.

We could have it on push, preventing you to push if there is any change in the generated files, but it might be a bit overkill (it’s a long task).

What about removing these files entirely, and run `npm run icons:build && npm run theme:build` on `prepare`? I think `npm run optimize:svg` should be the responsibility of anyone adding a new SVG file to the repo, and should never be needed afterwards.

@sohkai @AquiGorka thoughts?

sohkai: I think running it on `prepare` makes sense :+1:.

Maybe it'd make sense to run the optimize script on each push or commit, by [configuring `lint-staged` to only run if a certain path is changed](https://github.com/okonet/lint-staged#advanced-config-format).

bpierre: Sounds good yes! I added an issue to do it as a separate task https://github.com/aragon/aragon-ui/issues/295

PR #290, opened 2019-01-22
Popover changes

bpierre:
- Supports a `center` placement.
- The component closing behavior is now based on the focus rather than the click events outside, allowing to close it e.g. when a sandboxed iframe is clicked.
- Opening + closing transition (rather than opening only).
- Throws an error if the component is not wrapped in `<Root.Provider>`.

API changes:

- `openerRef` becomes `opener` (to better reflect that it’s a DOM element and not a React ref)
- `left` and `right` were not playing nicely with Popper.js and have been removed.
- A new prop, `visible` allows to declare the `<Popover />` element without condition, and to let the component handle its opening / close transitions.
- A new placement, `center`, has been added, and centers the `Popover` on the opener element.

This PR also updates `TransactionProgress` and the documentation.


AquiGorka:
🤣🤣 "Appease ESLint’s wrath" what happened there? Why did `ESLint` dislike your initial implementation?

devbox/apps/Popover.js
```diff
@@ -47,19 +50,16 @@ class App extends React.Component {
 }
 
 const Wrapper = styled.div`
-  display: flex;
+  display: grid;
+  width: 800px;
```
AquiGorka: This is not mobile-friendly-first no?

bpierre:
> 🤣🤣 "Appease ESLint’s wrath"
> What happened here? Why did ESLint dislike your initial implementation?

It didn’t like that the `propTypes` were defined outside of the component 💁‍♂️

But I didn’t have a reason to do this anymore, so all good 👌

devbox/apps/Popover.js
bpierre: This is only the demo page here, the size of the component itself can be anything, and is dynamic by default :-)

src/components/Popover/Popover.js
```diff
@@ -12,138 +12,198 @@ import { springs } from '../../utils/styles'
...
-  box-shadow: 0 4px 4px 0 rgba(0, 0, 0, 0.06);
+  filter: drop-shadow(0 4px 4px rgba(0, 0, 0, 0.06));
 `
 
 const Popover = props => (
-  <Root>{el => <PopoverBase rootElement={el} {...props} />}</Root>
+  <Root>
+    {rootElement => {
+      if (!rootElement) {
+        throw new Error('<Popover> need to be nested in <Root.Provider>.')
```
AquiGorka: I think it should read: `'<Popover> needs to be nested in <Root.Provider>.'` with the `s`

src/components/Popover/Popover.js
```diff
@@ -12,138 +12,198 @@ import { springs } from '../../utils/styles'
...
-    return ReactDOM.createPortal(
-      <div
+    const { zIndex, children, transitionStyles, placement } = this.props
+    const { scale, opacity } = transitionStyles
+    const openerHeight = this._openerRect ? this._openerRect.height : 0
+    return (
+      <Main
         ref={this._element}
-        style={{ position: 'absolute', top, left, zIndex }}
+        tabIndex="0"
```
AquiGorka: ❤️ now the `div` can receive focus

PR #291, opened 2019-01-22
RootProvider: move to React.createRef()

bpierre:
It also now doesn’t render its children until `element` is present (at the second rendering), preventing any unexpected behavior in `Root` consumers.


src/providers/Root/Root.js
```diff
@@ -7,16 +7,19 @@ class RootProvider extends React.Component {
...
+  componentDidMount() {
+    this.setState({ element: this._element.current })
   }
   render() {
     const { element } = this.state
     const { children } = this.props
     return (
       <Provider value={element}>
-        <div ref={this.handleRef}>{children}</div>
+        <div ref={this._element}>{element ? children : null}</div>
```
AquiGorka: `{element && <div ref={this._element}>{children}</div>}`?
Thinking about avoiding the empty div in that first render?

bpierre: We have to render the it at least once so that `this._element` gets the reference, which is then used to set `element` and trigger a second render with the children. Rendering an empty `<div>` shouldn’t have any noticeable impact on the performances (except if someone tries to render `<Root />` 60 times per second maybe 😆), so rendering twice is acceptable here IMO.

sohkai: Could we document why we only render `children` only if `element` is present?

It seems intuitive that you wouldn't want `children` to accidentally try and use the provider if the ref hasn't been set yet... but it relies on [knowing the order refs are set in comparison to components mounting](https://reactjs.org/docs/refs-and-the-dom.html#adding-a-ref-to-a-dom-element).

bpierre: I added a comment https://github.com/aragon/aragon-ui/pull/291/commits/c9ee953ee6c1f7357dbab1d036157016559e4b33

PR #303, opened 2019-02-02
Add Viewport

bpierre:
_builds on #305_

# Viewport

image

## Why?

Truly adaptive layouts in CSS is hard. Media queries enabled websites to adapt to the viewport width (at the global level), and `window.matchMedia` / `window.onResize` allow to get this information from JavaScript. But some essential features are still missing to get at the level of other platforms:

- There is no native solution to conditionally apply CSS declarations, based on the dimensions of an element ([“element queries”](https://elementqueries.com/)) rather than the entire viewport.
- Only `window` provides a resize event, not elements themselves (e.g. resizing a sidebar inside of an app wouldn’t trigger a `window.onResize` event).
- There is no unit referring to the parent, the same way we do have `vw` / `vh`. The `%` unit can have this role in certain cases, but it depends on the context, and don’t allow to refer to the parent width from the height property, or vice versa.

Until these problems get solved ([CSS Layout API](https://github.com/w3c/css-houdini-drafts/blob/master/css-layout-api/EXPLAINER.md) 🤞), a common solution is to combine several APIs and techniques:

- Use global CSS Media queries when possible, using a defined set of viewport-based breakpoints.
- Adapt React props and elements rendering, using the same global breakpoints, when possible. When local adaptations are, use custom values rather than the breakpoints.
- Elements dimensions can be measured when the viewport gets resized if needed
- When an element dimensions doesn’t depend on the viewport size, it is possible to keep track of them to calculate the dimensions of other elements (e.g. the [horizontal padding of the sidebar](https://github.com/aragon/aragon-ui/blob/master/src/components/SidePanel/SidePanel.js#L184)).

## What about &lt;BreakPoint />?

We currently have two solutions to adapt apps to the viewport dimensions. [`breakpoint()`](https://github.com/aragon/aragon-ui/blob/master/src/utils/styles/breakpoint.js#L13-L18), which generates CSS media queries, and [`<BreakPoint />`](https://github.com/aragon/aragon-ui/blob/master/src/utils/styles/breakpoint.js#L20-L35), which renders its children based on matched media queries. They were created to be complementary: `breakpoint()` for pure CSS adaptations, and `<BreakPoint />` for components.

`breakpoint()` could be improved (e.g. by supporting custom breakpoints and providing a better API), but it remains useful and efficient as it is. This PR doesn’t change anything about this utility, apart from adding new named breakpoints.

`<BreakPoint />`, on the other hand, was created in the context of a website, and is showing serious limitations in a web app context:

- Because of the way it works, it is not possible to change the value of a prop without unmounting and mounting a component, which can cause a full rendering of an app if anything needs to be adapted near the top level of the app tree.
- Only one media query is supported at a time.
- Only named media queries are supported, making it hard to react to custom breakpoints, or to adapt something based on the dimensions of the viewport.

## A better approach: &lt;Viewport />

`<Viewport />` combines named breakpoints with viewport dimensions, allowing any level of flexibility. Instead of relying on conditional rendering of its children, it provides a render prop that provides a set of values and utilities. A React hook (`useViewport()`) will be added later, providing the same features.

Note: it only relies on `window.onResize`: `width` and `height` are known already, so there is no need for another listener added through `window.matchMedia`.

### How to test

1. Run these commands in your terminal app:

```sh
git clone https://github.com/aragon/aragon-ui.git
cd aragon-ui
git checkout -b windimensions origin/windimensions
npm install
cd devbox && npm install && npm start
```

2. Open http://localhost:1234/#Viewport

### Usage

```jsx
// The window width and height can be used directly if needed.

<Viewport>
  {({ width, height }) => (
    <div>Dimensions: {width}x{height}</div>
  )}
</Viewport>
```

```jsx
// A set of utilities (`above()`, `below()`, `within()`) allow to handle the most
// common use cases in a concise manner. They are also taking care of excluding
// the ending value of a range, ensuring that two breakpoints ranges don’t overlap
// on their starting / ending values.

<Viewport>
  {({ above }) => (
    <App largeMode={above('medium')} />
  )}
</Viewport>


<Viewport>
  {({ below }) => (
    <TextInput wide={below('small')} />
  )}
</Viewport>

<Viewport>
  {({ above }) => (
    <TextInput wide={!above('small')} />
  )}
</Viewport>

// Custom numbers are supported too:
<Viewport>
  {({ above }) => (
    <TextInput wide={below(500)} />
  )}
</Viewport>
```

```jsx
// It still allows for conditional rendering of components.

<Viewport>
  {({ within }) => (
    <div>
      {below('medium') && <SmallVariant />}
      {within('medium', 'large') && <MediumVariant />}
      {above('large') && <LargeVariant />}
    </div>
  )}
</Viewport>
```

```jsx
// Breakpoint values can be used directly too.

<Viewport>
  {({ width, breakpoint }) => (
    <div css={`min-width: ${breakpoint.min}`}>
      <App largeMode={width >= breakpoint.medium} />
    </div>
  )}
</Viewport>
```

#### Width-first API
As with `<BreakPoint />` and `breakpoint()`, its API is focused on the width rather than the height, as it is the main axis used to adapt layouts. `height` is available if needed, but everything else (breakpoints and utilities) is related to the viewport `width`.

#### Rendering frequency

Getting live updates (rather than `matchMedia`) is powerful, but costly: the render function gets called every time the viewport dimensions are updated. It is not a huge performance issue as animations are unlikely to happen during a resize and they don’t happen often, but two solutions can mitigate it.

The first one is to throttle the updates, which `Viewport` does internally: updates only happen every 100ms (by default, can be changed using `throttleWait`).

The second one is that we should be particularly careful of skipping renders when applicable (using the usual `shouldComponentUpdate()` / `React.PureComponent` / `memo()`) on the immediate children of `<Viewport>`. This is something that should be done anyway, so I don’t see it as an issue.

If needed, we could also add a way for `<Viewport>` to declare that we don’t want to get live `width` / `height` updates, but only to listen to breakpoints changes.


src/providers/Viewport/Viewport.js
```diff
@@ -0,0 +1,115 @@
...
+    const { throttleWait } = this.props
+    this._handleResize = throttle(this.updateWindowSize, throttleWait)
+    this.updateWindowSize()
+    window.addEventListener('resize', this._handleResize)
+  }
+
+  resizeStop() {
+    if (this._handleResize) {
+      window.removeEventListener('resize', this._handleResize)
+      delete this._handleResize
```
AquiGorka: Maybe [`cancel`](https://lodash.com/docs/4.17.11#throttle) the throttled function before deleting it?

[required]

src/providers/Viewport/Viewport.js
```diff
@@ -0,0 +1,115 @@
...
+    const { within, above, below } = this
+    return (
+      <Provider value={{ ...windowSize, within, above, below }}>
+        {children}
+      </Provider>
+    )
+  }
+}
+
+const Viewport = props => <Consumer {...props} />
```
AquiGorka: Can you do `const Viewport = Consumer`?

src/providers/Viewport/Viewport.js
```diff
@@ -0,0 +1,115 @@
...
+      ...WINDOW_SIZE_BASE,
+      width,
+      height,
+    }
+  }
+
+  // Check if the current width is between two points.
+  // Accepts a breakpoint string ('small', 'large') or numbers (width in pixels).
+  // `min` is inclusive and `max` is exclusive.
+  within = (...[min, max]) => {
```
AquiGorka: Just wondering if `between` would be a better name, did you choose `within` for any specific reason?

src/utils/styles/breakpoint.js
```diff
@@ -4,8 +4,9 @@ import Media from 'react-media'
 import { css } from 'styled-components'
 
 // These breakpoints values represent minimum screen sizes.
-// Small screen sizes should be targetted by default (mobile first).
-const BREAKPOINTS = {
+export const BREAKPOINTS = {
```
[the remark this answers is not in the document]
bpierre: We could freeze it, but I’m not sure it’s trying to prevent this kind of things would be worth it. If someone decides to mutate imported objects, bad things will inevitably happen 😆

src/providers/Viewport/Viewport.js
bpierre: Ah yes good catch!

src/providers/Viewport/Viewport.js
bpierre: I should definitely add a comment, but this is to prevent a React warning:

React used to add `Provider` to `Consumer` but they are deprecate it using a console warning.

The problem is that if you want to do the same thing yourself, like we do here, React will only detect that `Provider` is attached to `Consumer`, and emit a warning if you use it, even if it hadn’t been attached by React itself.

So this is only to make `Provider` not a direct member of `Consumer` in order to prevent this message. Things should go back to normal as soon as React removes the warning (maybe in the next version?).

devbox/apps/Viewport.js
```diff
@@ -0,0 +1,153 @@
...
+                  overflow-x: auto;
+                `}
+              >
+                <div
+                  css={`
+                    padding: 2px 0;
+                    font-size: 13px;
+                  `}
+                >
+                  Current size: {width}x{height}
```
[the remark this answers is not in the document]
bpierre: Yes, I plan to add this to the documentation (for `above()` / `below()` / `within()`).

src/utils/styles/breakpoint.js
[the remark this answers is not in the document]
bpierre: No risk: apps are running totally isolated from the wrapper and have have their own version of aragonUI, so they can only break themselves 😁

devbox/apps/Viewport.js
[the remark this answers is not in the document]
sohkai: 👍 I get really confused trying to figure out which is inclusive and which is exclusive every time!

src/providers/Viewport/Viewport.js
```diff
@@ -0,0 +1,115 @@
...
+
+  updateWindowSize = () => {
+    this.setState({ windowSize: this.getWindowSize() })
+  }
+
+  getWindowSize() {
+    const { width, height } = windowSize()
+    return {
+      ...WINDOW_SIZE_BASE,
+      width,
```
sohkai: We could just do `return { ...WINDOW_SIZE_BASE, ...windowSize() }` here since we already have this pattern in https://github.com/aragon/aragon-ui/pull/303/files#diff-f30406c493a82295910d4d4e38083f2eR11.

src/providers/Viewport/Viewport.js
```diff
@@ -0,0 +1,115 @@
...
+    this.updateWindowSize()
+    window.addEventListener('resize', this._handleResize)
+  }
+
+  resizeStop() {
+    if (this._handleResize) {
+      window.removeEventListener('resize', this._handleResize)
+      delete this._handleResize
+    }
+    this.updateWindowSize()
```
sohkai: Seems a bit unnecessary to do an update here if the only place we use this function by itself is on unmounting.

src/providers/Viewport/Viewport.js
sohkai: Wondering if just `(min, max)` might be better (or an object options form if we think someone would want to specify `max` without `min`)?

src/providers/Viewport/Viewport.js
```diff
@@ -0,0 +1,115 @@
...
+
+    // Convert breakpoints into numbers
+    if (typeof min === 'string') min = BREAKPOINTS[min]
+    if (typeof max === 'string') max = BREAKPOINTS[max]
+
+    if (typeof min !== 'number') {
+      throw new Error('Viewport: invalid minimum value.')
+    }
+    if (typeof max !== 'number') {
+      throw new Error('Viewport: invalid maximum value.')
```
sohkai: Would be nice to include the value in the error string for easier debugging

src/providers/Viewport/Viewport.js
bpierre: @sohkai 

> Wondering if just `(min, max)` might be better

Ah yes! I moved from using array and I’m not sure why, but I made the args an array to destructure it, rather than using simple parameters haha.

> or an object options form if we think someone would want to specify `max` without `min`

If someone really wants to use `within`, `-1` means “no value”, e.g. `within(300, -1)` or `within(-1, 300)`. But I added `above()` and `below()` for convenience. Do you think we should only have one method with named parameters (object)?

Also what do you think about `within` vs. `between`?

src/providers/Viewport/Viewport.js
```diff
@@ -106,6 +106,8 @@ class ViewportProvider extends React.Component {
   }
 }
 
+// React emits a warning message if `Provider` is attached to `Consumer`, this
```
AquiGorka: ❤️

src/providers/Viewport/README.md
```diff
@@ -0,0 +1,130 @@
...
+      {within('medium', 'large') && <div>medium</div>}
+      {above('large') && <div>large</div>}
+    </div>
+  )}
+</Viewport>
+```
+
+## Properties
+
+### `throttleWait`
```
AquiGorka: Maybe just `throttle`?

bpierre: Yes good idea, I avoided `throttle` because it sounded more like a boolean, but I agree it does look nicer. 👍

src/providers/Viewport/Viewport.js
sohkai: Good point, I really like the `above()` and `below()` shortcuts.

No real preference on `within` and `between`, but perhaps `within` is better as it's very different from `below`. `between` also suggests an open interval (`(a, b)`) whereas `within` suggests a closed interval (`[a, b]`) but we have a half-closed one 🤷‍♂️.

PR #304, opened 2019-02-02
Update RadioList.md

PeterMPhillips:
onChange returns an event object whereas onSelect returns the index

### DIFF
--- a/src/components/Radio/RadioList.md
+++ b/src/components/Radio/RadioList.md
@@ -34,7 +34,7 @@ class App extends Component {
         description="You have two options:"
         items={items}
         selected={this.state.selected}
-        onChange={this.handleChange}
+        onSelect={this.handleChange}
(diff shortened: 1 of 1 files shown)


bpierre:
Hi @PeterMPhillips, thanks for the PR! 🤗

The API of `RadioGroup` and `RadioList` has changed since [`0.28.0`](https://github.com/aragon/aragon-ui/releases/tag/v0.28.0): [`onChange`](https://github.com/aragon/aragon-ui/blob/master/src/components/Radio/RadioList.js#L18) is now receiving the item identifier (which can be its index). You must be using `0.27.0` or lower?  If that’s the case, I encourage you to upgrade to newer versions, there are several new components that could be quite useful! https://github.com/aragon/aragon-ui/releases

I am closing this, please let me know if I missed something.

PR #305, opened 2019-02-02
Add Main

bpierre:
_builds on https://github.com/aragon/aragon-ui/pull/303_

# Main

## Features

- Rename `AragonApp`, to make it more descriptive (`AragonApp` still works but emits a deprecation warning).
- Remove the default styles that `AragonApp` was applying.
- Provide defaults for every prop. If the assets are served on `/aragon-ui/`, which is the convention on Aragon Apps, nothing need to be changed.
- Set the providers required by other components:
  - `<Root.Provider />`
  - `<Viewport.Provider />`
  - `<PublicUrl.Provider />`
  - `<BaseStyles />`
  - `<ToastHub />`

## Usage

```jsx
import { Main } from '@aragon/ui'

const App = () => (
  <Main>
    {/* aragonUI components can be used now */}
  </Main>
)
```


src/components/AragonApp/AragonApp.js
```diff
@@ -38,6 +39,11 @@ class AragonApp extends React.Component {
   }
 
   render() {
+    if (!AragonApp._warned) {
+      warn('<AragonApp /> is deprecated: please use <Main /> instead.')
```
AquiGorka: This only warns in dev mode, should we always show the warning to make it more explicit?

src/components/Main/Main.js
```diff
@@ -0,0 +1,35 @@
...
+    legacyFonts: false,
+    assetsUrl: '/aragon-ui/',
+  }
+  render() {
+    const { children, assetsUrl, legacyFonts } = this.props
+    return (
+      <Root.Provider>
+        <Viewport.Provider>
+          <PublicUrl.Provider url={ensureTrailingSlash(assetsUrl)}>
+            <BaseStyles enableLegacyFonts={legacyFonts} />
```
AquiGorka: Is there a case where someone would want to not use the base styles? or how can they override base styles to use their own theme with this approach?

src/components/Main/Main.js
```diff
@@ -0,0 +1,35 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import { ensureTrailingSlash } from '../../utils/url'
+import { PublicUrl } from '../../providers/PublicUrl'
+import { Root } from '../../providers/Root'
+import { Viewport } from '../../providers/Viewport'
+import BaseStyles from '../BaseStyles/BaseStyles'
+import { ToastHub } from '../ToastHub/ToastHub'
+
+class Main extends React.Component {
```
AquiGorka: If you are not using lifecycle methods does this really need to be a class rather than a function?

src/components/AragonApp/AragonApp.js
bpierre: I don’t think we should display any warning in production code, as we are targeting the final user. The same way errors should be managed by the UI in production (rather than the console).

Ideally, we should even remove entire chunks of code related to development, like the prop types or the code managing the warnings. But it’s a bit complicated with JS libraries, because bundlers don’t really have a common standard to import a development (with all the debug artifacts) vs. a production build of a library, depending on the app environment.

AquiGorka: I agree with what you said yet IMO letting users know their `production` build is running a `deprecated` Component is important

src/components/Main/Main.js
bpierre: Themes will be supported when the move to Lorikeet happen (see [Theme.js](https://github.com/lorikeetui/lorikeet/blob/master/src/components/Theme/Theme.js)).

About `BaseStyles`, they are required to make our components work and they would break without them. People can always decide to override them, or to define a custom font, but there is no guarantee that things would keep working properly.

I would like to make some providers explicitly required by the components that need them, and emit errors if not. The same way `Toast` doesn’t work without `ToastHub`, it also doesn’t work without `BaseStyles`. `Main` is a step forward (we only ask users to define a single component), but I think we should warn if users forget to set these base components.

src/components/AragonApp/AragonApp.js
bpierre: Do you have a specific case in mind where you think it might be a problem?

I think the development build is already warning developers about the deprecation, and end users shouldn’t have to see or even know about it, as it doesn’t impact them. Deprecated doesn’t mean broken, it means developers consuming the library have some time to migrate.

For example, an app development team might decide to use a deprecated feature because they don’t have the time to migrate right away. For example, React doesn’t emit any warning in production, which is pretty standard behavior AFAIK: the console represents the app at this point, not its libraries.

src/components/Main/Main.js
sohkai: Maybe we could leave a prop where someone could use their own `createGlobalStyle()` object?

src/components/Main/Main.js
```diff
@@ -0,0 +1,35 @@
...
+    assetsUrl: '/aragon-ui/',
+  }
+  render() {
+    const { children, assetsUrl, legacyFonts } = this.props
+    return (
+      <Root.Provider>
+        <Viewport.Provider>
+          <PublicUrl.Provider url={ensureTrailingSlash(assetsUrl)}>
+            <BaseStyles enableLegacyFonts={legacyFonts} />
+            <ToastHub>{children}</ToastHub>
```
sohkai: Curious if we should add the styles from [`StyledAragonApp`](https://github.com/aragon/aragon-ui/blob/5d68042e5c/src/components/AragonApp/AragonApp.js#L11) here?

src/components/Main/Main.js
bpierre: The way I was seeing it, `Main` would give you the most basic configuration options, and everything more advanced could be achieved by composing the providers yourself.

I think we should avoid encouraging people to replace `BaseStyles` with their own, because these styles are essential to the components to work properly. By doing so (e.g. duplicating its styles in a custom `BaseStyles`), components might break in subtle ways in future upgrades, depending on how the global styles evolve.

I much prefer to let users extend the base styles if they need it (e.g. for fonts or colors), which they can do already by using `createGlobaStyle()` anywhere in their app. If someone really want to get rid of `BaseStyles`, or any other custom setup, it will always be possible to declare the providers yourself (but in that case, you’ll have to follow how `Main` evolves).

What do you think?

bpierre:
This PR was closed by inadvertence, see https://github.com/aragon/aragon-ui/pull/306 for the updated one.

src/components/Main/Main.js
AquiGorka: If anyone wants to override styles and a) we have not yet offered some mechanism to do it or b) we chose not to offer it they can always use their own "`Main`" with the Providers and avoid using the `BaseStyles`. If we see such behavior we could come back to this no?

I'm 👌 with the current implementation

bpierre: > If we see such behavior we could come back to this no?

Of course, we can add the prop if it appears to be needed 👍

sohkai: Sounds good! Didn't think that users would be able to use `createGlobalStyles` themselves so that's a good way to extend it 👍.

PR #310, opened 2019-02-12
Add documentation for the Main component

bpierre:
_Builds on #309_


bpierre:
> Do you normally update the package-lock.jsons on publish or when the dependencies are upgraded?

Oh yes I forgot that, I’m doing it in the other PR.

PR #311, opened 2019-02-13
Responsive common components

AquiGorka:
So far these changes include:

- Menu icon
- `MenuButton` component (using `ButtonIcon`)
- Uses `FocusVisible` on `MenuButton`
- Update to `NavigationBar` for responsive
- Plus icon
- `PlusButton` component (using `ButtonIcon`)
- Arrow right icon
- `ArrowRightButton` component (using `ButtonIcon`)
- Borderless `TableCell`

Refs:
- [`ButtonIcon`](https://github.com/aragon/aragon/pull/539#discussion_r250502770)
- [`FocusVisible`](https://github.com/aragon/aragon-apps/pull/606#discussion_r248661009)
- [`PlusButton`](https://github.com/aragon/aragon/pull/583/files#r250715297)
- [`ArrowRightButton`](https://github.com/aragon/aragon/pull/583/files#r250714912)


src/components/NavigationBar/NavigationBar.js
```diff
@@ -7,8 +7,11 @@ import LeftIcon from './LeftIcon'
 
 class NavigationBar extends React.Component {
   static propTypes = {
-    onBack: PropTypes.func,
+    backPaddingLeft: PropTypes.number,
```
bpierre: I would rather avoid having this kind of props here, this component is meant to be high level and shouldn’t require users to set the spacing manually. Maybe we could have a way to inject the menu button before, and adjust the spacing accordingly? e.g. a `beforeBackButton` prop.

src/components/Table/TableCell.js
```diff
@@ -3,39 +3,45 @@ import PropTypes from '../../proptypes'
...
 
-  /* First and last cell styling */
-  &:first-child {
-    border-left: 1px solid ${contentBorder};
-  }
-  &:last-child {
-    border-right: 1px solid ${contentBorder};
-  }
+  ${breakpoint(
+    'medium',
```
bpierre: We could have a prop to enable this (`noSideBorders` or something like this), but we can’t have this as a default: users might not want this behavior, or use tables in another context than in the main screen.

src/components/ArrowRightButton/ArrowRightButton.js
```diff
@@ -0,0 +1,30 @@
+import React from 'react'
```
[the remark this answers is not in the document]
AquiGorka: I think I hear you, I'll move `ArrowRightButton`, `PlusButton` and `MenuButton` to `aragon/aragon` instead of having this here.

src/components/NavigationBar/NavigationBar.js
AquiGorka: My concern here is that the back arrow adds left padding and I believe it shouldn't, I rather remove the padding but I did not want to break backwards functionality thus I added the props.
Thoughts?

bpierre: The reason why there is a left padding is to get a large clickable area:

image

But we should reduce it when the menu is here yes. If we had a `beforeBackButton` prop, we could give an equal amount of padding to both the menu button (on its right side), and the back button (on its left side).

AquiGorka:
Closing in favor of independent PRs:
- [`MenuButton`](https://github.com/aragon/aragon-ui/pull/318)
- [`Plus` and `Arrow right` icons](https://github.com/aragon/aragon-ui/pull/319)
- [Borderless `TableCell`](https://github.com/aragon/aragon-ui/pull/320)
- [`AppBar` paddings](https://github.com/aragon/aragon-ui/pull/321)
- [`NavigationBar` paddings](https://github.com/aragon/aragon-ui/pull/322)

PR #312, opened 2019-02-14
ButtonIcon: require a label prop

bpierre:
To make sure the button is always described with text.

### DIFF
--- a/src/components/Button/ButtonIcon.js
+++ b/src/components/Button/ButtonIcon.js
@@ -1,10 +1,13 @@
 import React from 'react'
+import PropTypes from 'prop-types'
 import ButtonBase from './ButtonBase'
 
 class ButtonIcon extends React.PureComponent {
   render() {
+    const { label, children, ...props } = this.props
     return (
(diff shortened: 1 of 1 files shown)


src/components/Button/ButtonIcon.js
```diff
@@ -1,10 +1,13 @@
 import React from 'react'
+import PropTypes from 'prop-types'
 import ButtonBase from './ButtonBase'
 
 class ButtonIcon extends React.PureComponent {
   render() {
+    const { label, children, ...props } = this.props
     return (
       <ButtonBase
+        title={label}
```
AquiGorka: Would it make sense to call this prop `title` instead of `label`?

bpierre: The idea is to use `label` every time we need an alternative text in the aragonUI components, no matter how the internal implementation is (`title` here, but it could be `alt` for an `img` or something else). We are also already using `label` as a prop for other components (e.g. `Field`).

PR #317, opened 2019-02-19
TabBar: rename onSelect to onChange

bpierre:
Also add documentation and move to `React.createRef()`.

### DIFF
--- a/devbox/apps/TabBar.js
+++ b/devbox/apps/TabBar.js
@@ -31,7 +31,7 @@ class TabBarDemo extends React.Component {
     const { index } = this.state
     const { items } = this.props
     return (
-      <TabBar onSelect={this.handleSelect} selected={index} items={items} />
+      <TabBar onChange={this.handleSelect} selected={index} items={items} />
     )
   }
 }
--- a/src/components/TabBar/README.md
+++ b/src/components/TabBar/README.md
@@ -1,0 +1,43 @@
+# TabBar
+
+A simple tabs component.
+
+## Usage
+
+```jsx
+import { useState } from 'react'
+import { TabBar } from '@aragon/ui'
+
+const App = () => {
+  const [selected, setSelected] = useState(0)
+  return (
+    <TabBar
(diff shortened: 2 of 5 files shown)


devbox/apps/TabBar.js
```diff
@@ -31,7 +31,7 @@ class TabBarDemo extends React.Component {
     const { index } = this.state
     const { items } = this.props
     return (
-      <TabBar onSelect={this.handleSelect} selected={index} items={items} />
+      <TabBar onChange={this.handleSelect} selected={index} items={items} />
```
AquiGorka: This is a breaking change no? We should update the apps using `TabBar`

bpierre: No it won’t break, the plan is to support it for a few versions before removing it: https://github.com/aragon/aragon-ui/pull/317/files#diff-0bebd46dc3c0aa731b1bc633cafa3777R51

src/components/TabBar/README.md
```diff
@@ -0,0 +1,43 @@
...
+A simple tabs component.
+
+## Usage
+
+```jsx
+import { useState } from 'react'
+import { TabBar } from '@aragon/ui'
+
+const App = () => {
+  const [selected, setSelected] = useState(0)
```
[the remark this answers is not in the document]
bpierre: Yes let’s start using hooks in the gallery demos 👍

They are so much better than all the `state = {}` / `handleX` / `setState()` / `const { X } = this.state` noise for code snippets!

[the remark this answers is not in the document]
bpierre: Done https://github.com/aragon/aragon-ui/pull/317/commits/de61ba0af90373308b8290e2a082a2f800a107b5

devbox/apps/TabBar.js
AquiGorka: Still, when we publish this it should bump the interface flag as this will make components using `onChange` to stop working.

PR #318, opened 2019-02-20
Add IconMenu component

AquiGorka:


bpierre:
I’m wondering if it’s worth adding this button to aragonUI, since we are discussing moving it outside of apps? Maybe we could only have the icon for now?

AquiGorka:
Done. Still made sense to me as every app out there is currently using this.

bpierre:
> Still made sense to me as every app out there is currently using this.

It is more about trying to not add things that we might want to remove soon. Once a new component is added, we have to either maintain it or go through a deprecating process: aragonUI is not only used by us 😄

bpierre:
Could you also add the same changes I added in https://github.com/aragon/aragon-ui/pull/319#pullrequestreview-205901071?

- Run `optimize:svg` to optimize the SVG file.
- Use `#8B9396` for monochromatic icons.
- Add the icon to the gallery.

Thanks!

sohkai:
> Done. Still made sense to me as every app out there is currently using this.

I think we'll have a shared react library in `aragon-apps` for these types of components, that don't necessarily belong in aragonUI.

bpierre:
> I think we should have a shared react library in `aragon-apps` for these types of components, that don't necessarily belong in aragonUI.

Yes it could be nice to have a shared library there. I think that everything that is being used by several of our apps should eventually end up in `aragonUI` (and I start to think that after the Lorikeet move, aragonUI should also be aware of aragon.js), but it would be a great solution as a buffer until we figure out APIs − rather than copying things in every app.

sohkai:
Yes, I see the very gradual transition of components from `individual app` -> `aragon-apps` -> `aragonUI` -> `lorikeet`.

There are also all sorts of little frontend utils that we use and could make shared in the apps but don't really belong in aragonUI.

bpierre:
> Yes, I see the very gradual transition of components from `individual app` -> `aragon-apps` -> `aragonUI` -> `lorikeet`.

Exactly how I see it 👌, with Aragon-specific things staying in aragonUI.

> There are also all sorts of little frontend utils that we use and could make shared in the apps but don't really belong in aragonUI.

We’ll discuss these but it’s probably the same things I would like to see in aragonUI, once it moves from a standalone UI library to an opinionated library to build Aragon apps.

For example, I would like to abstract the aragon.js connection (in `src/index.js` of our apps). It would expand the scope of aragonUI (rather than having e.g. a aragon.js-react library), but I don’t see it as a problem: in the end, everything we do here is about the UI. To be more precise, I think the API should still let you inject aragon.js, to be consistent with the worker that uses aragon.js directly, but could handle the rest:

```jsx
import Aragon, { providers } from '@aragon/client'
import { Main, useAragonJsApp } from '@aragon/ui'

const App = () => {
  const { appState, accounts, network } = useAragonJsApp(
    // it would also be nice to assume a default provider in `new Aragon()` by detection,
    // so we don’t have to inject it 😁
    new Aragon(
      new providers.WindowMessage(window.parent)
    )
  )

  return (
    <Main>
      {/* your app */}
    </Main>
  )
}
```

We could also make aragonUI a monorepo (or make aragonUI part of a bigger monorepo, maybe with aragon.js), so that e.g. a `formatting-utils` module would be published both independently (e.g. `@aragon/formatting-utils`) and as part of aragonUI, but would only be in one place in the codebase.

AquiGorka:
@bpierre updated & optimized as requested

src/icons/components/Menu.js
```diff
@@ -0,0 +1,7 @@
+import React from 'react'
+
+const Menu = props => (
+  <svg width={18} height={14} viewBox="0 0 18 14" {...props}><path d="M1 1h16M1 7h16M1 13h16" stroke="#8B9396"/></svg>
```
bpierre: That’s weird, `#8B9396` should have been replaced by `icons:build` 🤔

Did you run it again? https://github.com/aragon/aragon-ui/blob/master/package.json#L25

AquiGorka: Right, so, no. Had no idea there where 2 scripts for this. TIL about `icons:build` and `optimize:svg`

PR #322, opened 2019-02-20
NavigationBar: update spacing + add documentation

AquiGorka:
This change keeps the alignments in sync for small screens.

### DIFF
--- a/src/components/NavigationBar/NavigationBar.js
+++ b/src/components/NavigationBar/NavigationBar.js
@@ -9,10 +9,12 @@ class NavigationBar extends React.Component {
   static propTypes = {
     onBack: PropTypes.func,
     items: PropTypes.arrayOf(PropTypes.node),
+    compact: PropTypes.bool,
   }
   static defaultProps = {
     onBack: () => {},
     items: [],
+    compact: false,
   }
   state = {
     cachedItems: null,
@@ -29,7 +31,7 @@ class NavigationBar extends React.Component {
     }
   }
   render() {
-    const { onBack, items } = this.props
+    const { onBack, items, compact } = this.props
     const displayedItems = items
       .map((node, index) => ({ node, index }))
       .slice(-1)
@@ -55,6 +57,7 @@ class NavigationBar extends React.Component {
               label={item.node}
               onBack={onBack}
               displayBack={item.index > 0}
+              compact={compact}
               {...styles}
             />
           )}
@@ -64,7 +67,7 @@ class NavigationBar extends React.Component {
   }
 }
 
-const Item = ({ opacity, position, displayBack, onBack, label }) => (
+const Item = ({ opacity, position, displayBack, onBack, label, compact }) => (
   <animated.span
     style={{
       display: 'flex',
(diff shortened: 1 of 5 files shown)


src/components/NavigationBar/NavigationBar.js
```diff
@@ -115,7 +115,6 @@ const Title = styled.span`
 `
 
 const Label = styled.span`
-  padding-left: 30px;
```
sohkai: Wouldn't we still need this at `20px` in case the `BackButton` isn't rendered? Maybe it makes sense to move the `padding-left` to the `Title` and then only the `BackButton` needs to define a `padding-right`?

AquiGorka: When the back button isn't rendered there are two scenarios:

1. `MenuButton` - it takes care of its own padding
1. Nothing to render - the app adds the missing 10px to align the Title.

bpierre: Is it supposed to work with https://github.com/aragon/aragon-ui/pull/322?

It breaks the layout as it is (see the NavigationBar demo):

image

AquiGorka: In this case I can update the demo

bpierre: @AquiGorka What I meant is that I don’t think the default should break?

AquiGorka: Right, so, I'd like us (@sohkai and @bpierre) to define the default value we are using (mobile first I think 20 should be used and medium to larger 30 - similar to the apps app layout [here](https://github.com/aragon/aragon-apps/pull/673/files#diff-2e20014b3077a7461c11aabfbe2c53abR62))

bpierre: Yes that sounds good to me! Let’s try these and see how it looks with the arrow (maybe we’ll have to tweak a bit the spacing between the arrow and the label).

AquiGorka: It has already been implemented [here](https://github.com/aragon/aragon/pull/615)

Sneak peek:
alligned arrow

AquiGorka: @sohkai and @bpierre changes approved?

AquiGorka:
@bpierre you removed the original changes https://github.com/aragon/aragon-ui/pull/322/commits/dce9396863bb28df3fce071d1671ad3f13b6919a, without that change the added padding in the `arrow` button in navigation bar does not keep the alignment.

bpierre:
@AquiGorka yes I pinged you on Rocket Chat about it! The idea I wanted to discuss with you would be to move this exception outside of the component, until we officially support a menu button (if we ever do).

AquiGorka:
@bpierre your suggestion means any usage of `NavigationBar` with `MenuButton` has to remove margin to correctly align for mobile?

bpierre:
@AquiGorka Yes, or more specifically having this inside of `MenuButton`, since it is meant to be used there. Then `MenuButton` can be used normally (see [the demo](https://github.com/aragon/aragon-ui/pull/322/files#diff-2f96ba47345dec33577055cef9126647R50)). Another solution would be to make `NavigationBar` aware of having a `MenuButton` specifically, but since we don’t have this component in aragonUI and might not have it, it feels a bit weird to have an exception for it already.

PR #323, opened 2019-02-20
AppBar + AppView: TabBar integration

bpierre:
# AppBar & AppView tabs

image

A `tabs` prop has been added to both `AppBar` and `AppView` to integrate `TabBar`.

## Usage

```jsx
// `tabs` can be set on `AppBar`…

const SCREENS = ['Home', 'Settings', 'About']

const App = () => {
  const [screen, setScreen] = useState(0)
  return (
    <AppView
      appBar={
        <AppBar
          title="My App"
          tabs={
            <TabBar
              items={SCREENS}
              selected={screen}
              onSelect={setScreen}
            />
          }
        />
      }
    >
      <h1>{SCREENS[screen]}</h1>
    </AppView>
  )
}
```

```jsx
// …or directly on `AppView`.

const SCREENS = ['Home', 'Settings', 'About']

const App = () => {
  const [screen, setScreen] = useState(0)
  return (
    <AppView
      title="My App"
      tabs={
        <TabBar
          items={SCREENS}
          selected={screen}
          onSelect={setScreen}
        />
      }
    >
      <h1>{SCREENS[screen]}</h1>
    </AppView>
  )
}
```


src/components/AragonApp/AppView.js
```diff
@@ -16,12 +16,13 @@ class AppView extends React.Component {
     children: PropTypes.node,
     padding: PropTypes.number,
     height: PropTypes.string,
+    tabs: PropTypes.element,
```
sohkai: Could we do something like `PropTypes.instanceOf(TabBar)`?

src/components/AragonApp/AppBar.js
```diff
@@ -40,33 +85,12 @@ const StyledAppBarTitle = PublicUrl.hocWrap(styled.h1`
...
-          )}
-        </StyledAppBarTitle>
-      </StyledAppBarStart>
-    )}
-    {children}
-    {endContent && <StyledAppBarEnd>{endContent}</StyledAppBarEnd>}
-  </StyledAppBar>
-)
-
 AppBar.propTypes = {
```
sohkai: Could we move this `propTypes` and `defaultProps` declarations to be with the component?

src/components/AragonApp/AppBar.js
```diff
@@ -8,28 +8,73 @@ import { unselectable } from '../../utils/styles'
...
+      ${unselectable()};
+    `}
+  >
+    <div
+      css={`
+        display: flex;
+        align-items: center;
+        justify-content: flex-start;
+        width: 100%;
+        height: 63px;
```
sohkai: Should this be `64px`?

src/components/AragonApp/AppBar.js
```diff
@@ -40,33 +85,12 @@ const StyledAppBarTitle = PublicUrl.hocWrap(styled.h1`
...
-    {endContent && <StyledAppBarEnd>{endContent}</StyledAppBarEnd>}
-  </StyledAppBar>
-)
-
 AppBar.propTypes = {
   children: PropTypes.node,
   title: PropTypes.node,
   endContent: PropTypes.node,
   onTitleClick: PropTypes.func,
+  tabs: PropTypes.element,
```
sohkai: Might as well add this to `AppBar.md` too :)

src/components/AragonApp/AppView.js
bpierre: I don’t think there is any good solution for that… apparently the element created by React is not an instance of the component (in the JavaScript sense).

There is a solution here using `type`: https://github.com/facebook/react/issues/2979#issuecomment-349181415

But even here, checking for `TabBar` explicitly would prevent a component extending it to be accepted :-(

src/components/AragonApp/AppBar.js
bpierre: Nope, the border takes `1px` 😄

src/components/AragonApp/AppBar.js
bpierre: Oh yes thanks!

src/components/AragonApp/AppBar.js
AquiGorka: Aren't we `box-sizing` everything?

src/components/AragonApp/AppView.js
sohkai: ☹️ ☹️ ☹️

src/components/AragonApp/AppBar.js
sohkai: Let's use some constants and math for this :)

bpierre: We are, but the border is on the parent: https://github.com/aragon/aragon-ui/pull/323/files/2333965c1fd95a7e772205463cde0d86145f8487#diff-8eafad76c274d5f737bd3f98e1e51fe8R26

bpierre: Done https://github.com/aragon/aragon-ui/pull/323/commits/4bcd1bdfc5d359ad205338007c0134b314098843

bpierre:
I added a transition:



Also the TabBar is now aware of being in the AppBar, and decides itself to not display its border. This will make things easier for styling (e.g. a theme could have a different bar color for the two things).

.eslintrc
```diff
@@ -21,7 +21,11 @@
...
   "plugins": ["prettier", "react", "import", "promise"],
   "rules": {
-    "import/no-unresolved": ["error", { ignore: ["^react(-dom)?$", "^styled-components$"] }],
-    "linebreak-style": ["error", "unix"]
+    "import/no-unresolved": [
+      "error",
+      { "ignore": ["^react(-dom)?$", "^styled-components$"] }
+    ],
+    "linebreak-style": ["error", "unix"],
+    "react/jsx-boolean-value": "off"
```
bpierre: This is to allow us to pass booleans to props when it makes sense semantically, like here: https://github.com/aragon/aragon-ui/pull/323/files#diff-8eafad76c274d5f737bd3f98e1e51fe8R48

src/components/AragonApp/AppBar.js
```diff
@@ -11,71 +13,123 @@ import chevronSvg from './assets/chevron.svg'
...
-        display: flex;
-        align-items: center;
-        justify-content: flex-start;
-        width: 100%;
-        height: ${BAR_HEIGHT - 1}px;
-      `}
-      {...props}
-    >
-      {title && (
+export const InAppBarContext = React.createContext(false)
```
sohkai: What do you think about attaching the consumer of this to `AppBar` so users of the library can have access to it if they need to define something custom?

src/components/AragonApp/AppBar.js
```diff
@@ -11,71 +13,123 @@ import chevronSvg from './assets/chevron.svg'
...
+          >
+            {tabs =>
+              tabs &&
+              (styles => (
+                <TabsWrapper style={styles}>
+                  <div
+                    ref={this._tabsRef}
+                    css={`
+                      /*margin-bottom: -1px;*/
+                    `}
```
sohkai: Should this bit of `css` be removed?

bpierre: Oops 😅

src/components/AragonApp/AppBar.js
bpierre: Great idea!

.eslintrc
AquiGorka: Curious, is this similar to the use case when we pass down a `node/element`? Now it can be one of `node/element/boolean`?

src/components/AragonApp/AppBar.js
```diff
@@ -1,39 +1,162 @@
...
+    children: PropTypes.node,
+    endContent: PropTypes.node,
+    onTitleClick: PropTypes.func,
+    padding: PropTypes.number,
+    tabs: PropTypes.element,
+    title: PropTypes.node,
+  }
+  static defaultProps = {
+    onTitleClick: noop,
+    padding: 30,
```
AquiGorka: Should this be `20`? [ref](https://github.com/aragon/aragon-apps/pull/673/files#diff-2e20014b3077a7461c11aabfbe2c53abR62)

.eslintrc
bpierre: Ah sorry I pointed to the PR diff instead of a commit, and lines have moved since then. I wanted to point to this line: https://github.com/aragon/aragon-ui/blob/96c05a044a0f56dd9fcccc11b5f11e2fad68680b/src/components/AragonApp/AppBar.js#L64

In that specific case, I want to pass the value `true`, and it would be weird to just do `<InAppBarContext.Provider value>`. I think we already all use shortened props for booleans by default except when we have a good reason to not do it, so I don’t think this ESLint rule is much needed.

sohkai: Although perhaps this asks why is the prop named `value`; could it be given a more descriptive name?

bpierre: This component is a React context provider, so not really 😅

src/components/AragonApp/AppBar.js
bpierre: I think we should leave the default as `30` for now to not break compat, and move it to `0` later, when we’ll make `AppView` a bit smarter (by adding features from [`AppLayout`](https://github.com/aragon/aragon-apps/blob/819de2da7a37edbc401cb1e983ccf75f75c4eff3/apps/voting/app/src/components/AppLayout.js)) and responsible of setting the padding of `AppBar`.

PR #327, opened 2019-02-26
Identity badge custom label action

AquiGorka:
The modal is as raw as it can get:

### DIFF
--- a/src/components/IdentityBadge/IdentityBadge.js
+++ b/src/components/IdentityBadge/IdentityBadge.js
@@ -15,6 +15,7 @@ class IdentityBadge extends React.PureComponent {
     popoverAction: PopoverActionType,
     entity: PropTypes.string,
     fontSize: PropTypes.string,
+    customLabel: PropTypes.string,
     networkType: PropTypes.string,
     shorten: PropTypes.bool,
   }
@@ -40,10 +41,13 @@ class IdentityBadge extends React.PureComponent {
       popoverAction,
       entity,
       fontSize,
+      customLabel,
       networkType,
       shorten,
     } = this.props
     const address = isAddress(entity) ? entity : null
+    const label =
+      customLabel || (address && shorten ? shortenAddress(address) : entity)
 
     return (
       <React.Fragment>
@@ -62,9 +66,7 @@ class IdentityBadge extends React.PureComponent {
                 <EthIdenticon scale={1} address={address} />
               </Identicon>
             )}
-            <Label size={fontSize}>
-              {address && shorten ? shortenAddress(address) : entity}
-            </Label>
+            <Label size={fontSize}>{label}</Label>
           </Main>
         </ButtonBase>
         {address && (
--- a/src/components/IdentityBadge/IdentityBadge.js
+++ b/src/components/IdentityBadge/IdentityBadge.js
@@ -12,11 +12,12 @@ import PopoverActionType from './PopoverActionType'
 class IdentityBadge extends React.PureComponent {
   static propTypes = {
     connectedAccount: PropTypes.bool,
-    popoverAction: PopoverActionType,
+    customLabel: PropTypes.string,
     entity: PropTypes.string,
     fontSize: PropTypes.string,
-    customLabel: PropTypes.string,
     networkType: PropTypes.string,
+    popoverTitle: PropTypes.node,
+    popoverAction: PopoverActionType,
     shorten: PropTypes.bool,
   }
   static defaultProps = {
@@ -39,6 +40,7 @@ class IdentityBadge extends React.PureComponent {
     const {
       connectedAccount,
       popoverAction,
+      popoverTitle,
       entity,
       fontSize,
       customLabel,
@@ -78,6 +80,7 @@ class IdentityBadge extends React.PureComponent {
             opener={this._element.current}
             onClose={this.handleClose}
             popoverAction={popoverAction}
+            title={popoverTitle}
           />
         )}
       </React.Fragment>
--- a/src/components/IdentityBadge/IdentityBadge.js
+++ b/src/components/IdentityBadge/IdentityBadge.js
@@ -7,14 +7,16 @@ import Text from '../Text/Text'
 import EthIdenticon from '../EthIdenticon/EthIdenticon'
 import ButtonBase from '../Button/ButtonBase'
 import IdentityBadgePopover from './IdentityBadgePopover'
+import PopoverActionType from './PopoverActionType'
 
 class IdentityBadge extends React.PureComponent {
   static propTypes = {
+    connectedAccount: PropTypes.bool,
+    popoverAction: PopoverActionType,
     entity: PropTypes.string,
-    shorten: PropTypes.bool,
     fontSize: PropTypes.string,
     networkType: PropTypes.string,
-    connectedAccount: PropTypes.bool,
+    shorten: PropTypes.bool,
   }
   static defaultProps = {
     entity: '',
@@ -34,11 +36,12 @@ class IdentityBadge extends React.PureComponent {
   render() {
     const { opened } = this.state
     const {
+      connectedAccount,
+      popoverAction,
       entity,
-      shorten,
       fontSize,
       networkType,
-      connectedAccount,
+      shorten,
     } = this.props
     const address = isAddress(entity) ? entity : null
 
@@ -72,6 +75,7 @@ class IdentityBadge extends React.PureComponent {
             networkType={networkType}
             opener={this._element.current}
             onClose={this.handleClose}
+            popoverAction={popoverAction}
           />
         )}
       </React.Fragment>
--- a/src/components/IdentityBadge/IdentityBadgePopover.js
+++ b/src/components/IdentityBadge/IdentityBadgePopover.js
@@ -5,28 +5,34 @@ import { theme } from '../../theme'
 import { IconClose } from '../../icons'
 import SafeLink from '../Link/SafeLink'
 import Popover from '../Popover/Popover'
+import Button from '../Button/Button'
 import ButtonIcon from '../Button/ButtonIcon'
 import AddressField from '../AddressField/AddressField'
 import Badge from '../Badge/Badge'
+import PopoverActionType from './PopoverActionType'
 
 class IdentityBadgePopover extends React.PureComponent {
   static propTypes = {
     address: PropTypes.string,
-    visible: PropTypes.bool,
-    opener: PropTypes.instanceOf(Element),
-    onClose: PropTypes.func,
-    networkType: PropTypes.string,
     connectedAccount: PropTypes.bool,
+    popoverAction: PopoverActionType,
+    networkType: PropTypes.string,
+    onClose: PropTypes.func,
+    opener: PropTypes.instanceOf(Element),
+    visible: PropTypes.bool,
   }
   render() {
     const {
       address,
-      visible,
-      opener,
-      onClose,
-      networkType,
       connectedAccount,
+      popoverAction,
+      networkType,
+      onClose,
+      opener,
+      visible,
     } = this.props
+    const { title = 'Address' } = popoverAction || {}
+
     return (
       <Popover visible={visible} opener={opener} onClose={onClose}>
         <section
@@ -62,7 +68,7 @@ class IdentityBadgePopover extends React.PureComponent {
                 color: ${theme.textSecondary};
               `}
             >
-              Address
+              {title}
             </h1>
             {connectedAccount && (
               <Badge.Identity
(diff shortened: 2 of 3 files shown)


src/components/IdentityBadge/IdentityBadge.js
```diff
@@ -7,14 +7,16 @@ import Text from '../Text/Text'
 import EthIdenticon from '../EthIdenticon/EthIdenticon'
 import ButtonBase from '../Button/ButtonBase'
 import IdentityBadgePopover from './IdentityBadgePopover'
+import customLabelActionShape from './CustomLabelActionShape'
 
 class IdentityBadge extends React.PureComponent {
   static propTypes = {
+    connectedAccount: PropTypes.bool,
+    customLabelAction: PropTypes.shape(customLabelActionShape),
```
bpierre: What would you think of `customAction`?

src/components/IdentityBadge/CustomLabelActionShape.js
```diff
@@ -0,0 +1,6 @@
+import PropTypes from 'prop-types'
```
bpierre: What do you (and @sohkai) think of always ending types by `Type`? In that case, the file would be named `CustomLabelActionType.js` and would export the `PropType.shape` directly.

src/components/IdentityBadge/CustomLabelActionShape.js
```diff
@@ -0,0 +1,6 @@
...
+
+export default {
+  label: PropTypes.string,
+  onClick: PropTypes.func,
```
bpierre: We should probably make them both `isRequired` (e.g. we don’t want users to only set `label` but not `onClick`).

src/components/IdentityBadge/IdentityBadge.js
AquiGorka: I'll update it to `customAction` but let me explain my reason for going with`customLabelAction`: with @dizzypaty we discussed how this component is very important and maybe adding a custom action makes sense now but we didn't want to leave the door open to add more and more extra functionality here, so I wanted to make it explicit this is for `custom labels` not any extra `custom action`.

bpierre: Oooh I see, sorry I was seeing the “Label” part as the _label of the button_, (which I was seeing as “any custom action”). But I understand now that it refers to the custom labels project specifically.

In that case, maybe we could go full specific by having a single `onRename` prop? The presence of this prop would determine if the button is displayed or not, and the (button’s) label would be managed, and later translated, by aragonUI directly. What do you think?

AquiGorka: The thing is the label comes from whoever sets the badge because it can read `add` or `edit` if there is a custom label set.

bpierre: Oh good point I didn’t think about that, let’s keep it outside then.

The thing that still bothers me is that the current version is quite neutral (we don’t control the button, but we can set anything we want as a label and callback function), but its naming (`customLabelAction`) is very specific. I think we should go either way (specific or generic) but making sure both the naming and the implementation are following it, or we’ll start seeing `customLabelAction` being used to implement other types of actions, only because it would be possible and the only way to do it.

To make it generic, we could rename `customLabelAction` to be `secondaryAction`, to make it a neutral action that can be set on the badge?

```jsx
const MyIdentityBadge = ({ address }) => {
  // Let’s assume useAddressLabel() does all
  // the work to get & update the stored label.
  const [label, renameLabel] = useAddressLabel(address)
  return (
    <IdentityBadge
      identity={label || address}
      secondaryAction={{
        label: label? 'Edit custom label' : 'Add custom label',
        onClick: renameLabel,
      }}
    />
  )
}
```

We could even support other buttons rather than having an object (so people can use an API they know already 👍), I don’t think it would be a big issue:

```jsx
const MyIdentityBadge = ({ address }) => {
  const [label, renameLabel] = useAddressLabel(address)
  return (
    <IdentityBadge
      identity={label || address}
      secondaryAction={
        <Button mode="outline" onClick={renameLabel}>
          {label? 'Edit custom label' : 'Add custom label'}
        </Button>
      }
    />
  )
}
```

And the other way would be to be fully specific, having the button label managed internally, and having names specific to the feature:

```jsx
// Since `label` is passed independently, we have enough
// information to display either the “Edit” or “Add” string.
const MyIdentityBadge = ({ address }) => {
  const [label, renameLabel] = useAddressLabel(address)
  return (
    <IdentityBadge
      identity={address}
      customLabel={{ label, onRename: renameLabel }}
    />
  )
}
```

What do you think?

src/components/IdentityBadge/IdentityBadgePopover.js
```diff
@@ -95,10 +99,23 @@ class IdentityBadgePopover extends React.PureComponent {
...
+            </Button>
+          )}
         </section>
       </Popover>
     )
   }
+
+  handleCustomLabelActionClick = onClose => () => {
+    onClose()
+    this.props.customLabelAction.onClick()
```
sohkai: We should check that this exists, in case this gets used somewhere else later (which may not have the `customLabelAction` check)

src/components/IdentityBadge/IdentityBadge.js
sohkai: I do think that we should be generic here; since this is a generic UI component in aragonUI, we shouldn't confine a particular feature to something related to the Aragon client.

Even if we do consider Aragon client use-cases, this may be used by other identity providers down the line to do specific actions related to them. 

`popoverAction` could be another name.

src/components/IdentityBadge/IdentityBadge.js
```diff
@@ -72,6 +77,7 @@ class IdentityBadge extends React.PureComponent {
             networkType={networkType}
             opener={this._element.current}
             onClose={this.handleClose}
+            customLabelAction={customLabelAction}
```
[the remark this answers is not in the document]
AquiGorka: Yes, I wanted to get the functionality changes in - get them "approved" and then work on the aesthetics...

src/components/IdentityBadge/IdentityBadge.js
AquiGorka: Alright then, I agree about the generic nature suggested. I'll update to `popoverAction` 😄

src/components/IdentityBadge/IdentityBadge.js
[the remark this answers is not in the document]
AquiGorka: Added a generic property `title`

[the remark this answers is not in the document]
bpierre: What about `label`? Since it’s going to be displayed on the badge too, and the component is named after it (and we use the term `label` for badges). Or do we want a separate prop for the badge label and the popover title?

Also I originally used the term `entity` for the address, because I wanted to use the same prop to have other things displayed on the badge, like an ENS domain or another name, and recognize them automatically somehow. But I think now that `address` should be a separate prop, especially since the badge doesn’t resolve anything by itself.

We also need to:

- Being able to override the identicon (I think we could just have a node?).
- Set a “label source” (e.g. “custom label” in this example). Maybe this could be a node too, recommending people to use `Badge`? And if we decide to separate the popover title and the badge label, this could be part of the title (which could be named `popoverHeading` since it would then be more than a title).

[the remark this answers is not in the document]
AquiGorka: Already using a `label` prop for the [button label](https://github.com/aragon/aragon-ui/pull/327/files#diff-8e2e35885a1aff8dad7184bc7ac8e625R4)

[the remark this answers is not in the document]
AquiGorka: > Being able to override the identicon

For custom labels functionality first iteration?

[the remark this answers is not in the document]
AquiGorka: > Set a “label source” (e.g. “custom label” in this example). Maybe this could be a node too, recommending people to use `Badge`? And if we decide to separate the popover title and the badge label, this could be part of the title (which could be named `popoverHeading` since it would then be more than a title).

I think this is not generic as discussed earlier, that's why I send the whole `title` to be displayed (ps it is displayed inside an h1 which I think refers better to title no?)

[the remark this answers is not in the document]
bpierre: Oh, I thought we were going to be fully generic by letting people set the button directly. But I don’t think it’s a problem otherwise, we are talking about a button label in both cases.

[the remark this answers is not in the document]
AquiGorka: I'd rather set the type of button in stone and only allow to set its `label` and `onClick` handler. This way we can avoid having who knows what kind of button or even other nodes/elements in there.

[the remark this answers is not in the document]
bpierre: Cool, let’s do this then!

src/components/IdentityBadge/IdentityBadgePopover.js
```diff
@@ -5,28 +5,34 @@ import { theme } from '../../theme'
...
 
 class IdentityBadgePopover extends React.PureComponent {
   static propTypes = {
     address: PropTypes.string,
-    visible: PropTypes.bool,
-    opener: PropTypes.instanceOf(Element),
-    onClose: PropTypes.func,
-    networkType: PropTypes.string,
     connectedAccount: PropTypes.bool,
+    popoverAction: PopoverActionType,
```
sohkai: Might as well make this alphabetized 😄

src/components/IdentityBadge/IdentityBadgePopover.js
```diff
@@ -5,28 +5,34 @@ import { theme } from '../../theme'
...
-      onClose,
-      networkType,
       connectedAccount,
+      popoverAction,
+      networkType,
+      onClose,
+      opener,
+      visible,
     } = this.props
+    const { title = 'Address' } = popoverAction || {}
```
sohkai: Should we use a default prop instead?

src/components/IdentityBadge/IdentityBadgePopover.js
```diff
@@ -95,10 +102,26 @@ class IdentityBadgePopover extends React.PureComponent {
...
     )
   }
+
+  handlePopoverActionClick = onClose => () => {
+    onClose()
+    const {
+      popoverAction: { onClick = () => {} },
+    } = this.props
+    onClick()
+  }
```
sohkai: Could we avoid this extra binding since `onClose()` seems to be available from the props? It's also bad for performance, since we'd create a new function on every render.

Should we also use a default prop for `popOverAction` instead (so a new object / function doesn't get created on each invocation)? Note that this currently also doesn't handle the case where `this.props.popoverAction` is `nullable`.

src/components/IdentityBadge/IdentityBadgePopover.js
AquiGorka: Ahh, I tried but a null `popoverAction` is needed to show/hide other elements.

src/components/IdentityBadge/IdentityBadgePopover.js
AquiGorka: >Note that this currently also doesn't handle the case where this.props.popoverAction is nullable.

The button won't show if `popoverAction` is `null` so no click.

src/components/IdentityBadge/IdentityBadgePopover.js
sohkai: Ah, that makes sense 👍.

src/components/IdentityBadge/IdentityBadgePopover.js
```diff
@@ -95,10 +102,26 @@ class IdentityBadgePopover extends React.PureComponent {
               See on Etherscan
             </SafeLink>
           </p>
+          {popoverAction && (
+            <Button
+              mode="outline"
```
bpierre: I think the size of the button should be mini: `size="mini"`

AquiGorka: Let's make it smaller then 😄

src/components/IdentityBadge/IdentityBadgePopover.js
```diff
@@ -5,28 +5,34 @@ import { theme } from '../../theme'
...
-      onClose,
-      networkType,
       connectedAccount,
+      popoverAction,
+      networkType,
+      onClose,
+      opener,
+      visible,
     } = this.props
+    const { title = 'Address' } = popoverAction || {}
```
bpierre: Sorry to come late, but I think `title` should be moved outside of `popoverAction` since it’s not related to the action button 😄

src/components/IdentityBadge/PopoverActionType.js
```diff
@@ -0,0 +1,7 @@
+import PropTypes from 'prop-types'
+
+export default PropTypes.shape({
+  label: PropTypes.string.isRequired,
+  onClick: PropTypes.func.isRequired,
+  title: PropTypes.oneOfType([PropTypes.string, PropTypes.element]).isRequired,
```
bpierre: Do we want to have `PropTypes.node` here, instead of `PropTypes.oneOfType([PropTypes.string, PropTypes.element])`?

AquiGorka:
Listo @bpierre commnets addressed 😄

src/components/IdentityBadge/IdentityBadgePopover.js
sohkai: @AquiGorka Just wanted to double check, could we just access `this.props.onClose` in `handlePopoverActionClick()` so we don't need to create a new function on every render?

AquiGorka: I'll set the default prop to `noop` to fix this

PR #329, opened 2019-02-27
Main: relative assets url

bpierre:
Related discussion: https://github.com/aragon/aragon-apps/pull/692

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,7 +47,7 @@ const App = () => (
 
 Open https://ui.aragon.org/ to see the list of the available components and learn how to use them.
 
-### Aragon UI assets
+### Assets
 
 Aragon UI require some assets (e.g. fonts) in order to work properly. These need to be copied where they can be accessed by the library, like the `public/` directory of a project using [Create React App](https://github.com/facebookincubator/create-react-app/blob/master/packages/react-scripts/template/README.md#adding-assets-outside-of-the-module-system).
 
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ Install it from npm:
 npm install --save @aragon/ui
 ```
 
-Copy its assets in your public directory:
+Copy its assets into your public directory:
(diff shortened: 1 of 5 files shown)


bpierre:
@sohkai I also updated the README (totally forgot to update it for `Main` 😱) and the default directory name for `copy-aragon-ui-assets` is now `aragon-ui`.

This is a breaking change but it shouldn’t affect our apps as I believe we are already always renaming it `aragon-ui` anyway.

bin/copy-aragon-ui-assets
```diff
@@ -51,7 +51,7 @@ async function main(argv, argspec) {
   const destination = path.resolve(
     process.cwd(),
     args._[0],
-    args['--dirname'] || 'aragon-ui-assets'
+    args['--dirname'] || 'aragon-ui'
```
sohkai: I had no idea this wasn't just `aragon-ui` before 😅

README.md
```diff
@@ -17,39 +17,61 @@ Used by:
...
 
+### Quick setup
+
 Install it from npm:
 
 ```sh
 npm install --save @aragon/ui
 ```
 
-Aragon UI require some assets (e.g. fonts) in order to work properly. These need to be copied where they can be accessed by the library, like the `public/` directory of a project using [Create React App](https://github.com/facebookincubator/create-react-app/blob/master/packages/react-scripts/template/README.md#adding-assets-outside-of-the-module-system).
```
sohkai: I think it'd still be useful to say why we provide this bit and that the library does requires these assets

README.md
```diff
@@ -17,39 +17,61 @@ Used by:
...
-
-To copy these assets, use the `copy-aragon-ui-assets` command provided by Aragon UI:
+Copy its assets in your public directory:
```
sohkai: in**to**

Issue #221, opened 2018-10-16
Add the TokenBadge component


bpierre:
See https://github.com/aragon/aragon-apps/issues/273#issuecomment-422874396

bpierre:
Done! https://github.com/aragon/aragon-ui/pull/334

Thanks @rperez89 :100: :100: :100:

PR #334, opened 2019-03-05
TokenBadge Component #221

rperez89:
@bpierre @sohkai Hey guys is not ready to be merged i have some questions about how the icon, how are we going to get it? and probably we need a new AddressField component right? because this one is using the EthIdentIcon.

Screen Shot 2019-03-27 at 4 17 05 PM

Screen Shot 2019-03-27 at 4 18 03 PM


AquiGorka:
Hey @rperez89 could you post a couple of screenshots of this new component in the description?

rperez89:
> Hey @rperez89 could you post a couple of screenshots of this new component in the description?

Yes, sure! it's not finished as i did some questions on how to handle the icon. But is pretty similar to the IdentityBadge

sohkai:
On the icon, we should expect the user to input it into this component. However, @bpierre, I'm not sure if there's a way to make it generic between an img src and an svg component?

For token icons I imagine most people would like to use an URI.

bpierre:
@rperez89 Sorry to have been so long to review this PR, it’s looking good!

> However, @bpierre, I'm not sure if there's a way to make it generic between an img src and an svg component?

Yes, we could use `` by default (for external SVG / PNG images), and maybe also have another prop to let users set their own component if they want an inline SVG or something else?

@rperez89 @sohkai I wonder if we shouldn’t let people pass a provider function to resolve tokens in general, rather than having to use an extended version of this component in their apps? That way, we could also use it from other components in the future.

```jsx
import { TokensProvider } from '@aragon/ui'
import trustedTokens from './trusted-tokens'

// This function could also be async
function resolveToken(address) {

  const token = trustedTokens.get(address)

  if (!token) {
    return null
  }

  return {
    name: token.name,
    symbol: token.symbol,
    imageUrl: token.imageUrl,
  }
}

function MyApp() {
  return (
    <TokensProvider resolver={resolveToken}>
      <div>
        <TokenBadge address="0xcafe" />
      </div>
    </TokensProvider>
  )
}
```

We could also let users access it with a Hook:

```jsx
import { useTokenResolver } from '@aragon/ui'

function TokenInfo({ address }) {

  const { name, symbol, imageUrl, loading, exists } = useTokenResolver(address)

  if (!exists) {
    return 'The token couldn’t be found.'
  }

  if (loading) {
    return 'Loading…'
  }

  return (
    <ul>
      <li>Image: <img src={imageUrl} alt="" /></li>
      <li>Name: {name}</li>
      <li>Symbol: {symbol}</li>
    </ul>
  )
}
```

In the future, we could also use this mechanism to provide a default list of trusted tokens from `aragon/aragon`, either from `@aragon/api-react` (by declaring `<TokensProvider>`) or from `@aragon/ui` (by pulling the data from `@aragon/api-react` then declaring `<TokensProvider>`), so that token badges would just work ® without having to do anything. App authors would then have the possibility to extend or replace this default provider.

devbox/apps/TokenBadge.js
```diff
@@ -0,0 +1,27 @@
+import React from 'react'
+import styled from 'styled-components'
+import { TokenBadge } from '@aragon/ui'
+
+class App extends React.Component {
+  render() {
+    return (
+      <Main>
+        <TokenBadge
+          entity="0x2c9341a52cfa3f2c2554ca1803134137b9366b3c"
```
bpierre: We should probably use `address` here, since it can not be something else (unlike `IdentityBadge`).

devbox/apps/TokenBadge.js
```diff
@@ -0,0 +1,27 @@
...
+          name='Aragon'
+          connectedAccount
```
bpierre: This shouldn’t be needed.

```suggestion
```

src/components/TokenBadge/TokenBadge.js
```diff
@@ -0,0 +1,115 @@
...
+}
+
+const Main = styled.div`
+  overflow: hidden;
+  display: flex;
+  align-items: center;
+  height: 24px;
+  background: #daeaef;
+  border-radius: 3px;
+  cursor: default;
```
bpierre:
```suggestion
  cursor: pointer;
```

sohkai:
> I wonder if we shouldn’t let people pass a provider function to resolve tokens in general, rather than having to use an extended version of this component in their apps?

Ohh yes, I think this would be great :+1:!

> In the future, we could also use this mechanism to provide a default list of trusted tokens from aragon/aragon

Yes, or we could have an API that allows the app to check if the token is trusted by the Aragon client (e.g. `app.verifyToken(<addr>)`. Would also make it a bit easier to work with an IPFS metadata registry in the future. https://github.com/aragon/aragon.js/issues/293

bpierre:
@rperez89 @sohkai I added a few more commits! This idea is to get a first version of the badge released without the tokens provider, then implement it as a second step.

Changes:

- Add a `useImageExists` /  `ImageExists` utility.
- `AddressField`: allow the icon to be replaced.
- Add a `tokenIconUrl()` utility (get a token icon URL from the contract address).
- TokenBadge:
  - Rename `entity` to `address` (since it is always an address in that case)
  - Make `symbol` a required prop.
  - Remove `shorten` (only used by IdentityBadge).
  - Remove `fontSize` (there should be no variations in font size).
  - Display the icon if it exists (badge + popover).
  - Add a label to the popover close button.

src/components/TokenBadge/TokenBadge.js
```diff
@@ -30,68 +31,94 @@ class TokenBadge extends React.PureComponent {
...
+
+    const label = name && symbol ? `${name} (${symbol})` : symbol
 
     return (
       <React.Fragment>
         <ButtonBase
           ref={this._element}
-          title={address}
-          onClick={address && this.handleOpen}
+          title={`${label} − ${address || 'No address'}`}
```
sohkai: Should this also use the `isValidAddress` check?

src/components/TokenBadge/TokenBadge.js
```diff
@@ -30,68 +31,94 @@ class TokenBadge extends React.PureComponent {
...
           css={`
             display: inline-flex;
+            overflow: hidden;
+            height: 24px;
             color: ${theme.textPrimary};
           `}
         >
-          <Main title={`${name} (${symbol})`}>
-            <Label>
-              <Icon src={`https://chasing-coins.com/coin/logo/${symbol}`} />
```
sohkai: Something we may want to consider is using this as a backup on networks other than mainnet (so the app still shows token icons on rinkeby)

src/components/TokenBadge/TokenBadgePopover.js
```diff
@@ -50,22 +52,29 @@ class TokenBadgePopover extends React.PureComponent {
...
-              `}
-            >
-              {`${name} (${symbol})`}
-            </h1>
-          </header>
-          <AddressField address={address} />
+            {label}
+          </h1>
+          {iconUrl ? (
+            <ImageExists src={iconUrl}>
```
sohkai: Wonder if it makes sense to bake this `url` check into `ImageExists` itself, so you can safely use this component at no performance cost if the `src` is falsey

bpierre: Yes totally, it doesn’t hurt :+1:

src/hooks/useImageExists.js
```diff
@@ -8,24 +11,35 @@ export function useImageExists(src) {
...
+    const init = () => {
+      setExists(false)
+      setLoading(true)
+
+      if (srcCache.get(src)) {
+        success()
+        return
       }
+
+      // TODO: ensure only one image is loading at a time for a given src.
```
sohkai: No need to do this now, but I implemented a super simple util for doing this in aragon.js: [`AsyncRequestCache`](https://github.com/aragon/aragon.js/blob/master/packages/aragon-wrapper/src/utils/AsyncRequestCache.js)

src/utils/cached-map.js
```diff
@@ -0,0 +1,66 @@
+const A_DAY = 1000 * 60 * 60 * 24
+
+function cachedMap({ expireAfter = A_DAY, size = 100 } = {}) {
+  const cache = new Map()
+
+  // Delete the first (oldest) entry until we are back to `size`,
+  // using a loop because getValue() is async.
```
sohkai: I don't think this part of the comment applies? `cache.keys().next()` should be synchronous.

src/utils/cached-map.js
```diff
@@ -0,0 +1,66 @@
...
+
+  // Return the value if the entry exists and is not expired, or fetch it using
+  // getValue() otherwise. Use forceUpdate to always call getValue().
+  async function fetch(key, getValue, forceUpdate) {
+    const now = new Date()
+
+    const cachedEntry = cache.get(key)
+
+    const update =
+      forceUpdate || !cachedEntry || now - cachedEntry.lastAccess > expireAfter
```
sohkai: We probably want to get the `lastAccess` variable before the `get()`, otherwise `get()` will update it to be now.

src/hooks/useImageExists.js
bpierre: Oh nice yes!

src/utils/cached-map.js
bpierre: Yes you’re right! The `Map` only gets updated after `getValue()` returned so it’s all sync at this point and `cache` should never be more than `101`. It might or might not have been useful at some point during the implementation :smile:

src/utils/cached-map.js
bpierre: Maybe I misunderstand, but we are calling `Map.prototype.get` here, not this module’s `get()` :+1:

bpierre:
> LRU cache

TIL :eyes:

src/utils/cached-map.js
sohkai: Ahh yes, gotcha!

PR #341, opened 2019-03-28
Viewport: export context for compatibility with hooks

AquiGorka:


bpierre:
I was thinking of exporting a hook directly for this:

`const { width, height } = useViewport()`

What do you think?

AquiGorka:
We could include those two props into the same `Viewport` context no?

bpierre:
Yes they all are in the context, but I see the context more as an implementation thing rather than something we should have as a public API.

I think it would be nice to provide hooks equivalents for every component in `providers/`: https://github.com/aragon/aragon-ui/tree/master/src/providers

AquiGorka:
Super, in the meantime, I added these changes so that we can start using `Viewport` with hooks (to make code easier on the 👀)

bpierre:
But if we export the context we’ll have to support it, and exposing contexts is not the kind of API we have elsewhere in aragonUI… why not adding a hook directly?

AquiGorka:
Ahhh, now I understand what you were saying: you don't want to expose the `context` but rather the custom hook.

AquiGorka:
Updated: context not exported and exports a hook as suggested.

bpierre:
Awesome! We should also export it here:

https://github.com/aragon/aragon-ui/blob/master/src/providers/index.js#L6

AquiGorka:
@bpierre listo 😄

src/providers/Viewport/README.md
```diff
@@ -6,6 +6,24 @@ to help building responsive apps.
...
+      {within('medium', 'large') && <div>medium</div>}
+      {above('large') && <div>large</div>}
+    </div>
+  )
+}
+```
+
+### Render prop
+
 ```jsx
```
bpierre: Could we add the import + wrap in in `MyComponent` here, to highlight the difference between the two?

AquiGorka: Sure. Done.

PR #342, opened 2019-03-28
Popover: fix the click outside detection

bpierre:
Fix the super annoying bug where outside clicks were not detected below the popover.

Also move from `Root` to `RootPortal`.

### DIFF
--- a/devbox/apps/Popover.js
+++ b/devbox/apps/Popover.js
@@ -1,61 +1,46 @@
-import React from 'react'
+import React, { useRef, useState, useCallback } from 'react'
 import styled from 'styled-components'
 import { Popover, Root, Button } from '@aragon/ui'
 
-class App extends React.Component {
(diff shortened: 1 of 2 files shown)


devbox/apps/Popover.js
```diff
@@ -1,61 +1,46 @@
-import React from 'react'
+import React, { useRef, useState, useCallback } from 'react'
```
AquiGorka: 😨 
I'd love to do this all around...

src/components/Popover/Popover.js
```diff
@@ -76,6 +71,9 @@ class PopoverBase extends React.Component {
 
   initPopper() {
     const { opener, placement } = this.props
+
+    this._openerRect = opener ? opener.getBoundingClientRect() : null
```
AquiGorka: ❤️

Issue #300, opened 2019-01-28
Timer: support year and month formats


sohkai:
Required for https://github.com/aragon/aragon-apps/issues/638.

Currently only supports up to days.

sohkai:
Fixed with https://github.com/aragon/aragon-ui/pull/345.

PR #345, opened 2019-03-29
Timer Formats

rperez89:
Years and Month formats for the timer. #300

Screen Shot 2019-03-28 at 8 48 18 PM

@bpierre @sohkai let me know your thoughts


devbox/apps/Timer.js
```diff
@@ -2,6 +2,7 @@ import React from 'react'
 import styled from 'styled-components'
 import { Timer } from '@aragon/ui'
 
+
```
AquiGorka: Any special reason for this extra line?

src/components/Timer/README.md
```diff
@@ -45,10 +45,10 @@ The start of a timer, as a `Date` instance.
 
 ### `format`
 
-- Type `Enum: [ 'dhms', 'dhm', 'hms', 'hm', 'ms', 'm', 's' ]`
+- Type `Enum: [ 'yMdms', 'yMdm', 'yMd', 'yM', 'Mdhms', 'Mdhm', 'Mdh', 'Md', 'dhms', 'dhm', 'hms', 'hm', 'ms', 'm', 's' ]`
```
AquiGorka: `yMdhms`?

rperez89: years, Months, days, hours and seconds. This is for the output that you want to see

devbox/apps/Timer.js
rperez89: ouuch i will remove it

devbox/apps/Timer.js
```diff
@@ -12,10 +13,10 @@ const now = Date.now()
 const timers = [
   { start: -40 * SECONDS },
   { start: -40 * SECONDS, showEmpty: true },
-  { start: -4 * HOURS, end: 8 * HOURS },
+  { start: -4 * DAYS, end: 20 * DAYS },
   { end: 1 * DAYS },
-  ...['dhms', 'dhm', 'hms', 'hm', 'ms', 'm', 's'].map(format => ({
-    end: 2 * DAYS - 55 * SECONDS,
+  ...['yMd','yM','Mdhms', 'Mdhm','Mdh', 'Md', 'dhms', 'dhm', 'hms', 'hm', 'ms', 'm', 's'].map(format => ({
+    end: 398 * DAYS - 55 * SECONDS,
```
AquiGorka: Maybe set this `398` into a constant to give it some semantic meaning?

src/components/Timer/README.md
AquiGorka: Yes, yes, how could I render `year, Month, day, hours, minutes and seconds`? I believe it is missing no?

rperez89: Done :)

AquiGorka:
@rperez89 when running the devbox I get the following:



If you take a close look, the first two timers are going forward instead of counting down, and the second one which has `showEmpty` set to `true`, is displaying elements with zeros (which is not what that [prop is for](https://github.com/aragon/aragon-ui/pull/264#issuecomment-448778703)).

rperez89:
Hi @AquiGorka  i think that i maybe broke something but i just switched to master and is working like this, the first two are going forward  because there is no end date seted. 

About the showEmpty i think that is the whay that shoud work.

Screen Shot 2019-05-02 at 8 16 48 PM

Maybe i am cofused with something.
If this is a bug let me know.

src/utils/date.js
```diff
@@ -18,7 +26,7 @@ export const difference = (date1, date2) => {
   const minutes = Math.floor(seconds / MINUTE_IN_SECONDS)
   seconds = seconds % MINUTE_IN_SECONDS
 
-  return { days, hours, minutes, seconds, totalInSeconds }
+  return { years, months, days, hours, minutes, seconds, totalInSeconds }
```
sohkai: The years and months calculation is a bit inaccurate here, and I'm wondering if we should use something from date-fns to do this instead?

```
365 * 24 * 60 * 60 =/= 12 * 30 * 24 * 60 * 60
```

rperez89: Hi @sohkai , some time ago we talked about that with @bpierre and you that is something that could not be done with date-fns so we took this way. https://date-fns.org/v2.0.0-alpha.22/docs/formatDistanceStrict  maybe you want to take a look to this function but i found it useless . In the other hand with moment js you can do exactly what we want here just doing something like this : `moment.duration(duration).days(); moment.duration(duration).hours(); etc`

sohkai: This might add a lot of bloat, since we'd use a lot more of `date-fns`, but we could use the `differenceInX` helpers (e.g. for [years](https://date-fns.org/v2.0.0-alpha.22/docs/differenceInYears)) and then use the `subX` helpers (e.g. for [years](https://date-fns.org/v2.0.0-alpha.22/docs/subYears)) to be more accurate.

bpierre: I added the change, thanks for the suggestion @sohkai, it works perfect! https://github.com/aragon/aragon-ui/pull/345/commits/6ae52cbc5cf9ca27906c41d9e547097430642b59

devbox/apps/Timer.js
```diff
@@ -8,12 +8,12 @@ const HOURS = 60 * MINUTES
 const DAYS = 24 * HOURS
 
 const now = Date.now()
-const OVER_A_YEAR_END = 398 * DAYS
+const OVER_A_YEAR_END = 298 * DAYS - 55 * SECONDS
```
sohkai: Should this still be `398`? `298 * DAYS` is not quite "over a year".

src/components/Timer/Timer.js
```diff
@@ -34,14 +35,18 @@ const getFormat = memoize(format =>
...
+function removeEmpty(units) {
+  // keep at least one unit even at 0
+  const lastZeroIndex = units.map(u => u[1]).lastIndexOf(0) + 1
+  for (let i = 0; i < units.length && i < lastZeroIndex; i++) {
+    if (units[i][1] !== 0 && units[i][1] !== null) {
+      return units
+    }
+    units[i][1] = null
+  }
+  return units
```
sohkai: This sounds like a particular good usecase of `array.reduceRight()`:

```js
const [sanitizedUnits] = units.reduceRight(
  ([sanitizedUnits, seenEmpty], unit) => {
    const empty = unit[1] === 0 || unit[1] === null
    if (!seenEmpty || !empty) {
      sanitizedUnits.unshift(unit)
    }
    return [sanitizedUnits, seenEmpty || empty]
  },
  [[], false]
)
return sanitizedUnits 
```

--------------

Otherwise, I think the loop should be changed to not have a `return`:

```js
if (units[i][1] === 0 || units[i][1] === null) {
  units[i][1] = null
}
```

Otherwise you'll be returning the `units` array immediately on non-zero units.

devbox/apps/Timer.js
bpierre: yes :s

PR #354, opened 2019-04-22
Popover: rely on Popper modifiers settings to center it

bpierre:


src/components/Popover/Popover.js
```diff
@@ -13,8 +13,8 @@ class PopoverBase extends React.Component {
   static propTypes = {
     opener: PropTypes.instanceOf(Element),
     placement: PropTypes.oneOf(
-      // See https://popper.js.org/popper-documentation.html#Popper.placements
-      // "center" is a value that doesn’t exits in Popper.
+      // "center" is a value that doesn’t exits in Popper, but we are using it
```
AquiGorka:
```suggestion
      // "center" is a value that doesn’t exist in Popper, but we are using it
```

src/components/Popover/Popover.js
```diff
@@ -123,7 +141,7 @@ class PopoverBase extends React.Component {
   }
 
   render() {
-    const { zIndex, children, transitionStyles, placement } = this.props
+    const { zIndex, children, transitionStyles } = this.props
```
AquiGorka: Out of curiosity and no need to change anything: any reason why you didn't destructure directly? `transitionStyles: { scale, opacity }`?

bpierre: Not really, I tend to use one or the other depending on how prettier makes it look :laughing:

PR #358, opened 2019-04-26
ProgressBar: indeterminate state + other improvements

bpierre:
image

### DIFF
--- a/src/components/ProgressBar/ProgressBar.js
+++ b/src/components/ProgressBar/ProgressBar.js
@@ -1,43 +1,49 @@
 import React from 'react'
 import PropTypes from '../../proptypes'
-import styled from 'styled-components'
-import { useSpring, animated } from 'react-spring'
+import { useSpring, animated, interpolate } from 'react-spring'
 import { springs } from '../../utils/styles'
 import { pause } from '../../utils/miscellaneous'
 import { warn } from '../../utils/environment'
 import { theme } from '../../theme'
 
-const BAR_HEIGHT = 6
 const RADIUS = 2
+const BAR_HEIGHT = 6
 
-const transformCss = (scale, translate) =>
-  `scale3d(${scale}, 1, 1) translate3d(${translate}%, 0, 0)`
+const INDETERMINATE_BAR_WIDTH = 1 / 4
+const INDETERMINATE_ANIM_DURATION = [1200, 400]
 
-const springNormal = value => ({
+const springNormal = (value, restart, animate) => ({
   config: { ...springs.smooth, precision: 0.001 },
-  from: { transform: transformCss(0, 0) },
-  to: { transform: transformCss(value, 0) },
-  reset: false,
+  from: { scale: 0, x: 0 },
+  to: { scale: restart ? 0 : value, x: 0 },
+  reset: restart,
+  immediate: restart || !animate,
 })
 
-const springIndeterminate = () => ({
-  config: {
-    ...springs.sluggish,
-    precision: 0.01,
-    clamp: true,
-  },
-  from: { transform: transformCss(0.5, -100) },
+const springIndeterminate = (restart, animate) => ({
+  config: { duration: INDETERMINATE_ANIM_DURATION[0] },
+  from: { scale: INDETERMINATE_BAR_WIDTH, x: -INDETERMINATE_BAR_WIDTH },
   to: async next => {
+    const scale = INDETERMINATE_BAR_WIDTH
+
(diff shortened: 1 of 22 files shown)


src/components/ProgressBar/ProgressBar.js
```diff
@@ -1,55 +1,114 @@
 import React from 'react'
-import PropTypes from 'prop-types'
-import styled from 'styled-components'
-import { Spring, animated } from 'react-spring'
+import PropTypes from '../../proptypes'
+import { useSpring, animated, interpolate } from 'react-spring'
 import { springs } from '../../utils/styles'
+import { pause } from '../../utils/miscellaneous'
```
AquiGorka: May I suggest `sleep`? When I read `pause` and knowing this PR has changes to animations I instantly thought it has to do with animation pausing instead of waiting for some time to do continue execution.

src/proptypes.js
```diff
@@ -1,5 +1,39 @@
...
+  return isEmpty()
+    ? new Error(
+        `The prop \`${propName}\` is required for \`${componentName}\`.`
+      )
+    : null
+}
+
+// Accept any number in the 0 => 1 range
+function _0to1(props, propName, componentName) {
+  if (isEmpty()) {
```
AquiGorka: Missing param here, is it on purpose?

src/proptypes.js
```diff
@@ -1,5 +1,39 @@
 import PropTypes from 'prop-types'
 
+// Check if a value is “empty”, in the prop-types sense (null or undefined)
+function isEmpty(value) {
+  return value === undefined || value === null
+}
+
+// Require a prop to not be empty
+function requireProp(props, propName, componentName) {
+  return isEmpty()
```
AquiGorka: Should it be `isEmpty(props[propName])`?

src/components/ProgressBar/ProgressBar.js
bpierre: Yes that’s better :+1:

src/proptypes.js
bpierre: :see_no_evil:

src/components/ProgressBar/ProgressBar.js
```diff
@@ -1,55 +1,114 @@
...
+    }
+  }
+
+  // The indeterminate state can be triggered either by not setting the value
+  // (to mimic the <progress> element in HTML), or by setting the -1 value (for
+  // convenience in React).
+  const indeterminate = value === -1
+
+  // `isSwitching` is set to `true` to reset the transition between the normal
+  // and indeterminate state, then set to `false` immediatly after.
```
AquiGorka:
```suggestion
  // and indeterminate state, then set to `false` immediately after.
```

src/components/ProgressBar/ProgressBar.js
```diff
@@ -1,55 +1,114 @@
...
+    // Center the bar when not animated.
+    if (!animate) {
+      await next({ scale, x: 0.5 - scale / 2 })
+      return
+    }
+
+    // Indeterminate animation.
+    while (true) {
+      await next({ scale, x: 1 })
+      await pause(INDETERMINATE_ANIM_DURATION[1])
```
AquiGorka: This works with `await pause(INDETERMINATE_ANIM_DURATION[0])` too, no?

AquiGorka: If not, can you help me out understand why there are two durations (one and one third of it)?

bpierre: The first one was the duration of the slide transition, and the second one the pause before we start another loop.

It has changed now that it is based on a CSS animation: this constant [now holds the full duration](https://github.com/aragon/aragon-ui/pull/358/files#diff-387e775f15537990b06fc4c0db050505R13), and the pause is expressed [as a percentage](https://github.com/aragon/aragon-ui/pull/358/files#diff-387e775f15537990b06fc4c0db050505R20) in the `@keyframes` declaration.

bpierre:
> Just wondering if the API change from progress to value will be painful, but probably not?

Yes we can keep it for a while, I think 1.0 could be a good opportunity to remove all the deprecated things at the same time.

> Let's make sure to highlight this in the release notes and remember to update any usage of this component elsewhere.

Yes and yes :+1:

sohkai:
Took a quick look and it turns out we're not using this component anywhere else, except `TransactionProgress` (which isn't used externally anyway) :smile:.

bpierre:
Awesome thanks!

PR #362, opened 2019-05-02
devbox: auto generate the apps index

bpierre:


devbox/scripts/apps-index.js
```diff
@@ -0,0 +1,14 @@
...
+// Do not edit: manual changes will be overwritten.`
+)
+console.log(
+  fs
+    .readdirSync(__dirname + '/../apps')
+    .filter(f => f !== 'index.js' && /\.js$/.test(f))
+    .map(f => path.basename(f, '.js'))
+    .map(f => `export { default as ${f} } from './${f}'`)
+    .join('\n')
+)
```
sohkai: Would it make sense for this just to write to `apps/index.js` directly?

bpierre: Yes :+1:

PR #366, opened 2019-05-07
IdentityBadge: badgeOnly prop 

delfipolito:
If the badgeOnly prop is set to true, the IdentityBadge will not have an `on click` behavior and will not display the IdentityBadgePopover

### DIFF
--- a/src/components/IdentityBadge/IdentityBadge.js
+++ b/src/components/IdentityBadge/IdentityBadge.js
@@ -114,7 +114,7 @@ class IdentityBadge extends React.PureComponent {
             </Text>
           </div>
         </ButtonBase>
-        {address && (
+        {address && !badgeOnly && (
           <IdentityBadgePopover
             address={address}
             visible={opened}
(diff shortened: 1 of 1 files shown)


src/components/IdentityBadge/IdentityBadge.js
```diff
@@ -58,12 +61,13 @@ class IdentityBadge extends React.PureComponent {
...
           ref={this._element}
           title={address}
-          onClick={address && this.handleOpen}
+          onClick={address && !badgeOnly && this.handleOpen}
           css={`
             display: inline-flex;
             overflow: hidden;
             color: ${theme.textPrimary};
             height: 24px;
+            ${badgeOnly && 'cursor: text;'};
```
bpierre: What would you think about setting `disabled` on `ButtonBase` to achieve this? It could also be helpful e.g. for screen readers. https://github.com/aragon/aragon-ui/blob/master/src/components/Button/ButtonBase.js#L42

delfipolito: Oh yes, much better :)

PR #368, opened 2019-05-08
feat: Enhance blockexplorer url

### LINKED ISSUE #367: Remove "provided not supported" message or at least make that verbosity optional
The message seems to be emitted from this line: https://github.com/aragon/aragon-ui/blob/f6dcd1b90d0758ab051c2404a8c646a17483caec/src/utils/web3.js#L22

For development purposes, would it make sense to remove that message or handle the localhost case independently?
As a developer I already know that localhost is not supported, but it is kind of annoying to debug actual error or warning messages when the browser console gets flood by that provider message...

image

Also instead of pointing the etherscan link to localhost:3000 just point to etherscan or hide it at all when provider is not supported? Seems like bad UX and having the link there into the badge is not adding any value on those unsupported scenarios, another solution would be to just point to `#`

image

sohkai:
Fixes #367 

The `IdentityBadge` without a link:

Screen Shot 2019-05-08 at 11 58 04 AM

### DIFF
--- a/src/utils/web3.js
+++ b/src/utils/web3.js
@@ -8,6 +8,7 @@ const ETHERSCAN_NETWORK_TYPES = new Map([
   ['kovan', 'kovan.'],
   ['rinkeby', 'rinkeby.'],
   ['ropsten', 'ropsten.'],
+  ['goerli', 'goerli.'],
 ])
 const ETHERSCAN_TYPES = new Map([
   ['block', 'block'],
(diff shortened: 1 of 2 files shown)


src/utils/web3.js
```diff
@@ -8,6 +8,7 @@ const ETHERSCAN_NETWORK_TYPES = new Map([
   ['kovan', 'kovan.'],
   ['rinkeby', 'rinkeby.'],
   ['ropsten', 'ropsten.'],
+  ['goerli', 'goerli.'],
```
bpierre: :ok_hand:

PR #377, opened 2019-05-17
Timer: add maxUnits

bpierre:
Allows to set a maximum of units to be displayed.

Other changes:

- Add Jest + tests for `difference()`.
- Move the logic of leading zeros removal + max units to difference().


src/components/Timer/Timer.js
```diff
@@ -31,73 +31,59 @@ const formats = {
...
+    (units, symbol) =>
+      formats[format].includes(symbol) ? [...units, unitNames[symbol]] : units,
+    []
+  )
+)
 
-  const period = end ? [end, new Date()] : [new Date(), start]
+function getTime(start, end, format, showEmpty, maxUnits) {
+  const date1 = end || new Date()
+  const date2 = end ? new Date() : start
```
sohkai: Not sure, but I think this shoudl use `start`?

```suggestion
  const date2 = start ? new Date() : start
```

bpierre: They are exclusive: if `end` is set, `start` is always `new Date()`.

src/utils/date.test.js
```diff
@@ -0,0 +1,154 @@
...
+      })
+    ).toEqual({
+      years: 0,
+      months: null,
+      days: null,
+      hours: null,
+      minutes: null,
+      seconds: null,
+    })
+  })
```
sohkai: Would be great to add a test for this with `difference(NOW, NOW)`; I'd expect it to have all null except for `seconds: 0`?

src/utils/date.js
```diff
@@ -23,32 +23,63 @@ const fnsByUnit = [
...
+      }
+
+      if (index === fnsByUnit.length - 1) {
+        delete result.remaining
+        delete result.remainingUnits
+        delete result.seenNonZero
+
+        // include at least one 0
+        if (Object.values(result).every(val => val === null)) {
+          result[fnsByUnit.find(([unit]) => units.includes(unit))[0]] = 0
```
sohkai: I think this is the only weird one that violates my assumptions of "at least one 0"; I'd prefer this to be the most precise unit available (so doing a `fnsByUnit.reverse()` would work).

bpierre: Yes definitely a bug, thanks! https://github.com/aragon/aragon-ui/pull/377/commits/e7a6e899d76fb287b885afb9b4bea6f74874adf5

src/utils/date.test.js
bpierre: Done https://github.com/aragon/aragon-ui/pull/377/commits/e7a6e899d76fb287b885afb9b4bea6f74874adf5

PR #379, opened 2019-05-18
Add LoadingRing

bpierre:


src/components/LoadingRing/LoadingRing.js
```diff
@@ -0,0 +1,67 @@
...
+const spin = keyframes`
+  from {
+    transform: rotate(0deg);
+  }
+  to {
+    transform: rotate(360deg);
+  }
+`
+
+function LoadingRing({ paused, ...props }) {
```
AquiGorka: Should we `React.memo` this components?

bpierre: Yes! :ok_hand:

PR #380, opened 2019-05-18
Add FloatIndicator

bpierre:
This is the generic component that displays a box at the bottom of the screen.

I’m not super happy with the name, I felt `FloatingIndicator` was too long… open to suggestions!


src/components/FloatIndicator/FloatIndicator.js
```diff
@@ -0,0 +1,74 @@
...
+        /* eslint-disable react/prop-types */
+        (({ progress }) => (
+          <div
+            css={`
+              position: absolute;
+              width: 100%;
+              display: flex;
+              justify-content: center;
+              bottom: 0;
+              padding-bottom: 25px;
```
sohkai: Wondering if there's a preference to use padding or if `bottom: 25px` would be fine too?

AquiGorka: I think he used `padding` to make it possible to use this with other elements below no? Prob can be done with `bottom` too, but would have to make the container aware of element underneath.

src/components/FloatIndicator/FloatIndicator.js
```diff
@@ -0,0 +1,74 @@
...
+  </RootPortal>
+))
+
+FloatIndicator.propTypes = {
+  visible: PropTypes.bool,
+  children: PropTypes.node.isRequired,
+}
+
+FloatIndicator.defaultProps = {
+  visible: true,
```
AquiGorka: Should this be `false`?

src/components/FloatIndicator/FloatIndicator.js
bpierre: I can’t remember haha, probably a relic of how it evolved during development: I changed it to be `bottom: 25px` :+1:

src/components/FloatIndicator/FloatIndicator.js
bpierre: I prefer to display things by default in these cases (e.g. with [Popover](https://github.com/aragon/aragon-ui/blob/a5567440e59281c069ada40b817ceb1ba042b0f0/src/components/Popover/Popover.js#L208)): the reasoning is that it requires to be controlled anyway, and having it displayed by default might be less surprising than the opposite.

PR #383, opened 2019-05-27
Move from date-fns to Day.js

bpierre:
Two reasons:

- The development of `date-fns` v2 [seems to be on halt](https://github.com/date-fns/date-fns/issues/1165).
- Day.js is smaller: date-fns should be tree shaked, but it is not the case with the default rollup configuration, resulting in the entire library being included in the final bundle.

As a result, the minified bundle size gets reduced from `386.43 KB` to `348.62 KB`, and the gzip size from `104.43 KB` to `98.12 KB`.


bpierre:
> Looks like we might need to re-sync the package-lock.json to make the CI happy?

oops

PR #385, opened 2019-05-27
Tweak the Rollup configuration

bpierre:
- Update rollup and its plugins.
- Update ESLint plugins.
- Add a dependencies visualizer (HTML).
- Add a dependencies sizes (CLI).
- Show the brotli size alongside gzip.
- **Disable the CommonJS build**.

PR #346, opened 2019-03-29
Update icons (22x22 + generation script)

bpierre:
<p align=center>image

- All the icon dimensions are now exactly 22x22.
- Some icons colors, alignments and vector lines have been fixed too.
- A new script, scripts/generate-icons, is used to optimizing the icons SVGs and create the corresponding components.
- The gallery now automatically displays all icons (no need to update the page every time we add an icon).

PR #391, opened 2019-06-01
New style

bpierre:
Breaking changes in [UPGRADE.md](https://github.com/aragon/aragon-ui/blob/newstyle/UPGRADE.md)

- [x] Improve Rollup configuration (https://github.com/aragon/aragon-ui/pull/385)
- [x] Improve icons generation (https://github.com/aragon/aragon-ui/pull/346)
- [x] New color utility (`color`) (https://github.com/aragon/aragon-ui/pull/391/commits/d7a274ccf0096ccf55891c00be55fed63ab97404).
- [x] New light theme.
- [x] New dark theme (incomplete).
- [x] Update `Button` (new style + API).
- [x] Add `Layout` / `useLayout().`
- [x] Update `Card` (new style + interactive).
- [x] Add `CardsLayout`.
- [x] Add `Bar`.
- [x] Add Box.
- [x] Add Split.
- [x] Add Pagination.
- [x] Add DataView.
- [x] Update `TabBar` for `Bar`.
- [x] `TabBar` becomes `Tabs` and adapts to small viewports.
- [x] ~Update `AppView` (new style + reduced API).~ `AppView` is not needed anymore.
- [x] Add ~`AppHeader`~ `Header` (component for the app name + main action).
- [x] Remove `AragonApp`.
- [x] Add `BackButton` (to be used in `Bar`).
- [x] Add `DiscButton`.
- [x] Update icons (the previous icon names are still supported, but are aliases of the new icons).
- [x] ~Move the layout-related features of `Viewport` into `Layout`.~ Let’s do this later.
- [x] Remove WOFF 1 fonts.
- [x] Update `Main` to use `Layout` automatically.
- [x] New theming system (`useTheme()`).
- [x] Migrate everything to `useTheme()`.
  - [x] AddressField
  - [x] AppBar
  - [x] Badge (replaced by `Tag`)
  - [x] Badge.App (replaced by `Tag`)
  - [x] Badge.Identity (replaced by `Tag`)
  - [x] Badge.Info (replaced by `Tag`)
  - [x] Badge.Notification (replaced by `Tag`)
  - [x] BadgeNumber (will be replaced by `Count`)
  - [x] Checkbox
  - [x] ~CircleGraph~
  - [x] ContextMenu
  - [x] ContextMenuItem
  - [x] DropDown
  - [x] DropDownItem
  - [x] EmptyStateCard
  - [x] Field
  - [x] FloatIndicator
  - [x] ~~IconInfo~~
  - [x] IdentityBadge
  - [x] IdentityBadgePopover
  - [x] Info
  - [x] LineChart
  - [x] LoadingRing
  - [x] PartitionBar
  - [x] Popover
  - [x] ProgressBar
  - [x] Radio
  - [x] RadioList
  - [x] RadioListItem
  - [x] SidePanelSeparator
  - [x] SidePanelSplit
  - [x] Slider
  - [x] Table
  - [x] TableCell
  - [x] TableHeader
  - [x] TextInput
  - [x] Timer
  - [ ] Toast
  - [x] TokenBadge
  - [x] TokenBadgePopover
  - [x] TransactionProgress
- Update gallery pages / READMEs:
  - [x] BackButton
  - [x] Bar
  - [x] Box
  - [x] ButtonIcon, ButtonText
  - [x] CardLayout
  - [x] DataView (provide example)
  - [x] DiscButton
  - [x] Info (provide more modes)
  - [x] Header
  - [x] Layout
  - [x] Main
  - [x] Pagination
  - [x] Split
  - [x] Tabs
  - [x] Theme
  - [x] TokenBadge


src/components/Button/Button.js
```diff
@@ -1,246 +1,99 @@
...
+              css={`
+                position: relative;
+                top: -1px;
+                display: flex;
+              `}
+            >
+              {icon}
+            </span>
+          )}
+          {icon && label && (
```
AquiGorka: Should we allow buttons to display a `label` without an `icon`?

src/components/Button/ButtonBase.js
```diff
@@ -1,54 +1,74 @@
...
-      />
-    )
-  }
-}
+function ButtonBase({
+  focusVisible,
+  showFocusRing,
+  focusRingSpacing,
+  focusRingRadius,
+  innerRef,
```
AquiGorka: It isn't necessary to de-structure these variables if they aren't consumed no? The `{...props}` sends them to `button`

src/components/Layout/Layout.js
```diff
@@ -0,0 +1,63 @@
...
+const SIZES = Object.entries(BREAKPOINTS)
+  .filter(([name]) => name !== 'min')
+  .sort((a, b) => a[1] - b[1])
+
+// Minimum margin around a layouts
+const MIN_MARGIN = 3 * GU
+
+function getLayoutSize(parentWidth) {
+  let index = SIZES.length
+  while (index--) {
```
AquiGorka: Why did you choose to use a `while` over a `reduce` here?

src/utils/styles/gu.js
```diff
@@ -0,0 +1 @@
+export const GU = 8
```
AquiGorka: ✊

src/components/Button/Button.js
bpierre: You can do that, this is only the spacing between the two :+1:

src/components/Button/ButtonBase.js
bpierre: I prefer to always filter out the component props before passing the rest to an element, as they could be part of the element API itself and create subtle bugs or generate warnings in the console.

src/components/Button/Button.js
AquiGorka: One of the reasons I like `grid` so much is `grid-gap`, no need for extra checks, if there are multiple items they get spacing.

src/components/Button/ButtonBase.js
AquiGorka: I agree, that's what I meant when `consuming` them, but here only the first ones are used in `css` - `innerRef` is not used at all, I think you could pass it down directly

src/components/Layout/Layout.js
bpierre: This is so I can interrupt the iteration as soon as I found the correct size. I could have used `Array#find()` with `Array#reverse()` here, but I find the `while` loop to be as readable, with less operations (even if negligible here).

src/components/Button/Button.js
bpierre: Yes and `gap` is coming to Flexbox too, it’s already in Firefox :-)

src/components/Button/ButtonBase.js
bpierre: `innerRef` is used here => https://github.com/aragon/aragon-ui/pull/391/files/d82ed08fddf3717dbc7c9b60909b165cb5166b6c#diff-3a065d36fc270bcf2674de6d9a1a4619R20

Passing it down directly would set the `innerRef` property on the DOM element itself, which is probably fine, but there is also no reason to do that so I’d prefer avoiding it :laughing:

AquiGorka: Ahhh, sorry about that, my confusion came because of the var name, yes, `innerRef` is consumed. Thanks for clarifying

src/components/Button/ButtonBase.js
```diff
@@ -31,7 +31,7 @@ function ButtonBase({
 
         background: none;
         border: 0;
-        border-radius: 4px;
+        border-radius: ${RADIUS}px;
```
AquiGorka: Wonder if we should make this similar to `Grid Unit (GU)` -> `Radius Unit (RU)`

src/components/Card/Card.js
```diff
@@ -1,12 +1,104 @@
...
+
+function Card({ children, width, height, onClick, ...props }) {
+  const theme = useTheme()
+  const insideCardsLayout = useInsideCardsLayout()
+  const interactive = Boolean(onClick)
+
+  const interactiveProps = interactive ? { role: 'button', tabIndex: '0' } : {}
+
+  return (
+    <FocusVisible>
```
AquiGorka: Why are you not using the hook version?

src/components/Button/ButtonBase.js
bpierre: I’m not sure about this, the idea of a grid unit is that we can have any number of it, while the radius is a fixed value.

src/components/Card/Card.js
bpierre: Do we have a hook version? :thinking: :laughing:

src/components/Button/ButtonBase.js
AquiGorka: Ahh, I see what you mean, but regular radius border can be seen as `${1 * RU}px` no?
`GU` is a constant too, we just multiply it by the times we want to use.

src/components/Card/Card.js
AquiGorka: Jajaja

src/components/Button/ButtonBase.js
bpierre: > Ahh, I see what you mean, but regular radius border can be seen as ${1 * RU}px no?

Not really, it’s more a “value” than a “unit”: it should not be multiplied, but used as it is.

luisivan:
needs #346

src/components/Button/ButtonBase.js
bpierre: @AquiGorka resolving this if it’s OK for you!

src/components/ContextMenu/ContextMenu.js
```diff
@@ -91,54 +111,18 @@ const Main = styled(animated.div)`
...
-  top: ${BASE_HEIGHT - 1}px;
-  right: 0;
-  padding: 10px 0;
-  background: ${theme.contentBackground};
-  border: 1px solid ${theme.contentBorder};
-  border-radius: 3px 0 3px 3px;
+  box-shadow: 0px 1px 3px rgba(0, 0, 0, 0.1);
 `
 
 ContextMenu.BASE_WIDTH = 46
```
sohkai: We can use the `BASE_WIDTH` and `BASE_HEIGHT` constants defined above here.

src/components/DropDown/DropDown.js
```diff
@@ -1,163 +1,322 @@
...
+
+    if (buttonRef.current) {
+      buttonRef.current.focus()
+    }
+  }, [setOpened])
+
+  const handleToggle = useCallback(() => setOpened(!opened), [
+    setOpened,
+    opened,
+  ])
```
sohkai: Could we do something like

```js
const handleToggle = useCallback(() => setOpened(opened => !opened), [])
```

Since `useState`'s update function doesn't need to be included in  dependency lists.

src/components/DropDown/DropDown.js
```diff
@@ -1,163 +1,322 @@
...
+    // if the popover is opened and the user clicks on the button
+    // this handler was being called before the click handler, so the
+    // click handler was re-opening the popover, by having this on the
+    // next tick things happen in order.
+    setTimeout(() => setOpened(false), 10)
+
+    if (buttonRef.current) {
+      buttonRef.current.focus()
+    }
+  }, [setOpened])
```
sohkai:
```suggestion
  }, [])
```

`useState`'s update function doesn't need to be included in dependency lists.

src/components/DropDown/DropDown.js
```diff
@@ -1,163 +1,322 @@
...
+
+  useEffect(() => {
+    if (selected === -1 || !items[selected]) {
+      if (placeholder || label) {
+        setSelectedLabel(placeholder || label)
+      }
+      return
+    }
+    setSelectedLabel(items[selected])
+  }, [items, selected, setSelectedLabel, label, placeholder])
```
sohkai:
```suggestion
  }, [items, selected, label, placeholder])
```

src/components/DropDown/DropDown.js
```diff
@@ -1,163 +1,322 @@
...
+
+function useDropDown({
+  buttonRef,
+  items,
+  label,
+  onChange,
+  placeholder,
+  selected,
+}) {
+  const [selectedLabel, setSelectedLabel] = useState(placeholder || label)
```
sohkai: Since `useDropDown()` is internal, we should just drop the `label` argument and only pass in `placeholder`.

src/components/DropDown/DropDown.js
```diff
@@ -1,163 +1,322 @@
...
-            items={opened}
-            from={{ scale: 0.98, opacity: 0, enabled: 1 }}
-            enter={{ scale: 1, opacity: 1, enabled: 1 }}
-            leave={{ scale: 1, opacity: 0, enabled: 0 }}
-            native
+    if (selected !== undefined) {
+      return selected
+    }
+    return -1
+  }, [active, selected])
```
sohkai: It's not a big deal, but we could just use

```js
const selectedIndex = selected || active || -1
```

and skip the memo since this is super cheap. Since `active` is also deprecated, we should also prefer `selected`.

src/components/DropDown/DropDown.js
```diff
@@ -1,163 +1,322 @@
...
+    setButtonWidth(el.clientWidth)
+  })
+
+  // And every time the viewport resizes
+  const { width: vw } = useViewport()
+  useEffect(() => {
+    if (buttonRef.current) {
+      setButtonWidth(buttonRef.current.clientWidth)
+    }
+  }, [vw, buttonRef])
```
sohkai: Do `refs` need to be in the dependency list? I had the impression no, but if so, we should add it elsewhere (e.g. in `useDropDown()`'s hooks).

src/components/Header/Header.js
```diff
@@ -0,0 +1,83 @@
...
+          box-shadow: ${fullWidth ? '0px 2px 3px rgba(0, 0, 0, 0.05)' : 'none'};
+        `}
+      >
+        <div
+          css={`
+            display: flex;
+            align-items: center;
+            justify-content: space-between;
+            height: ${fullWidth ? 8 * GU : 5 * GU}px;
+            padding: 0 ${fullWidth && !children ? 2 * GU : 0}px;
```
sohkai: Should we pass in `...props` here, to allow css overrides?

src/components/Layout/Layout.js
```diff
@@ -0,0 +1,71 @@
...
+        SIZES[index][1] - (index === 0 ? 0 : MIN_MARGIN * 2),
+      ]
+    }
+  }
+  return SIZES[0]
+}
+
+const LayoutContext = React.createContext({})
+
+function useLayout(parentWidth) {
```
sohkai: AFAIK this doesn't need the `parentWidth` argument?

src/components/Layout/Layout.js
```diff
@@ -0,0 +1,71 @@
...
+  const { layoutWidth, layoutName } = useContext(LayoutContext)
+  return {
+    layoutName,
+    layoutWidth,
```
sohkai: Seems really unnecessary to return different names for the same things but 🤷‍♂  I guess you could prefer one over the other in some contexes (for that reason I prefer `layoutName` and `layoutWidth`)

src/components/Layout/Layout.js
```diff
@@ -0,0 +1,71 @@
...
+    name: layoutName,
+    width: layoutWidth,
+  }
+}
+
+function Layout({ children, parentWidth }) {
+  const { width: viewportWidth } = useViewport()
```
sohkai: Not a big deal, but perhaps we can warn if you use `Layout` inside another `Layout`?

src/theme/MainTheme.js
```diff
@@ -0,0 +1,46 @@
...
+      <Theme theme={theme}>{children}</Theme>
+    </MainThemeContext.Provider>
+  )
+}
+
+MainTheme.propTypes = {
+  children: PropTypes.node,
+}
+
+export { MainTheme, useThemeMode }
```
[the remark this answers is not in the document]
bpierre: `MainTheme` is the “theme declared by `Main`”. It was necessary because `Theme` itself doesn’t have a notion of “being the main theme”. It can’t toggle between two themes (it doesn’t have any notion of it), and can be overridden at any level.

It gives the users the possibility to change the theme locally (e.g. in a `Modal`, `Popover`, etc.), while maintaining the possibility to alternate between a dark and light theme at the global level.

src/components/Card/Card.js
```diff
@@ -1,12 +1,113 @@
...
+}
+
+Card.propTypes = {
+  children: PropTypes.node,
+  height: PropTypes.oneOfType([PropTypes.string, PropTypes.number]),
+  width: PropTypes.oneOfType([PropTypes.string, PropTypes.number]),
+  onClick: PropTypes.func,
+}
+
+function FocusRing() {
```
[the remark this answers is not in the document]
bpierre: We might want to share one type of `FocusRing` yes, this one specifically… but the idea is really that each component is responsible of its own focus ring, which can take different shapes and be implemented in different ways.

src/components/Button/ButtonBase.js
```diff
@@ -1,56 +1,79 @@
...
+  showFocusRing,
+  ...props
+}) {
+  const theme = useTheme()
+  return (
+    <button
+      type="button"
+      ref={innerRef}
+      disabled={disabled}
+      {...props}
```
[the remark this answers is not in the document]
bpierre: I started to put `css` last because to make sure `{...props}` wasn’t missed, and I think it could be a good idea to dissociate `css` from the actual props.

I don’t think it has any impact on styled-components: `css` is only used at transform time to create the corresponding styled component, and it gets replaced by the `_css1`, `_css2` etc. series of props.

src/components/Button/ButtonIcon.js
```diff
@@ -1,34 +1,56 @@
...
-import ButtonBase from './ButtonBase'
+import { GU } from '../../style'
+import { Button } from './Button'
+import { ButtonBase } from './ButtonBase'
 
-class ButtonIcon extends React.PureComponent {
-  render() {
-    const { label, children, ...props } = this.props
+function ButtonIcon({ label, children, mode, ...props }) {
+  if (mode === 'button') {
```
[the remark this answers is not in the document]
bpierre: mmm yes maybe… but I wonder if it’s worth breaking the consistency of `mode` only because of an advanced usage. Also this component is not really meant to be used with other button modes, I consider `Button` and `ButtonBase` implementation details for now.

Actually I’m not even sure this mode is needed at all anymore, now that we can achieve the same using [`display` on `Button`](https://github.com/aragon/aragon-ui/pull/432). :laughing:

src/components/Card/Card.js
```diff
@@ -1,12 +1,113 @@
...
+                  .focus-ring {
+                    display: block !important;
+                  }
+                }
+              `}
+          `}
+          {...props}
+        >
+          {children}
+          {focusVisible && <FocusRing />}
```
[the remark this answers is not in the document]
bpierre: Yes I don’t think this is needed, I don’t think there is any way `focusVisible` could be `true` if the `<div>` itself is not focusable (kind of the same way we don’t check if a button is disabled, only if it’s focused).

src/components/CardLayout/CardLayout.js
```diff
@@ -0,0 +1,53 @@
+import React, { useContext } from 'react'
+import PropTypes from 'prop-types'
+import { GU } from '../../style'
+import { useLayout } from '../Layout/Layout'
+
+const CardLayoutContext = React.createContext(false)
+
+// Returns true if a component is declared inside of the CardLayout tree
+function useInsideCardLayout() {
```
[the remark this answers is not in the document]
bpierre: Yes! Forgot this one :+1:

src/components/Header/Header.js
bpierre: Yes it doesn’t hurt!

src/components/Layout/Layout.js
bpierre: Nope :+1:

src/components/Layout/Layout.js
bpierre: Ah yes, at the beginning there was only `name` and `width`, that were renamed every time they `useLayout()` was used, so I adedd `layoutName` and `layoutWidth`.

`name` and `width` are deprecated but I didn’t want to remove them now because it’s quite a central component that people have started to use. I will add a comment, and I’ll make them getters to warn when they are used.

src/components/Layout/Layout.js
bpierre: I need to add a mechanism on `Inside` to stop it at portals (so we can have a `Layout` in a `Modal` opened from another `Layout`), but yes it’s something we could add :-)

src/components/DropDown/DropDown.js
bpierre: > It's not a big deal, but we could just use

It would break with `0`.

> and skip the memo since this is super cheap

The memo is just an excuse to have a function with returns here, and avoid nested ternaries!

> Since active is also deprecated, we should also prefer selected.

Yes :+1:

src/components/DropDown/DropDown.js
bpierre: No I don’t think it’s needed (except when `.current` is used), `useRef()` should always return the same object.

src/components/DropDown/DropDown.js
bpierre: https://github.com/aragon/aragon-ui/pull/461

src/components/DropDown/DropDown.js
bpierre: https://github.com/aragon/aragon-ui/pull/461

src/components/DropDown/DropDown.js
bpierre: https://github.com/aragon/aragon-ui/pull/461

src/components/DropDown/DropDown.js
bpierre: Done https://github.com/aragon/aragon-ui/pull/461

src/components/DropDown/DropDown.js
bpierre: Done https://github.com/aragon/aragon-ui/pull/461

src/components/CardLayout/CardLayout.js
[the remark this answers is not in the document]
sohkai: Done in https://github.com/aragon/aragon-ui/pull/454.

src/components/DataView/DataView.js
```diff
@@ -0,0 +1,313 @@
...
+  const selectAll = useCallback(() => {
+    setSelectedIndexes(
+      selectedIndexes.length === 0 ? entries.map((_, index) => index) : []
+    )
+  }, [entries, selectedIndexes])
+
+  useEffect(() => {
+    if (onSelectEntries) {
+      onSelectEntries(
+        entries.filter((entry, index) => selectedIndexes.includes(index)),
```
sohkai: This is unlikely to really matter, but an optimization to avoid n^2 complexity is:

```js
selectedIndexes.sort().map(index => entries[index])
```

src/components/DataView/DataView.js
```diff
@@ -0,0 +1,313 @@
...
+        selectedIndexes
+      )
+    }
+  }, [onSelectEntries, selectedIndexes])
```
sohkai: Does this also need to include `entries`?

src/components/DataView/DataView.js
```diff
@@ -0,0 +1,313 @@
...
+  renderEntry,
+  renderEntryActions,
+  renderEntryChild,
+  renderSelectionCount,
+  mode,
+  tableRowHeight,
+}) {
+  if (renderEntryChild && onSelectEntries) {
+    throw new Error(
+      'A DataView cannot be made selectable if it has entry children. ' +
```
sohkai:
```suggestion
      'A DataView cannot be made selectable if its entries have children. ' +
```

src/components/DataView/DataView.js
```diff
@@ -0,0 +1,313 @@
...
+        'Please use either renderEntryChild or onSelectEntries.'
```
sohkai:
```suggestion
        'Please use only one of renderEntryChild or onSelectEntries at a time.'
```

src/components/DataView/DataView.js
```diff
@@ -0,0 +1,313 @@
...
+    )
+  }
+
+  if (currentPage !== undefined) {
```
sohkai: If `currentPage` is already deprecated we should just remove it

src/components/DataView/DataView.js
```diff
@@ -0,0 +1,313 @@
...
+    throw new Error('DataView: please use `page` instead of `currentPage`.')
+  }
+
+  // Only used if currentPage is not passed. The pagination supports both a
```
sohkai:
```suggestion
  // Only used if page is not passed. The pagination supports both a
```

src/components/DataView/DataView.js
```diff
@@ -0,0 +1,313 @@
...
+  // managed and a controlled mode, to provide a better developer experience
+  // out of the box.
+  const [pageManaged, setPageManaged] = useState(0)
+
+  const handlePageChange = useCallback(
+    newPage => {
+      // Managed state
+      if (page === undefined) {
+        console.log('SET', newPage)
```
sohkai: Let's not forget to remove these :)

src/components/DataView/DataView.js
```diff
@@ -0,0 +1,313 @@
...
+        setPageManaged(newPage)
+      }
+
+      // Useful even to notify, even in managed mode
```
sohkai:
```suggestion
      // Useful to notify, even in managed mode
```

src/components/DataView/DataView.js
```diff
@@ -0,0 +1,313 @@
...
+        renderEntryChild
+          ? renderEntryChild(value, index, { selected, mode })
+          : null
+      ),
+      selected,
+    }
+  })
+}
+
+function useSelection(entries, onSelectEntries) {
```
sohkai: It might be nice to split this out at some point into its own exposed hook

src/components/DataView/DataView.js
```diff
@@ -0,0 +1,313 @@
...
+        <div
+          css={`
+            padding: ${2 * GU}px ${3 * GU}px;
+          `}
+        >
+          {typeof heading === 'string' ? (
+            <h1
+              css={`
+                font-size: 16px;
+                font-weight: 400;
```
sohkai: We should probably use a text style here

src/components/DataView/README.md
```diff
@@ -0,0 +1,302 @@
...
+
+### mode
+
+| Type                              | Default value |
+| --------------------------------- | ------------- |
+| `"adaptive"`, `"table"`, `"list"` | `"adaptive"`  |
+
+Set this to force the display mode of DataView.
+
+### currentPage
```
sohkai:
```suggestion
### page
```

And we should note this prop is not required, and if it's not provided, `DataView` will manage its own state.

src/components/DataView/README.md
```diff
@@ -0,0 +1,302 @@
...
+| -------- | --------------- |
+| `Number` | None (required) |
+
+Index of the current page.
+
+### onPageChange
+
+| Type       | Default value   |
+| ---------- | --------------- |
+| `Function` | None (required) |
```
sohkai: This does not seem to be required anymore

src/components/DataView/DataView.js
```diff
@@ -0,0 +1,313 @@
...
+      )}
+    </Box>
+  )
+})
+
+DataView.propTypes = {
+  page: PropTypes.number,
+  entries: PropTypes.array.isRequired,
+  entriesPerPage: PropTypes.number,
+  fields: PropTypes.array,
```
sohkai:
```suggestion
  fields: PropTypes.array.isRequired,
```

Not providing `fields` will cause an error

src/components/DataView/ListView.js
```diff
@@ -0,0 +1,201 @@
...
+            {entry.actions && (
+              <div
+                css={`
+                  position: absolute;
+                  top: ${2 * GU}px;
+                  right: ${3 * GU}px;
+                `}
+              >
+                {entry.actions}
+              </div>
```
sohkai: cc @dizzypaty the alignment between the checkbox / toggle button and actions feels a bit off, since they're using different top positions (12px vs. 16px):

Screen Shot 2019-08-13 at 6 32 28 PM

This would look even worse with the checkbox, because it's smaller.

src/components/DataView/ListView.js
```diff
@@ -0,0 +1,201 @@
...
+            )}
```
sohkai: Another problem is with the z-indices we use; currently the ContextMenu collides with the full-width Tabs:

Screen Shot 2019-08-13 at 6 34 09 PM

src/components/DataView/README.md
```diff
@@ -0,0 +1,302 @@
...
+  fields={[
+    // `priority` allows to change the fields
+    // order when DataView is in list mode.
+    { label: 'Account', priority: 1 },
+    { label: 'Amount', priority: 2 },
+  ]}
+/>
+```
+
+#### field.label
```
sohkai: We should also document `field.align`

src/components/DataView/TableView.js
```diff
@@ -0,0 +1,415 @@
...
+
+  // Toggle child
+  if (hasAnyChild) {
+    cells.unshift([hasChildren && toggleChildContent, 'start', true])
+  }
+
+  // Checkbox
+  if (selectable && !hasAnyChild) {
+    cells.unshift([selectable && selectContent, 'start', true])
+  }
```
sohkai: We should reverse the priorities of `hasAnyChild` and `selectable` to be in sync with `ListView`.

src/components/DataView/TableView.js
```diff
@@ -0,0 +1,415 @@
...
+          />
+        ),
+        selectable,
+      }),
+    [fields, hasAnyChild, hasAnyActions]
+  )
+
+  useEffect(() => {
+    setOpened(false)
+  }, [entries, fields])
```
sohkai: This feels like a weird place to put this hook; maybe it can go after `entryRows`?

src/components/DataView/TableView.js
```diff
@@ -0,0 +1,415 @@
...
+          selectContent: selectable ? (
+            <Select
+              index={entry.index}
+              selected={entry.selected}
+              onSelect={onSelect}
+            />
+          ) : null,
+        })
+      }),
+    [opened, entries, hasAnyChild, hasAnyActions, toggleEntry]
```
sohkai: This seems to be missing a few dependencies

devbox/apps/Wrappers.js
```diff
@@ -0,0 +1,109 @@
...
+function Header1() {
+  return (
+    <Header
+      primary="Voting"
+      secondary={<Button mode="strong" label="New vote" icon={<IconPlus />} />}
+    />
+  )
+}
+
+function Bar1() {
```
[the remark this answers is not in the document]
bpierre: It’s just one type of bar, I think it’s fine to keep them neutral using Bar1, Bar2, etc. so we can change what’s inside when we develop.

devbox/apps/Wrappers.js
```diff
@@ -0,0 +1,109 @@
...
+
+function Bar3() {
+  return <Bar primary={<BackButton />} secondary={<Button label="Share" />} />
+}
+
+function Cards1({ interactive }) {
+  const cardProps = interactive ? { onClick: () => {} } : {}
+  return (
+    <CardLayout>
+      {[...Array(10)].map((v, i) => (
```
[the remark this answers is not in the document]
bpierre: Yes!

src/components/Button/ButtonBase.js
```diff
@@ -1,56 +1,79 @@
...
+      css={`
+        position: relative;
+        display: inline-block;
+        padding: 0;
+        white-space: nowrap;
+        ${font({ size: 'small', weight: 'normal' })};
+        ${unselectable};
+        color: ${theme.textSecondary};
```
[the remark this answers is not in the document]
bpierre: This made me think that it could be nice to warn when we try to use a non-existing entry from the theme. It would mean adding a getter, which might have an impact on performances… but it will probably not be perceptible − we’ll measure it anyway. Also it would allow us to warn about deprecated colors, which would be super useful.

src/components/Tabs/TabsFullWidth.js
```diff
@@ -0,0 +1,191 @@
...
+
+  const close = useCallback(() => {
+    setOpened(false)
+  }, [])
+
+  const toggle = useCallback(() => {
+    setOpened(opened => !opened)
+  }, [])
+
+  const blur = useCallback(event => {
```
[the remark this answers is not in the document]
bpierre: Yes :+1: 

Also I think we should rename `useOnBlur()` to be `useFocusLeave()` and add `useFocusEnter()`, to mimic `mouseenter` / `mouseleave` since it works in a similar way. /cc @AquiGorka

src/components/Tabs/TabsFullWidth.js
```diff
@@ -0,0 +1,191 @@
...
+            `}
+          >
+            {selectedItem}
+          </div>
+          <div
+            css={`
+              display: flex;
+              align-items: center;
+              padding: 0 ${2 * GU}px;
+              color: ${theme.surfaceIcon};
```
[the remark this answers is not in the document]
bpierre: Yes adding it. And yes it has been removed from the DropDown because of the vertical centering.

src/components/DataView/DataView.js
bpierre: Nice! :trophy:

src/components/DataView/DataView.js
bpierre: Oops

src/components/DataView/DataView.js
bpierre: Yes agreed

bpierre: https://github.com/aragon/aragon-ui/issues/473

src/components/DataView/TableView.js
bpierre: Well spotted!

src/components/Button/ButtonBase.js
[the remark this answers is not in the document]
sohkai: This is also something we could only do in development mode, where we export a wrapped version that has a getter.

src/components/Tabs/Tabs.js
```diff
@@ -0,0 +1,198 @@
...
+  // Use a separate component for Tabs in AppBar, to prevent breaking anything.
+  if (insideAppBar) {
+    return <TabBarLegacy {...props} inAppBar />
+  }
+
+  if (layoutName === 'small') {
+    return <TabsFullWidth {...props} />
+  }
+
+  if (insideBar) {
```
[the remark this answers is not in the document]
bpierre: https://github.com/aragon/aragon-ui/pull/474

src/components/Tabs/TabsFullWidth.js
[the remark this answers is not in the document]
bpierre: https://github.com/aragon/aragon-ui/pull/474

src/components/Tabs/TabsFullWidth.js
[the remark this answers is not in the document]
bpierre: useFocusEnter

src/components/DataView/README.md
```diff
@@ -0,0 +1,302 @@
...
+    return [<IdentityBadge entity={account} />, <Amount>{amount}</Amount>]
+  }}
+/>
+```
+
+### heading
+
+| Type         | Default value |
+| ------------ | ------------- |
+| `React node` | `10`          |
```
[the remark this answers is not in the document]
bpierre: https://github.com/aragon/aragon-ui/pull/474

devbox/apps/Wrappers.js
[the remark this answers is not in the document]
bpierre: https://github.com/aragon/aragon-ui/pull/474

src/components/Button/ButtonBase.js
```diff
@@ -1,56 +1,79 @@
 import React from 'react'
 import PropTypes from 'prop-types'
-import styled from 'styled-components'
 import FocusVisible from '../FocusVisible/FocusVisible'
-import { theme } from '../../theme'
-import { unselectable, font } from '../../utils/styles'
+import { useTheme } from '../../theme'
+import { RADIUS } from '../../style'
+import { unselectable, font } from '../../utils'
```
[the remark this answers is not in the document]
bpierre: https://github.com/aragon/aragon-ui/pull/474

src/components/Button/ButtonBase.js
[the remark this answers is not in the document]
bpierre: https://github.com/aragon/aragon-ui/issues/474 and https://github.com/aragon/aragon-ui/issues/476 (for the warning)

src/components/ContextMenu/ContextMenu.js
bpierre: https://github.com/aragon/aragon-ui/issues/474

src/components/Header/Header.js
bpierre: https://github.com/aragon/aragon-ui/issues/474

src/components/DataView/DataView.js
bpierre: https://github.com/aragon/aragon-ui/issues/474

src/components/DataView/DataView.js
bpierre: Yes! https://github.com/aragon/aragon-ui/issues/474

src/components/DataView/DataView.js
bpierre: https://github.com/aragon/aragon-ui/issues/474

src/components/DataView/DataView.js
bpierre: https://github.com/aragon/aragon-ui/issues/474

src/components/DataView/README.md
bpierre: https://github.com/aragon/aragon-ui/issues/474

src/components/DataView/README.md
bpierre: https://github.com/aragon/aragon-ui/issues/474

src/components/DataView/ListView.js
bpierre: Fixed in https://github.com/aragon/aragon-ui/issues/474

src/components/DataView/TableView.js
bpierre: https://github.com/aragon/aragon-ui/issues/474

src/components/DataView/TableView.js
bpierre: https://github.com/aragon/aragon-ui/issues/474

src/components/DataView/README.md
bpierre: https://github.com/aragon/aragon-ui/issues/474

src/components/DataView/TableView.js
bpierre: https://github.com/aragon/aragon-ui/issues/474

src/components/DataView/DataView.js
bpierre: https://github.com/aragon/aragon-ui/issues/474

src/components/DataView/DataView.js
bpierre: https://github.com/aragon/aragon-ui/issues/474

src/style/text-styles.js
```diff
@@ -0,0 +1,85 @@
...
+  weight: value => `font-weight: ${value};`,
+  transform: value => `text-transform: ${value};`,
+  lineHeight: value => `line-height: ${value};`,
+}
+
+function textStyleToCss(textStyle) {
+  return Object.entries(textStyle).reduce(
+    (cssBlock, [name, value]) => cssBlock + STYLES_CSS[name](value),
+    ''
+  )
```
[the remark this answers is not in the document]
bpierre: https://github.com/aragon/aragon-ui/issues/474

src/components/Button/ButtonIcon.js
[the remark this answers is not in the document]
bpierre: I deprecated `mode` on `ButtonIcon` https://github.com/aragon/aragon-ui/issues/474

src/components/DataView/ListView.js
sohkai: cc @dizzypaty the alignment between the checkbox / toggle button and actions feels a bit off, since they're using different top positions (12px vs. 16px):

Screen Shot 2019-08-13 at 6 32 28 PM

This would look even worse with the checkbox, because it's smaller.

dizzypaty: Both the toggle icon and the contextual actions menu look aligned to me 🤔 with 2GU top margin:

Group 121

bpierre:
Closing for the cleaner https://github.com/aragon/aragon-ui/pull/662 :fire:

Issue #149, opened 2018-03-30
Autocomplete component


onbjerg:
It would be nice to have an autocomplete feature for e.g. address fields, displaying different entities and apps in and outside your organisation.

I imagine something like this (excuse the non-jpg):

```
╔══════════════════════════════════╗
║ 0xdeadbeef                       ║ <- What you type
╞══════════════════════════════════╡
│ 0xdeadbeef                       │ <- Autocomplete of what you type
├──────────────────────────────────┤
│ APPS                             │ <- Label
│ Permissions (0xddadbeef)         │ <- Apps in DAO
├──────────────────────────────────┤
│ ENTITIES                         │ <- Label
│ satoshi.aragonid.eth (0xdddbeef) │ <- Interesting entities in DAO
└──────────────────────────────────┘
```

izqui:
I love this. This could totally be part of the address input component we have been talking about. We should talk about how we could build the backend for this is aragon.js, which would require full, both ways, aragon-id support.

onbjerg:
Yeah. I thought about wanting to do the permissions app today, but in the mockup there's a lack of flexibility for the input fields - they're all dropdowns, but should probably be a hybrid between an autocomplete field and a dropdown, or at least only autocomplete.

luisivan:
Closing this since the PR is already sent and this issue doesn't add much now being open

PR #394, opened 2019-06-12
Autocomplete

AquiGorka:
Ref: https://github.com/aragon/aragon-ui/issues/149


luisivan:
@bpierre are we gonna get this merged?

src/components/AutoComplete/AutoCompleteSelected.js
```diff
@@ -0,0 +1,110 @@
...
+  wide,
+}) {
+  const ref = forwardedRef
+  const selectedRef = useRef()
+
+  const handleSelect = useCallback(
+    selected => {
+      onSelect(selected)
+      setTimeout(() => {
+        selectedRef.current.focus()
```
sohkai: I wonder if we should do a `selectedRef.current` check here; we're more or less assuming the `onSelect()` handler is going to change `props.selected`.

AquiGorka: Added it here: I'll work on it as part of an update I need to work on this component

PR #397, opened 2019-07-01
Devbox fix

chadoh:
Fixes 3599f33afac53b6b1d979dbb3c1f2dc3154d5736

Import style changed and `Scratchpad` is no longer defined in `devbox/index.js`

### DIFF
--- a/devbox/index.js
+++ b/devbox/index.js
@@ -24,7 +24,7 @@ class Index extends React.Component {
   }
   render() {
     const { appName } = this.state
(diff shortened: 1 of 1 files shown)


bpierre:
Thanks Chad!

Could you also filter it out from the list so it doesn’t appear twice? (or sort the list so that it appears first)

```jsx
            {Object.keys(APPS)
              .sort()
              .filter(appName => appName !== 'Scratchpad')
              .map(appName => (

```

chadoh:
Ah, thanks @bpierre. Updated.

PR #398, opened 2019-07-01
Devbox: add react-axe

chadoh:
This adds [react-axe](https://github.com/dequelabs/react-axe) to `devbox` so that we can catch some a11y problems early in the development lifecycle.

It may be useful to have a way to opt out of this using a command line option. Or would opt-in be better?

With `react-axe` enabled, here's what shows up in the console when loading the index page:

Screen Shot 2019-07-01 at 16 11 11

And here's what shows up when looking at `EmptyStateCard`:

Screen Shot 2019-07-01 at 16 11 11

Let me know if any of these should be fixed as part of this PR.


bpierre:
Super useful! :boom: 

> It may be useful to have a way to opt out of this using a command line option. Or would opt-in be better?

I would prefer having it opt-in for `devbox`, so that developing on a component in this environment stays as bare bones as possible. Maybe we could have it as `npm start:a11y-checks`?

> Let me know if any of these should be fixed as part of this PR.

We can fix these later, especially if we make it optional :+1:

chadoh:
@bpierre I added a `run start:a11y-check` script. Let me know if you know a more minimal way to get it in there, other than using `--env`.

PR #401, opened 2019-07-03
Add JSON plugin for rollup

### LINKED ISSUE #400: Rollup configuration causes build scripts to fail, requires JSON plugin
Here is the job that failed when trying to add the Markdown component: https://travis-ci.org/aragon/aragon-ui/jobs/553950501 

It's complaining about not knowing how to handle JSON files. I run into the same issue on my local machine.

gh1dra:
Fixes #400

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9131,6 +9131,32 @@
         "terser": "^3.10.0"
       }
     },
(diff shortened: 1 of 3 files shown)


gh1dra:
Issue addressed: #400

PR #407, opened 2019-07-15
Info: use theme

AquiGorka:
Updated to use [these values](https://www.figma.com/file/WN8Yuc544XoH5NosMBPclTBx/Info-components?node-id=0%3A1)

image


src/theme/theme-light.js
```diff
@@ -27,6 +27,10 @@ export default {
 
   surfaceInteractive: colors.White,
 
+  info: colors.AragonBlue,
+  infoSurfaceContent: colors.AragonBlueDark,
+  infoSurface: colors.AragonBlueLight,
```
bpierre: I just updated the colors (with these exact names :ok_hand: for `info`) in both themes: https://github.com/aragon/aragon-ui/commit/b12d1f9a89fd699cb757157ce8827e3fc2b1ab8d

AquiGorka: I guess I'll remove this, we could've merged this instead...

bpierre: Yes sorry, we reworked several colors at the same time, and this group only happened to use the same names at the end. :man_shrugging:

PR #408, opened 2019-07-15
Toast: use theme + shift position

AquiGorka:
- Updates spacing, background color and margins
- Adds new prop to add room for extra space between the edges (useful when rendering the help scout bubble)


src/components/ToastHub/ToastHub.js
```diff
@@ -191,16 +197,15 @@ const Message = styled(animated.div)`
...
   color: white;
-  background: #445159;
-  opacity: 0.9;
+  background: rgba(48, 64, 79, 0.8);
   margin-top: ${props => (props.top ? '0' : '10px')};
   margin-bottom: ${props => (props.top ? '10px' : '0')};
-  padding: 12px 22px;
-  font-size: 1em;
+  padding: 15px 20px;
+  font-size: 14px;
```
bpierre: I can’t find a [text style](https://github.com/aragon/aragon-ui/blob/newstyle/src/style/text-styles.js) for this :thinking: 

@owisixseven @dizzypaty Do we need a “Body 3” style for the Toast component (and maybe other places in the future)? The font size is `14px` here, but the minimum we have in the text styles is `16px` with “Body 2” (or `14px` uppercase for “Label 1”).

dizzypaty: Mmm... true! Yes, I think we do actually need it! 

## Body 3
Size: `14px` 
Weight: Light (400)
LineHeight: 1.5

Alternatively, we could use Body 2 and make it Regular in this instance? I think that can work too : )

src/components/ToastHub/ToastHub.js
```diff
@@ -157,11 +161,13 @@ ToastList.propTypes = {
...
   z-index: 1000;
-  top: ${props => (props.top ? '30px' : 'unset')};
-  bottom: ${props => (props.top ? 'unset' : '30px')};
+  top: ${props => (props.top ? `${3 * GU}px` : 'unset')};
+  bottom: ${props => (props.top ? 'unset' : `${3 * GU}px`)};
   margin: 0 auto;
-  left: 30px;
-  right: 30px;
+  left: ${({ extraMargin }) =>
+    `calc(${3 * GU}px + ${extraMargin ? `${extraMargin}px` : '0px'})`};
```
bpierre: @AquiGorka What would you think about renaming this prop `shift`?

src/components/ToastHub/ToastHub.js
bpierre: @AquiGorka `body3` is here: https://github.com/aragon/aragon-ui/commit/9cb6d751170e5796c702c095ba94a6738e16789a

PR #413, opened 2019-07-24
New Dropdown style

AquiGorka:


src/components/DropDown/_DropDown.js
```diff
@@ -0,0 +1,178 @@
...
+          )}
+          <ul
+            css={`
+              margin: 0;
+              list-style: none;
+              width: 100%;
+            `}
+          >
+            {items.map((item, index) => {
+              const handleChangeWithIndex = handleChange(index)
```
bpierre: A problem here is that the function will get recreated every time, making it impossible to memoize individual items.

It is generally recommended to follow this pattern instead:

```jsx
function MyComponent({ items, selected, onChange }) {
  return (
    <ul>
      {items.map((item, index) => (
        <MyItem
          index={index}
          selected={index === selected}
          onChange={onChange}
        >
          {item}
        </MyItem>
      ))}
    </ul>
  )
}

function MyItem ({ index, onChange, children }) {

  const handleClick = useCallback(() => {
    onChange(index)
  }, [onChange, index])

  return (
    <li onClick={handleClick}>{children}</li>
  )
}
```

src/components/DropDown/_DropDown.js
```diff
@@ -0,0 +1,178 @@
...
+}
+
+function DropDown({ activeIndex, items, label, header, onChange, width }) {
+  const {
+    accent,
+    content,
+    surfaceContentSecondary,
+    surfacePressed,
+    surfaceSelected,
+  } = useTheme()
```
bpierre: Just for consistency with the other components using the theme, could we avoid destructing the object returned by `useTheme()`? And use `theme.accent`, `theme.content` etc.

src/components/DropDown/_DropDown.js
```diff
@@ -0,0 +1,178 @@
...
+  )
+}
+
+DropDown.propTypes = {
+  activeIndex: PropTypes.number.isRequired,
+  items: PropTypes.array.isRequired,
+  label: PropTypes.string.isRequired,
+  header: PropTypes.string,
+  onChange: PropTypes.func.isRequired,
+  width: PropTypes.string,
```
bpierre: Is there a reason to not have `wide` anymore? If yes, we should at least display a warning in the console.

src/components/DropDown/_DropDown.js
```diff
@@ -0,0 +1,178 @@
...
+              margin: 0;
+              list-style: none;
+              width: 100%;
+            `}
+          >
+            {items.map((item, index) => {
+              const handleChangeWithIndex = handleChange(index)
+
+              return (
+                <li key={item}>
```
bpierre: `item` can be any React node (I also commented on the prop type), so we should probably use the index here − which should be safe since we don’t need to update the DOM nodes themselves.

src/components/DropDown/_DropDown.js
```diff
@@ -0,0 +1,178 @@
...
+        </div>
+      </Popover>
+    </div>
+  )
+}
+
+DropDown.propTypes = {
+  activeIndex: PropTypes.number.isRequired,
+  items: PropTypes.array.isRequired,
+  label: PropTypes.string.isRequired,
```
bpierre: We should probably default to something like “Select an item” rather than making the label required. /cc @dizzypaty

src/components/DropDown/_DropDown.js
```diff
@@ -0,0 +1,178 @@
...
+  header: PropTypes.string,
```
bpierre: Should `header` be a React node? :thinking:

src/components/DropDown/_DropDown.js
```diff
@@ -0,0 +1,178 @@
...
+    // if the popover is opened and the user clicks on the button
+    // this handler was being called before the click handler, so the
+    // click handler was re-opening the popover, by having this on the
+    // next tick things happen in order.
+    setTimeout(() => setOpened(false), 10)
+  }
+  const handleToggle = () => setOpened(!opened)
+  const handleChange = index => () => {
+    onChange(index)
+    handleClose()
```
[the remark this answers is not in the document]
AquiGorka: Wait, this is how I've been doing things all around for some time now :(
We should reach agreement so not to have to change a lot of code

src/components/DropDown/_DropDown.js
```diff
@@ -0,0 +1,178 @@
...
+        opener={containerRef.current}
+        placement="bottom-start"
+      >
+        <div
+          css={`
+            width: calc(${width} - 2px);
+            color: ${surfaceContentSecondary};
+          `}
+        >
+          {!!header && (
```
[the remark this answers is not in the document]
AquiGorka: If it is not a `Boolean` won't this render the header?

src/components/DropDown/_DropDown.js
AquiGorka: Sure, I think I saw this elsewhere and I do like destructuring but I can adhere to the convention :)

src/components/DropDown/_DropDown.js
AquiGorka: Because of the `Popover`, if I want it to have the same width as the container it needs to be set - we can't set the `Popover`s width to `100%` as it is not inside the container

src/components/DropDown/_DropDown.js
AquiGorka: Sure, this first version was meant only for strings but I can uppdate

AquiGorka:
@bpierre I'll work on the changes and then ping you to let you know which ones have been updated and which ones we can work on after mreging

src/components/DropDown/_DropDown.js
```diff
@@ -0,0 +1,178 @@
...
+  const [selectedLabel, setSelectedLabel] = useState(label)
+  const [opened, setOpened] = useState(false)
+  const handleClose = () => {
+    // if the popover is opened and the user clicks on the button
+    // this handler was being called before the click handler, so the
+    // click handler was re-opening the popover, by having this on the
+    // next tick things happen in order.
+    setTimeout(() => setOpened(false), 10)
+  }
+  const handleToggle = () => setOpened(!opened)
```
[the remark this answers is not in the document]
AquiGorka: This is what I meant the other day about using the `handle` prefix, this handler is responsible for the toggling not for a button click and it is part of the hook and gets exposed.

src/components/DropDown/_DropDown.js
AquiGorka: TIL 😄

AquiGorka:
@bpierre updates have been pushed, if it is alright with you we can merge this and continue the conversation on the open comments here.

bpierre:
@AquiGorka Cool! Yes let’s merge and prerelease it :shipit:

PR #414, opened 2019-07-25
Unprefix DropDown

bpierre:
Also made a few minor changes:

- Fix an issue where `0` was not accepted on `active`
- Spacing
- Warn only in development, and once


src/components/DropDown/DropDown.js
```diff
@@ -1,164 +1,236 @@
...
+import React, { useState, useRef, useEffect, useCallback } from 'react'
 import PropTypes from 'prop-types'
-import styled from 'styled-components'
-import { Transition, animated } from 'react-spring'
-import ClickOutHandler from 'react-onclickout'
-import { theme } from '../../theme-legacy'
-import { springs } from '../../style'
-import { unselectable } from '../../utils'
-import { PublicUrl } from '../../providers/PublicUrl'
-import DropDownItem from './DropDownItem'
```
sohkai: Should we also remove `DropDownItem`?

PR #415, opened 2019-07-25
Empty state card new styles

AquiGorka:
Currently using it for Voting:


src/components/Card/_EmptyStateCard.js
```diff
@@ -0,0 +1,49 @@
...
+      <Button onClick={onClick} wide mode="strong">
+        {actionText}
+      </Button>
+    </Card>
+  )
+})
+
+EmptyStateCard.propTypes = {
+  actionText: PropTypes.string,
+  icon: PropTypes.oneOfType([PropTypes.element, PropTypes._component]),
```
bpierre: Maybe we could drop support for components and use `PropTypes.node` directly?

AquiGorka:
@bpierre done 😄

src/components/Card/EmptyStateCard.js
```diff
@@ -1,151 +1,53 @@
...
+        <Button onClick={onClick} wide mode="strong">
+          {actionText}
+        </Button>
+      )}
+    </Card>
+  )
+})
+
+EmptyStateCard.propTypes = {
+  actionButton: PropTypes.bool,
```
bpierre: @AquiGorka What would you think of removing this prop, and using `actionText` or `onClick` to know this?

```jsx
const actionButton = Boolean(actionText || onClick)
```

src/components/Card/EmptyStateCard.js
```diff
@@ -1,151 +1,53 @@
...
+  actionText: PropTypes.string,
+  icon: PropTypes.node.isRequired,
+  onClick: PropTypes.func.isRequired,
```
bpierre:
```suggestion
  onClick: PropTypes.func,
```

src/components/Card/EmptyStateCard.js
```diff
@@ -1,151 +1,53 @@
...
+          {actionText}
+        </Button>
+      )}
+    </Card>
+  )
+})
+
+EmptyStateCard.propTypes = {
+  actionButton: PropTypes.bool,
+  actionText: PropTypes.string,
```
bpierre: Should we rename this to `buttonLabel` or `actionLabel` (since we are using `label` for `Button`, `ButtonIcon` etc.)?

src/components/Card/EmptyStateCard.js
```diff
@@ -1,151 +1,53 @@
...
+  icon: PropTypes.node.isRequired,
+  onClick: PropTypes.func.isRequired,
+  text: PropTypes.node,
```
bpierre: @AquiGorka we used to have a default for this, but I think we should make it required as app authors should change it anyway. What do you think?

```suggestion
  text: PropTypes.node.isRequired,
```

gallery/src/pages/PageEmptyStateCard.js
```diff
@@ -9,21 +9,22 @@ const PageEmptyStateCard = ({ title }) => (
...
+              css={`
+                margin: auto;
+                width: 140px;
+                height: auto;
+              `}
+            />
+          }
+          actionButton={true}
+          actionText="Click me"
+          onClick={() => console.log('Button was clicked')}
```
sohkai: @AquiGorka I think we may have forgotten to update this gallery example with the new API.

AquiGorka: https://github.com/aragon/aragon-ui/pull/428

PR #418, opened 2019-07-26
Added render prop to SidePanel

### LINKED ISSUE #347: SidePanel: add render prop with the open status
See https://github.com/aragon/aragon-apps/pull/717#discussion_r272848595.

Nested components to the `SidePanel` will likely want to do different things based on whether or not it's opened.

kushagrasharma:
Resolves #347 

Creates a render prop and renders it over the children if present, else renders the children.

I was uncertain what the behavior should be if both a render prop and children are present, and chose to prioritize the render prop. Should the children be rendered as well?

### DIFF
--- a/src/components/SidePanel/SidePanel.js
+++ b/src/components/SidePanel/SidePanel.js
@@ -29,7 +29,7 @@ class SidePanel extends React.PureComponent {
     this.props.onTransitionEnd(this.props.opened)
   }
   renderIn = ({ progress }) => {
-    const { children, title, opened, blocking } = this.props
+    const { children, render, title, opened, blocking } = this.props
 
     return (
       <Main opened={opened}>
@@ -60,7 +60,7 @@ class SidePanel extends React.PureComponent {
             )}
           </PanelHeader>
           <PanelScrollView>
-            <PanelContent>{children}</PanelContent>
+            <PanelContent>{render ? render(opened) : children}</PanelContent>
           </PanelScrollView>
         </Panel>
       </Main>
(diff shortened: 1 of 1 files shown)


src/components/SidePanel/SidePanel.js
```diff
@@ -60,7 +60,7 @@ class SidePanel extends React.PureComponent {
             )}
           </PanelHeader>
           <PanelScrollView>
-            <PanelContent>{children}</PanelContent>
+            <PanelContent>{render ? render(opened) : children}</PanelContent>
```
sohkai:
```suggestion
            <PanelContent>{render ? render({ opened }) : children}</PanelContent>
```

Perhaps this would be a nicer API, thoughts @bpierre?

bpierre: Yes definitely, just in case we pass something else in the future!

Some other thoughts:

- Should we pass a `status` directly, which could be either `opening`, `closing`, `opened` or `closed`? This last one will be removed when we’ll move to a `Transition` rather than a `Spring`.
- What would you guys think of detecting if `children` is a function, and calling it in that case? That way we wouldn’t need a `render` prop behaving differently than `children`.
- Should we move to a `Transition` directly?

kushagrasharma: Makes sense! 

For thoughts:
* passing a `status` directly makes sense
* react doesn't allow passing functions as children, so will have to stick with render prop
* lets do it, i'll move to a transition to finish off the PR

bpierre: Hey @kushagrasharma, did you get any time to work on these? We could also give you a hand to get this PR ready, let me know what you prefer!

bpierre:
Closing this PR as it has been replaced by https://github.com/aragon/aragon-ui/pull/418

PR #423, opened 2019-07-30
Inside / useInside()

bpierre:
This is used to know when a component is inside another one.

Usage:

```jsx
import { Inside } from '@aragon/ui'

function MyParentComponent() {
  return (
    <Inside name="MyParentComponent">
      <div>
        <MyChildComponent />
      </div>
    </Inside>
  )
}
```
```jsx
import { useInside } from '@aragon/ui'

function MyChildComponent() {
  const isInsideMyParent = useInside('MyParentComponent')
  return (
    <p>{isInsideMyParent? 'Somewhere in MyParentComponent' : 'Not'}</p>
  )
}
```


src/utils/inside.js
```diff
@@ -0,0 +1,35 @@
+// This pair of component + hook can be used to know that a component is in the
+// tree of another component. It works by using a Map containing named contexts.
+
+import React, { useContext } from 'react'
+import PropTypes from 'prop-types'
+
+// List of contexts, created as they are requested
+const insideContexts = new Map()
```
AquiGorka: Why did you choose to use a `var` outside of the `useInside` context? What's the difference between using a `state` `var` and accompanying functionality there?

src/utils/inside.js
```diff
@@ -0,0 +1,35 @@
...
+
+// Creates the required context if it doesn’t exist.
+function getContext(name) {
+  if (!insideContexts.has(name)) {
+    insideContexts.set(name, React.createContext(false))
```
sohkai: A small concern is what happens when you nest the same (or similar) components.

Imagine both

```js
<Bar.Primary>
  <Bar.Primary>
    <YourComponent />
  </Bar.Primary>
</Bar.Primary>
```

And 

```js
<Bar.Primary>
  <Bar.Secondary>
    <YourComponent />
  </Bar.Secondary>
</Bar.Primary>
```

In the first case, we would be updating the context’s value with the nested `<Bar.Primary>`, and this should be fine.

In the second case, both of those contexts would return true, which makes it difficult for the programmer in `YourComponent` to figure out exactly what scenario it’s in.

I don’t think this is a big problem at the moment, but can lead to surprises.

sohkai: Perhaps we may want to warn if we’ve detected you’re nested the same `Inside` namespace?

src/utils/inside.js
bpierre: To declare the contexts, and being able to access them from the provider and consumer.

The same way we use `React.createContext()` outside of components :-)

src/utils/inside.js
bpierre: mmm I’m not sure I understand :thinking: 

For example, we have [`Bar`](https://github.com/aragon/aragon-ui/blob/newstyle/src/components/Bar/Bar.js#L49) and [`Bar:primary`](https://github.com/aragon/aragon-ui/blob/newstyle/src/components/Bar/Bar.js#L33).

`Bar` tells you that you are inside the Bar component (you could switch to an alternative style for your component), while `Bar:primary` tells you that you are also specifically in this slot (you could, for example, cancel the left padding in that case).

> In the second case, both of those contexts would return true, which makes it difficult for the programmer in `YourComponent` to figure out exactly what scenario it’s in.

If you want to handle this specific case, you could do `insideBarPrimary && insideBarSecondary`?

bpierre: > Perhaps we may want to warn if we’ve detected you’re nested the same Inside namespace?

You mean to prevent people to nest a component that shouldn’t be nested? I’m not sure `Inside` would be the best place to handle this, as we might want to have components that can be nested, in aragonUI or apps. Perhaps we could do it at the components level, rather than in `Inside`?

Example:

```jsx
function MyButton() {
  if (useInside('MyButton')) {
    throw new Error('You can’t nest MyButton in MyButton')
  }
  return (
    <Inside name="MyButton">
      <button {...props} />
    </Inside>
  )
}
```

sohkai: Ah yes, I think doing that in each component makes sense 👍.

> If you want to handle this specific case, you could do `insideBarPrimary` && `insideBarSecondary`?

What if a component wanted to do different things if it was in the primary slot or the secondary slot? In that second construction, it would get confused (and it would have to basically say “if it’s in both, somehow, it prefers one, **but** not the most direct parent”.

bpierre: Oh I think I see, are you thinking of components like `Modal`?

Example with pseudo implementations:

```jsx
function Card({ children }) {
  return (
    <Inside name="Card">
      {children}
    </Inside>
  )
}

function Button(props) {
  const insideBox = useInside('Card')

  if (insideBox) {
    // specific style for a Button inside a Card
    props.style.width = '100%'
  }

  return <button {...props} />
}

function Modal({ children }) {
  return (
    <Inside name="Modal">
      <Root>
        {children}
      </Root>
    </Inside>
  )
}

function MyApp() {
  return (
    <Card>
      <Button>This button will have styles adapted to Card</Button>
      <Modal visible>
        <div>Modal content</div>
        <Button>This Button too, which is unwanted ☹</Button>
      </Modal>
    </Card>
  )
}
```

We could have different ways to solve this:

- In Button, test if we are inside of `Card`, but not inside `Root`. That would be my preferred option.
- We could exclude components using `Root` from inheriting `Inside`.
- Components like Button could specify the list of domains that can stop the `Inside` inheritance. This one is probably not great as it would require to maintain these lists when we add new components like `Modal`.

Is it what we were referring to? I might be a bit confused with the example, because the `Bar:Primary` and `Bar:Secondary` names are created by the `Bar` component, and can not be nested :-)

sohkai: Yes, that would be a similar example, although my concern was a bit more generic (components down-tree can never know which "inside" is their most direct parent, and thereby, if they're listening for more than one, it becomes impossible to react correctly 100% of the time).

In any case, perhaps this is more of an academic concern at this point and it's unlikely that a component will clash in that most generic sense. Each specific case, like the Card and Modal, can have workarounds as you've mentioned.

bpierre: Cool, yes let’s come back to this conversation after having used it a bit :+1:

PR #424, opened 2019-07-30
Add warnOnce()

bpierre:
Makes it easier to warn about deprecated things :hugs:

### DIFF
--- a/src/utils/environment.js
+++ b/src/utils/environment.js
@@ -14,3 +14,12 @@ export const log = devOnly((...params) => {
 export const warn = devOnly((...params) => {
   console.warn(...params)
 })
+
+// Like warn(), but only once
+const Warned = new Map()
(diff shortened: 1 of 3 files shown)


src/utils/environment.js
```diff
@@ -14,3 +14,12 @@ export const log = devOnly((...params) => {
 export const warn = devOnly((...params) => {
   console.warn(...params)
 })
+
+// Like warn(), but only once
+const Warned = new Map()
+export const warnOnce = devOnly((domain, ...params) => {
```
AquiGorka: Curious, why didn't you implement a `hook`?

bpierre: To make it work across component instances, e.g. if `<Button deprecatedProp />` gets used several instances in a tree, or unmounted / remounted several times.

AquiGorka: Ahhhhh, makes sense, thank you.

PR #425, opened 2019-07-31
Button: positive / negative + iconOnly

bpierre:


src/components/Button/Button.js
```diff
@@ -1,47 +1,102 @@
...
-          ? theme.accentContent
-          : theme.surfaceContent};
-        min-width: ${16 * GU}px;
-        height: ${(size === 'small' ? 4 : 5) * GU}px;
+        background: ${background};
+        color: ${color};
+        min-width: ${minWidth};
+        width: ${width};
+        height: ${height};
         padding: 0 ${3 * GU}px;
```
AquiGorka: Does this work for small layout buttons? I think we should remove the padding on smaller layouts no?

src/components/Button/Button.js
```diff
@@ -57,20 +112,20 @@ function Button({ children, icon, label, mode, size, innerRef, ...props }) {
...
                 display: flex;
-                color: ${theme.surfaceIcon};
+                color: ${iconColor};
               `}
             >
               {icon}
             </span>
           )}
-          {icon && label && (
+          {icon && label && !iconOnly && (
```
AquiGorka: Why is `icon` needed in this check?

src/components/Button/Button.js
```diff
@@ -1,47 +1,102 @@
...
+  const height = `${(size === 'small' ? 4 : 5) * GU}px`
+  const width = iconOnly ? height : 'auto'
+  const minWidth = `${iconOnly ? 0 : 16 * GU}px`
+
+  if (iconOnly) {
+    props.title = label
+  }
+
   return (
     <ButtonBase
```
AquiGorka: What about passing down other props to `ButtonBase` (eg `disabled`)

src/components/Button/Button.js
bpierre: Yes let’s reduce it a bit on `small` size :+1:

Let’s see how `small` and `large` evolve, as we are only using the normal size for now in the new style.

src/components/Button/Button.js
bpierre: Yes totally! I’ll even throw an error in that case.

src/components/Button/Button.js
bpierre: Done already :-) https://github.com/aragon/aragon-ui/pull/425/files/fff387cef8568468e60142af6173b98d295a671f#diff-d1110304895e39c6e9fd2be689023291R105

src/components/Button/Button.js
bpierre: After checking again: we need this specific check in case someone uses `icon` without `iconOnly`, not the other way.

I added a warning though, in case `iconOnly` gets used without having set `icon`.

PR #428, opened 2019-07-31
Empty state card update gallery

AquiGorka:


gallery/src/pages/PageEmptyStateCard.js
```diff
@@ -11,7 +11,7 @@ const PageEmptyStateCard = ({ title }) => (
       <Container>
         <EmptyStateCard
           text="You seem to not have any content on your wall."
-          icon={
+          illustration={
```
bpierre: We should remove this prop in the gallery demo, and finally get a super nice illustration instead of the blue house :+1: :laughing:

AquiGorka: Done

PR #431, opened 2019-08-01
Info modes

bpierre:
Info can now take a `mode` prop, with one of the following: `info` (default), `warning`, `error`.

Note: I went for a prop rather than separated `Warning` + `Error` components because `Error` is already used by JavaScript.


src/components/Info/Info.js
```diff
@@ -1,43 +1,92 @@
...
+        padding: ${2 * GU}px;
+        border-radius: ${RADIUS}px;
+        word-wrap: break-word;
+        ${textStyle('body3')};
+      `}
       {...props}
     >
-      {title && <Title>{title}</Title>}
+      {title && (
+        <h1
```
AquiGorka: I'd suggest to not use an `h1` here, there should only be one `h1` per page, maybe `h2` to be on the safe side? We don't really know the context where this is going to be used.

src/components/Info/Info.js
```diff
@@ -1,43 +1,92 @@
...
-    </Main>
+    </section>
   )
 }
 
 Info.propTypes = {
-  background: PropTypes.string,
   children: PropTypes.node,
   title: PropTypes.node,
+  mode: PropTypes.oneOf(['info', 'warning', 'error']),
```
AquiGorka: Would it make sense to change `warning` to `warn`?

src/components/Info/Info.js
bpierre: It’s in a section :-)

src/components/Info/Info.js
bpierre: It’s an object, so I would rather use object names rather than verbs (e.g. we have `info` not `inform`).

src/components/Info/index.js
```diff
@@ -1,8 +1,9 @@
...
-Info.Action = Action
-Info.Alert = Alert
-Info.Permissions = Permissions
+// Backward compatibility
+Info.Action = Info
+Info.Permissions = Info
+Info.Alert = props => <Info mode="warning" {...props} />
 
-export default Info
+export { Info }
```
sohkai: I may have missed the memo, but are we preferring non-default exports now?

src/components/Info/index.js
```diff
@@ -1,8 +1,9 @@
...
-import { Action, Alert, Permissions } from './IconInfo'
+import React from 'react'
+import { Info } from './Info'
 
-Info.Action = Action
-Info.Alert = Alert
-Info.Permissions = Permissions
+// Backward compatibility
+Info.Action = Info
+Info.Permissions = Info
```
sohkai: From what I can tell in the [designs](https://www.figma.com/file/M9AbLkUkeRCFKQt6qdFUsCks/Tokens?node-id=200%3A1939), I think `Info.Permissions` is also the warning mode.

src/components/Info/index.js
bpierre: Not a strict decision, I’ve just been experimenting with it recently.

Some advantages I found:

- Not having to [transform default imports](https://github.com/aragon/aragon-ui/blob/newstyle/src/components/index.js) into named imports.
- We can export everything [at once](https://github.com/aragon/aragon-ui/blob/newstyle/src/components/Layout/Layout.js#L71) (instead of using [two exports](https://github.com/aragon/aragon-ui/blob/newstyle/src/providers/PublicUrl/PublicUrl.js#L41-L42)).
- To avoid the mixed import syntax: `import A, { useB } from './A'`
- Works great with files mixing components and hooks.

@sohkai @AquiGorka What do you think?

src/components/Info/index.js
bpierre: Oh yes!

src/components/Info/index.js
sohkai: > To avoid the mixed import syntax: import A, { useB } from './A'

I actually prefer this, as it tells me what the priority is. The default export is the "main" use, and anything else I consider utilities that are optional.

But since most users of aragonUI will be importing what they need through `@aragon/ui` directly through named exports, this is an internal thing and in that sense no external users should be impacted by this decision.

bpierre: > The default export is the "main" use, and anything else I consider utilities that are optional.

Yeah, ok let’s move back to have a main + named, at least for component modules, I don’t really have a strong opinion on this.

> But since most users of aragonUI will be importing what they need through @aragon/ui directly through named exports, this is an internal thing and in that sense no external users should be impacted by this decision.

Yes exactly, that’s why I was freely trying this style to see how it feels.

AquiGorka: I don't have a strong opinion here, the benefits Pierre outlined work for me and this is an internal thing so 👍

PR #432, opened 2019-08-02
Button: disabled state + display prop

bpierre:
image

The `display` prop can be used to force the button to behave in certain ways. It replaces `iconOnly`, which is now deprecated.

`display` values:

- `all`: both the icon and the label are displayed.
- `icon`: only the icon gets displayed.
- `label`: only the label gets displayed.
- `auto` (default): like `all` except in some cases (one for now, see below).

If the Button is inside the secondary slot of Header and its display is set to auto, it will alternate between displaying only the label and only the icon, depending on the current layout.

### DIFF
--- a/devbox/apps/Button.js
+++ b/devbox/apps/Button.js
@@ -8,6 +8,7 @@ import {
   Button,
   Text,
   theme,
+  GU,
 } from '@aragon/ui'
 
 const MODES = ['normal', 'strong', 'positive', 'negative']
@@ -38,12 +39,16 @@ function App() {
         bottom: 0;
       `}
     >
-      <Box css="padding: 24px">
+      <Box
+        css={`
+          padding: ${3 * GU}px;
+        `}
+      >
         <div
           css={`
             display: grid;
             grid-template-columns: repeat(4, auto);
-            grid-gap: 24px;
+            grid-gap: ${3 * GU}px;
           `}
         >
           {SIZES.map(size =>
--- a/devbox/apps/Button.js
+++ b/devbox/apps/Button.js
@@ -53,9 +53,9 @@ function App() {
                   key={size + mode + i}
                   mode={mode}
                   size={size}
-                  label={mode}
+                  label={mode[0].toUpperCase() + mode.slice(1)}
                   icon={getIcon(mode)}
-                  display={i === 1 ? 'icon' : 'label'}
+                  display={i === 1 ? 'icon' : i === 0 ? 'all' : 'label'}
                   disabled={i === 3}
                 />
               ))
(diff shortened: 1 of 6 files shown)


src/components/Header/Header.js
```diff
@@ -3,51 +3,58 @@ import PropTypes from 'prop-types'
 import { GU, textStyle } from '../../style'
```
AquiGorka: How did this get into this PR? 😄

devbox/apps/Button.js
```diff
@@ -28,60 +36,33 @@ function App() {
...
-            <Text size="xxlarge">Button variations</Text>
-          </h1>
-        )}
-        <div css="padding: 40px 0">
+      <Box css="padding: 24px">
+        <div
+          css={`
+            display: grid;
+            grid-template-columns: repeat(4, auto);
+            grid-gap: 24px;
```
AquiGorka: GUs?

src/components/Button/Button.js
```diff
@@ -107,25 +148,25 @@ function Button({
...
-        width: ${iconOnly ? height : width};
+        width: ${displayIconOnly ? height : width};
         height: ${height};
-        padding: ${iconOnly ? 0 : padding};
-        min-width: ${iconOnly ? 0 : 16 * GU}px;
+        padding: ${displayIconOnly ? 0 : padding};
+        min-width: ${displayIconOnly ? 0 : minWidth}px;
         border: ${border};
-        box-shadow: 0 1px 3px rgba(0, 0, 0, 0.1);
+        box-shadow: ${disabled ? 'none' : '0 1px 3px rgba(0, 0, 0, 0.1)'};
```
AquiGorka: Should we have a boxShadow prop in colors/theme?

src/components/Header/Header.js
bpierre: haha yes it might look like I changed the entire file! It only adds `Inside` to `Header`, so that `Button` can act accordingly.

I recommend the “hide whitespace changes” setting :laughing:

devbox/apps/Button.js
bpierre: Yes!

src/components/Button/Button.js
bpierre: Yes totally, we could start defining some standard shadow values and export them.

We might not want to have them in the theme right now, but it might become necessary depending on how the dark theme evolves (e.g. a surface with a shadow in the light theme might become a flat surface with a border in the dark theme).

For the time being, they could live in `style/` with the other static styles (`RADIUS`, `GU`, text styles, etc.). I’ll check with @dizzypaty and @owisixseven and add this soon :+1:

PR #434, opened 2019-08-02
Tag component

AquiGorka:
image


bpierre:
:raised_hands: :raised_hands: :raised_hands: 

I was imagining we would remove / deprecate `Badge` and its sub components at the same time we add `Tag` :thinking: What do you think?

AquiGorka:
@bpierre I fear this commit https://github.com/aragon/aragon-ui/pull/434/commits/a526fe12ffba08e1c7a86c9d765c42a1add07cb5 is as dangerous as deploying to production on a Friday.

bpierre:
@AquiGorka haha yes it would probably break a few apps in the current state :s

I think we should ensure the previous APIs don’t break by redirecting some components, while recommending something else:

- `Badge` => render & recommend `Tag` (already done).
- `Badge.Info` => render & recommend `Tag` (later we’ll recommend `Help`).
- `Badge.Identity` => render `Tag`, recommend `IdentityBadge`.
- `Badge.App` => render & recommend `Tag` (later we’ll recommend `AppBadge`).
- `Badge.Notification` => render & recommend `Count`.
- `BadgeNumber` => render & recommend `Count`.

Could we also add a quick note to `UPGRADE.md` about this? Just to tell people that Badge is now Tag in most cases.

src/components/Tag/Tag.js
```diff
@@ -0,0 +1,88 @@
...
+import PropTypes from 'prop-types'
+import { useTheme } from '../../theme'
+import { GU } from '../../style'
+
+const INFO = 'info'
+const WARNING = 'warning'
+const HELP = 'help'
+const TAG = 'tag'
+const APP = 'app'
+const IDENTITY = 'identity'
```
sohkai: Are we going to be supporting the app and identity modes? Can we just remove them (and from the examples)?

src/components/Tag/Tag.js
```diff
@@ -0,0 +1,88 @@
...
+function Tag({ children, mode = 'info', ...props }) {
+  const modeStyles = useMode(mode)
+
+  return (
+    <span
+      css={`
+        display: inline-flex;
+        align-items: center;
+        justify-content: center;
+        border-radius: 100px;
```
sohkai: It would be nice to give a reason for this value (not sure what other `border-radius` we now use, but this seems quite high).

AquiGorka: Yeah! @dizzypaty and @owisixseven explain yourselves. I followed the designs and this was the value used, for one line items it works perfectly, maybe a smaller value could be used.

src/components/Tag/Tag.js
AquiGorka: Before merging these should be updated.

src/components/Tag/Tag.js
```diff
@@ -0,0 +1,88 @@
...
+  }
+
+  // info
+  return `
+    background: ${theme.infoSurface.alpha(0.08)};
+    color: ${theme.infoSurfaceContent};
+  `
+}
+
+function Tag({ children, mode = 'info', ...props }) {
```
bpierre: Should we add `color` and `background`, as an officially supported way to set custom colors? It would make it aligned with [`Info`](https://github.com/aragon/aragon-ui/pull/431/files#diff-cd8900545676308fc1c1deb01a4f579fR86).

UPGRADE.md
```diff
@@ -62,3 +62,14 @@ Support for the `onSelect` prop has been removed. Please use `onChange` instead.
...
+## Badge
+
+If you were using the `Badge` component and/or its variants (`Badge.Info`, `Badge.Identity`, `Badge.App`, `Badge.Notification`, `BadgeNumber`), you might want to switch as follows:
+
+- `Badge` => `Tag`.
+- `Badge.Info` => `Help`.
+- `Badge.Identity` => `IdentityBadge`.
+- `Badge.App` => `AppBadge`.
+- `Badge.Notification` => `Count`.
+- `BadgeNumber` => `Count`.
```
bpierre: Just a reminder that `Help`, `AppBadge` and `Count` remain to be implemented.

- `Help` should arrive soon, but we can recommend `Tag` for now.
- `AppBadge` will likely follow later, so we should probably recommend `Tag` instead.
- `Count` is a variation of `Tag` (with a “99+” behavior) and should be implemented before we merge this PR to maintain `BadgeNumber` compatibility.

src/components/Tag/Tag.js
sohkai: I also notice in https://github.com/aragon/aragon/pull/901/files#diff-f111ecb6e7e901ba1e0a599993305779R53 that we use `foreground`.

bpierre: Yes, the original intention was to make components independent from CSS concepts and prepare aragonUI APIs for non-web environments… but I recently thought that it might feel more intuitive to adopt the CSS definition of “color” (= “foreground color”), even if it’s not as descriptive. Also because web is our only target for now. Thoughts?

src/components/Tag/Tag.js
bpierre: I’ll let @dizzypaty and @owisixseven confirm, but I think the idea is to use a high value so that the sides always have half circles even if the size changes (e.g. if the badge height is `22px`, sides will be perfect half circles starting from `11px` as a border-radius, but it won’t break if a higher value is set). I don’t know if we can do this in Figma, but in CSS we can do `border-radius: 50%` to achieve the same thing.

> for one line items it works perfectly

We probably want it to add a `white-space: nowrap` (with `overflow: hidden` + `text-overflow`) to always have one line?

AquiGorka: > border-radius: 50%
Creates a sphere always, and with odd results when dimensions are not 1:1
image

What Pierre described about the big value is correct. We might be able to use a smaller value, if we find a need for it.

About `white-space` and rest for one liners is it a hard dependency? Containers can take care of word breaking no?

bpierre: >  Creates a sphere always, and with odd results when dimensions are not 1:1

Of course, it’s not a disc :man_facepalming: :laughing:, sorry about giving a wrong solution. We could do half the height in `px` then.

bpierre: > About white-space and rest for one liners is it a hard dependency? Containers can take care of word breaking no?

I think components should ensure to not break as much as reasonably possible, no matter where they are being used, so I would set this on the component itself.

src/components/Tag/Tag.js
sohkai: Ah, I didn't realize that `color` and `foreground` were meaning the same thing. `color` sounds good, I think some other components also expose `color` (rather than `foreground`).

src/components/BadgeNumber/BadgeNumber.js
```diff
@@ -1,31 +1,9 @@
...
-)
+export default props => {
+  warnOnce('"BadgeNumber" has been deprecated. Please use "Count" instead.')
 
-BadgeNumber.propTypes = {
-  number: PropTypes.number,
-  small: PropTypes.bool,
-  color: PropTypes.string,
-  background: PropTypes.string,
+  return <Tag {...props} />
```
bpierre: It could be nice to have `Count` here, and do some props mapping to maintain some level of compatibility with `BadgeNumber`.

src/components/Count/Count.js
```diff
@@ -0,0 +1,22 @@
...
+
+function Count({ value, digits, ...props }) {
+  const parsed = parseInt(value, 10)
+  const max = Math.pow(10, digits) - 1
+  const formatedValue = parsed >= max ? `${max}+` : parsed
+
+  return <Tag {...props}>{formatedValue}</Tag>
+}
+
+Count.propTypes = {
```
bpierre: We should probably extend `Tag.propTypes` here, since we are passing the props to it (so that users don’t get an error from `Tag` when they use `Count`).

src/components/Count/Count.js
```diff
@@ -0,0 +1,22 @@
...
+  value: PropTypes.number.isRequired,
+  digits: PropTypes.number,
+}
+
+Count.defaultProps = {
+  digits: 4,
```
bpierre: Isn’t `2` what we use in most cases? e.g. for the notifications.

AquiGorka: Sure, no strong preference here, I recalled having a conversation about using 4 on `Voting` so that's why I went with this default.

src/components/Tag/README.md
```diff
@@ -0,0 +1,49 @@
...
+
+- Type: `String`
+- Values: `info` (default), `warning`, `help`, `tag`, `app`, `identity` and `notification`
+
+Set of predefined color and background combinations.
+
+### `size`
+
+- Type: `String`
+- Values: `normal` (default) or `compact`
```
bpierre: To align with other component’s APIs, we should either use `"small"` rather than `"compact"`, or use `compact` but as a boolean.

src/components/Tag/README.md
```diff
@@ -0,0 +1,49 @@
...
+
+- Type: `String` (color)
+
+Optional. The background color.
+
+### `uppercase`
+
+- Type: `Boolean`
+
+Optional. Use in case it is needed to `unset` the predefined text styles that uppercase the contents.
```
bpierre: Should we have it on by default? It seems like the way we use it for now.

src/components/Tag/Tag.js
bpierre: Do we have cases where identity should be used? It seems like `IdentityBadge` should always be used instead? /cc @dizzypaty

src/components/Tag/Tag.js
```diff
@@ -0,0 +1,119 @@
...
+  }
+
+  if (mode === APP) {
+  }
+
+  if (mode === IDENTITY) {
+  }
+
+  if (mode === NOTIFICATION) {
+  }
```
bpierre: I think the mode notification could use `theme.accent` and `theme.accentContent`?

image

src/components/Tag/Tag.js
```diff
@@ -0,0 +1,119 @@
...
+  const sizeStyles = useSize(size)
+
+  return (
+    <span
+      css={`
+        display: inline-flex;
+        align-items: center;
+        justify-content: center;
+        border-radius: 100px;
+        white-space: nowrap;
```
bpierre: Should we add a fixed height, since it’s only using one line anyway? At the moment the height is 22 and 15, while it should be 24 and 16 (from the Figma file).

Also could we add, in normal mode and when `uppercase` is on, a padding-top of `1.5px` to adjust the font alignment? Something like this (except you might want to have these in `useSize()`):

```suggestion
        white-space: nowrap;
        height: ${size === NORMAL? 3 * GU : 2 * GU}px;
        padding-top: ${size === NORMAL && uppercase? '1.5px' : '0'};
```

src/components/Tag/README.md
AquiGorka: Need to update this, the default is on yes

src/components/Tag/Tag.js
```diff
@@ -0,0 +1,119 @@
...
+Tag.propTypes = {
+  children: PropTypes.node,
+  mode: PropTypes.oneOf([
+    INFO,
+    WARNING,
+    HELP,
+    TAG,
+    APP,
+    IDENTITY,
+    NOTIFICATION,
```
bpierre: I’m wondering if we need all these modes?

- Is `warning` used anywhere?
- Could `app` be replaced entirely by `tag`? IIRC we chose the name tag for the purple in the theme because it is supposed to be the color for the “neutral tag”.
- Is `help` used anywhere?
- I think `IdentityBadge` should always be preferred over `identity`.

And one that I can’t see is this one (or should it be `info`?):



Also the background color for `tag` should have 16% opacity (at least for now: I think we need to update this color in the theme to not have any opacity on it, then we’ll update `Tag` if needed).

src/components/Tag/Tag.js
```diff
@@ -84,7 +90,7 @@ function Tag({ children, mode, size, uppercase, color, background, ...props }) {
         white-space: nowrap;
         ${modeStyles};
         ${sizeStyles};
-        ${!uppercase ? 'text-transform: unset' : ''};
+        ${uppercase ? 'text-transform: unset' : ''};
```
sohkai: Wouldn't we want `'text-transform: uppercase'` if `uppercase` was true?

src/components/Count/Count.js
```diff
@@ -16,7 +26,19 @@ Count.propTypes = {
...
+    WARNING,
+    HELP,
+    TAG,
+    APP,
+    IDENTITY,
+    NOTIFICATION,
+  ]),
+  color: PropTypes.string,
+  background: PropTypes.string,
+  size: PropTypes.oneOf([NORMAL, SMALL]),
```
sohkai: Don't think we need the rest of these, I think only `digits` needs to be a default prop?

I think we can leave the other prop validation to Tag so we don't have to duplicate these.

src/components/Tag/Tag.js
AquiGorka: The `textStyles` already have that `text-transform: uppercase`, here the intention is to remove it by unsetting when `!uppercase`

src/components/Count/Count.js
AquiGorka: Hmmm, @bpierre suggested I added Tag prop types here.

bpierre: That’s what I meant :-) https://github.com/aragon/aragon-ui/pull/434/commits/50cd0b8b5f7bc00b34501d00bafa903151743290

They were also declared on `defaultProps` rather than `propTypes` so I moved them back.

src/components/Tag/Tag.js
sohkai: I've made a fix for this in c3bda11; it was not applying `!uppercase` correctly 😄

UPGRADE.md
```diff
@@ -67,9 +67,9 @@ If you were using the `Table` component, you might want to switch to
...
-- `Badge.Info` => `Tag`.
-- `Badge.Identity` => `IdentityBadge`.
-- `Badge.App` => `Tag`.
-- `Badge.Notification` => `Count`.
-- `BadgeNumber` => `Count`.
+- `Badge` => `<Tag />`.
+- `Badge.Info` => `<Tag />`.
+- `Badge.Identity` => `<IdentityBadge />`.
+- `Badge.App` => `<Tag />`.
+- `Badge.Notification` => `<Tag count />`.
```
AquiGorka: Wait what? I have already been using `Count`

bpierre: Yes sorry… after having a good overview of the modes, it was really weird to have a separate component for a given mode, only because it displays a number sometimes. Do you mind updating wherever you were using it? Modes names need to be updated as well.

bpierre: Yes sorry… after having a good overview of the modes, it was really weird to have a separate component for a given mode, only because it displays a number sometimes. Do you mind updating wherever you were using it? Modes names need to be updated as well.

src/components/Tag/README.md
```diff
@@ -0,0 +1,85 @@
...
+- Default value: `false`
+
+Use this to pass a number as the `label` prop of the Tag, and limit it to a certain number of digits (2 by default, see `countDigits`).
+
+### `countDigits`
+
+- Type: `Number`
+- Default value: `2`
+
+When `count` is `true`, set the maximum number of digits you want. For example, `250` will be displayed “99+” if the number of digits is `2`.
```
sohkai:
```suggestion
When `count` is `true`, set the maximum number of digits you want. For example, `250` will be displayed as “99+” if the number of digits is `2`.
```

src/components/Tag/README.md
```diff
@@ -1,50 +1,85 @@
...
 Defines the text style uppercase transformation for the contents.
+
+### `count`
+
+- Type: `Boolean`
+- Default value: `false`
+
+Use this to pass a number as the `label` prop of the Tag, and limit it to a certain number of digits (2 by default, see `countDigits`).
+
+### `countDigits`
```
sohkai: Could we collapse `count` and `countDigits` into a single prop?

It could accept boolean or a number; if it's truthy, it uses the default number setting (2).

src/components/Tag/README.md
```diff
@@ -1,50 +1,85 @@
...
+- Default: `"indicator"`
 
-Set of predefined color and background combinations.
+Set of predefined color, background and size combinations (which can be overridden by the corresponding props).
 
 ### `size`
 
-- Type: `String`
-- Values: `normal` (default) or `compact`
+- Type: `"normal"` or `"compact"`
```
sohkai: I think it's `'small'` rather than `'compact'`?

src/components/Tag/Tag.js
```diff
@@ -0,0 +1,190 @@
...
+              align-items: center;
+              margin-top: ${finalSize === SIZE_NORMAL ? '-3px' : '0'};
+            `}
+          >
+            {icon}
+          </span>
+          {icon && finalLabel && (
+            <span
+              css={`
+                width: ${0.25 * GU}px;
```
sohkai: Could we set this as a `margin-right` in the span containing the icon instead?

src/components/Tag/Tag.js
```diff
@@ -0,0 +1,190 @@
...
+        ${sizeStyles};
+        ${!uppercase && 'text-transform: unset'};
+        color: ${color || modeProps.color};
+        background: ${background || modeProps.background};
+      `}
+      {...props}
+    >
+      {children || (
+        <React.Fragment>
+          <span
```
sohkai: Not a big thing, but we might as well check that there is an icon before rendering this?

src/components/Tag/Tag.js
```diff
@@ -0,0 +1,190 @@
...
+  const finalSize = size || modeProps.size
+  const finalLabel = getLabel({
+    label: label || (!icon && children),
+    count,
+    countDigits,
+  })
+
+  const childrenOrLabel = children || finalLabel
+  const singleChar =
+    !icon && typeof childrenOrLabel === 'string' && childrenOrLabel.length < 2
```
sohkai: From what I can tell, we don't render `children` if there's an icon. Perhaps we can just use `finalLabel` here and do away with `childrenOrLabel`?

Also, should it be in disc mode if there's no label (`''`)?

src/components/Tag/README.md
bpierre: I did exactly this at some point haha.

But it felt weird to read this:

```jsx
<Tag count={2} />
```

But yeah maybe it’s better than having to do:

```jsx
<Tag count countDigits={2} />
```

We can always explain it in the documentation.

src/components/Tag/Tag.js
bpierre: Sure, do you see an issue with a spacing element or is it a personal preference?

src/components/Tag/Tag.js
bpierre: Oh yes!

src/components/Tag/README.md
bpierre: Done in https://github.com/aragon/aragon-ui/pull/434/commits/04c91192dc7550af3048feb4457dab40ee445c4e

src/components/Tag/Tag.js
bpierre: Fixed with the new modes:

image

src/components/Tag/README.md
bpierre: https://github.com/aragon/aragon-ui/pull/434/commits/443cbbdca0e644428b4d59033529c50f27927814

src/components/Tag/Tag.js
bpierre: Done https://github.com/aragon/aragon-ui/pull/434/commits/8d733c6045993649ead8bafb51cc1bf37b9986d4

src/components/Tag/README.md
AquiGorka: For real? I just updated to `countDigits`... :(

bpierre: haha nooo sorry Gorka! It’s the last API change hopefully :pray: :pray: :pray:

src/components/Tag/Tag.js
bpierre: > From what I can tell, we don't render children if there's an icon. Perhaps we can just use finalLabel here and do away with childrenOrLabel?

mmm the thing is that `children` can be set to a string too, and it’s not included in `finalLabel`.

> Also, should it be in disc mode if there's no label ('')?

Yes!

AquiGorka:
I don't think `<Tag count... />` is setting the correct padding for one char and it does not look like a circle:

image

bpierre:
@AquiGorka Looks like a bug, on it!

src/components/Tag/README.md
sohkai: I guess I should have been more explicit; using `countDigits` as the prop feels more natural than just `count`.

If `countDigits` is missing, we assume it's not in counter mode. If it's enabled, use the default. If it's a number, apply the number. To me that feels very natural, and fairly obvious (`<Tag countDigits />` tells me exactly what it's going to do.

src/components/Tag/Tag.js
sohkai: Personal preference, but I generally like to avoid creating new elements for spacing / styling if a margin or padding would do (similar principle for avoid `<br>` when a `margin-bottom` works, although less of an impact here).

I think it also makes it a little bit easier to reason about, because it relies on `icon`, and we already have that conditional to render the icon itself.

bpierre: Cool yes I made the change in the meantime, I agree it’s more readable here (at least until we can do `<span css="width: 0.25gu" />`. But more generally, I think spacing elements can sometimes be easier to reason about than a margin on one of the items. And yes, using `<br>` to achieve spacing doesn’t seem right :laughing:

src/components/Tag/README.md
bpierre: mmmm maybe that’s me, but “count digits” sounds a bit weird as a boolean. I think it’s ok to be a bit wrong when we use a single word for the sake of brevity, but I would expect two words to be more precise, at least in this example.

What about changing `count` to use a verb, like `limit`?

```jsx
<Tag limitDigits />
<Tag limitDigits={2} />
```

Or we could reverse it:

```jsx
// “enable the digits limit”
<Tag digitsLimit />

// “digits limit: 2”
<Tag digitsLimit={2} />
```

What about `digits`?

```jsx
// “digits (mode)”
<Tag digits />

// “(number of) digits”
<Tag digits={2} />
```

bpierre:
@AquiGorka It should be fixed now.

src/components/Tag/Tag.js
bpierre: I simplified by forbidding `children` to be used with `label` or `icon` => https://github.com/aragon/aragon-ui/pull/434/commits/0342711d8550469a1abb854b247924f5b7d1da4f and https://github.com/aragon/aragon-ui/pull/434/commits/a1905054ca81749532405324b12cd20702d8457f

src/components/Tag/Tag.js
bpierre: https://github.com/aragon/aragon-ui/commit/a1905054ca81749532405324b12cd20702d8457f

src/components/Tag/README.md
sohkai: `limitDigits` sounds more clear, although you'll notice it has the same structure as `countDigits` (`<verb><noun>`).

`digits` sounds weird as a boolean.

bpierre: > limitDigits sounds more clear, although you'll notice it has the same structure as countDigits (<verb><noun>).

Yes :-) But it sounds more like what it actually does (“limit digits” vs. “count digits”). Or do you understand it as “counting the digits”?

sohkai: I agree, `limitDigits` is a much better description of what the prop actually does.

bpierre: Done! https://github.com/aragon/aragon-ui/pull/434/commits/6a3f409f49eab98d85fa9f9ddacb809af0d91f48

PR #435, opened 2019-08-03
Add Distribution

bpierre:
image

`Distribution` can be used to visualize the distribution of a given value. It is often used inside the `Box` component, but can be used elsewhere too.


bpierre:
>  Question: Does `Box` render a `section` tag? If not, it should no?

It does, when a heading is set: https://github.com/aragon/aragon-ui/blob/newstyle/src/components/Box/Box.js#L14

>  In that case, if Distribution is used inside('Box') should the section container change for something else?

I don’t think it would be needed. Having:

```
section
  h1 (Box heading)
  section
    h1 (Distribution heading)
```

Is equivalent of having:

```
section
  h1 (Box heading)
  h2 (Distribution heading)
```

When used elsewhere, or in a Box without heading, it will be relative to the closest sectioning content.

src/components/Distribution/Distribution.js
```diff
@@ -0,0 +1,164 @@
...
+
+Distribution.defaultProps = {
+  display: 'all',
+  heading: null,
+  itemTitle: ({ value, percentage, index }) => {
+    return `${
+      typeof value === 'string' ? value : `Item ${index + 1}`
+    }: ${percentage}%`
+  },
+  renderLegendItem: ({ value, percentage, index }) => {
```
sohkai: Maybe we should use `name` instead of `value`?

src/components/Distribution/Distribution.js
```diff
@@ -0,0 +1,164 @@
...
+          ))}
+        </ul>
+      )}
+    </section>
+  )
+}
+
+Distribution.propTypes = {
+  colors: PropTypes.array,
+  display: PropTypes.oneOf(['all', 'bar']),
```
sohkai: Do you think we'll add more display modes later? Otherwise a `showLegend` prop may be more appropriate.

src/components/Distribution/Distribution.js
bpierre: Yes I don’t like it much either… but `value` can be anything, not only a string. For example, we could imagine that every item is an object with a name and an icon.

But I’m thinking now that we could move from:

```jsx
<Distribution
  values={[
    { value: 'Bananas', percentage: 37 },
    { value: 'Apples', percentage: 22 },
    { value: 'Cherries', percentage: 15 },
    { value: 'Oranges', percentage: 12 },
    { value: 'Grapefruits', percentage: 12 },
    { value: 'Rest', percentage: 2 },
  ]}
/>
```

To:

```jsx
<Distribution
  items={[
    { item: 'Bananas', percentage: 37 },
    { item: 'Apples', percentage: 22 },
    { item: 'Cherries', percentage: 15 },
    { item: 'Oranges', percentage: 12 },
    { item: 'Grapefruits', percentage: 12 },
    { item: 'Rest', percentage: 2 },
  ]}
/>
```

And then pass it to `renderLegendItem`:

```jsx
renderLegendItem: ({ item, percentage, index }) => {
```

It would also make `value` available for a future version: I decided to have percentages rather than number values for now, but I would like to calculate them like [`stakesPercentages()`](https://github.com/aragon/aragon-apps/blob/master/apps/token-manager/app/src/utils.js#L100-L204) does, by passing a list of values and a maximum number of items we want to render (grouping the extra items in “rest” automatically). It would require to be adapted a bit so that BN.js values are supported without bundling the library into aragonUI ([to save 10k](https://bundlephobia.com/result?p=bn.js@5.0.0)). I think the API of BN.js is stable enough to take that risk?

Another option could be to convert BN.js objects into BigInt [with a babel transform](https://github.com/Yaffle/babel-plugin-transform-bigint) for non-supported browsers (which would use [jsbi − 8k](https://bundlephobia.com/result?p=jsbi@3.0.0)), so that we are ready [for the day Safari starts supporting BigInt](https://caniuse.com/#feat=bigint).

src/components/Distribution/Distribution.js
bpierre: I was kind of anticipating other modes yes, but let’s use `showLegend` for now, we can always deprecate as needed.

src/components/Distribution/Distribution.js
sohkai: I like `item` more than `value`, but mainly `value` is confusing because we only use it for a label.

> It would also make value available for a future version: I decided to have percentages rather than number values for now, but I would like to calculate them

I like the simplicity of the current approach, and I feel like this concern could be lifted outside of this component. The current ability to render more than 100% total is actually a feature in my eyes.

bpierre: > I like the simplicity of the current approach, and I feel like this concern could be lifted outside of this component.

Maybe yes… I just feel that passing a list of values would be the easiest, rather than a having to calculate a list of percentages which can get pretty complex, like in `stakesPercentages()`. We could go for a default way of doing this calculation, and in cases where users want something diffferent, they’ll always be able to do it outside and pass their `percentage` manually. In other words, the component would still support the current mode, but also have a “simple mode” that doesn’t require to calculate percentages.

> The current ability to render more than 100% total is actually a feature in my eyes.

I started by warning when the total doesn’t equal 100%, but then I remembered cases like 33% - 33% - 33%, so the component now only warns [if the total is greater than 100%](https://github.com/aragon/aragon-ui/pull/435/files#diff-2bfb838f9b8108a568e45ec4e26679a1R29). Do you think we should remove it as well? Do you have a case in mind where having a total greater than 100% would be needed?

PR #436, opened 2019-08-06
DropDown: add placeholder, renderLabel and wide props

deamme:
## Changes
- `label` is going to be deprecated and is now renamed to `placeholder`.
- `renderLabel` prop has been added to force display. An example use case is to force display a shorten address instead of the whole address.
- `wide` prop has been added again (was missing since the `newstyle` implementation).
- Update docs.

### DIFF
--- a/devbox/apps/DropDown.js
+++ b/devbox/apps/DropDown.js
@@ -16,7 +16,7 @@ class App extends React.Component {
         <Container>
           <DropDown
             items={items}
-            label="Which fruit?"
+            placeholder="Which fruit?"
             header="Fruits"
             selected={active}
             onChange={this.handleChange}
--- a/devbox/apps/DropDown.js
+++ b/devbox/apps/DropDown.js
@@ -20,7 +20,7 @@ class App extends React.Component {
             header="Fruits"
             selected={active}
             onChange={this.handleChange}
-            width="400px"
+            wide
           />
         </Container>
       </Main>
--- a/src/components/DropDown/DropDown.js
+++ b/src/components/DropDown/DropDown.js
@@ -8,9 +8,9 @@ import { GU, RADIUS, textStyle } from '../../style'
 import { useTheme } from '../../theme'
 import { warnOnce, unselectable } from '../../utils'
 
-function useDropDown({ items, selected, onChange, label }) {
+function useDropDown({ items, selected, onChange, label, placeholder }) {
   const buttonRef = useRef()
-  const [selectedLabel, setSelectedLabel] = useState(label)
+  const [selectedLabel, setSelectedLabel] = useState(placeholder || label)
   const [opened, setOpened] = useState(false)
   const handleClose = useCallback(() => {
     // if the popover is opened and the user clicks on the button
@@ -60,6 +60,8 @@ const DropDown = React.memo(function DropDown({
   header,
   items,
   label,
+  placeholder,
+  renderLabel,
   onChange,
   selected,
   width,
@@ -94,6 +96,7 @@ const DropDown = React.memo(function DropDown({
     selected,
     items,
     label,
+    placeholder,
     onChange,
   })
   const closedWithChanges = !opened && selectedIndex !== -1
@@ -119,7 +122,7 @@ const DropDown = React.memo(function DropDown({
             ${textStyle('body2')};
           `}
         >
-          {selectedLabel}
+          {renderLabel || selectedLabel}
         </span>
         <IconDown
           size="tiny"
@@ -185,17 +188,19 @@ const DropDown = React.memo(function DropDown({
 DropDown.propTypes = {
   header: PropTypes.node,
   items: PropTypes.arrayOf(PropTypes.node).isRequired,
-  label: PropTypes.string,
+  placeholder: PropTypes.string,
+  renderLabel: PropTypes.string,
   onChange: PropTypes.func.isRequired,
   selected: PropTypes.number,
   width: PropTypes.string,
 
   // deprecated
   active: PropTypes.number,
+  label: PropTypes.string,
 }
 
 DropDown.defaultProps = {
-  label: 'Select an item',
+  placeholder: 'Select an item',
 }
 
 const Item = React.memo(function Item({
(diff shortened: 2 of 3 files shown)


src/components/DropDown/DropDown.js
```diff
@@ -8,9 +8,9 @@ import { GU, RADIUS, textStyle } from '../../style'
 import { useTheme } from '../../theme'
 import { warnOnce, unselectable } from '../../utils'
 
-function useDropDown({ items, selected, onChange, label }) {
+function useDropDown({ items, selected, onChange, label, placeholder }) {
   const buttonRef = useRef()
-  const [selectedLabel, setSelectedLabel] = useState(label)
+  const [selectedLabel, setSelectedLabel] = useState(placeholder || label)
```
bpierre: Could we also use `warnOnce()` (from `utils/environment.js`) when `label` is set?

```jsx
if (label !== undefined) {
  warnOnce('DropDown:label', 'DropDown: the “label” prop is deprecated, please use “placeholder” instead.')
}
```

src/components/DropDown/DropDown.js
```diff
@@ -185,17 +188,19 @@ const DropDown = React.memo(function DropDown({
 DropDown.propTypes = {
   header: PropTypes.node,
   items: PropTypes.arrayOf(PropTypes.node).isRequired,
-  label: PropTypes.string,
+  placeholder: PropTypes.string,
```
bpierre: Should `placeholder` be any node? (so we can have style variants like mini badges)

```suggestion
  placeholder: PropTypes.node,
```

src/components/DropDown/DropDown.js
```diff
@@ -185,17 +188,19 @@ const DropDown = React.memo(function DropDown({
...
+  renderLabel: PropTypes.string,
```
bpierre: We usually use the `render` prefix for render functions, maybe we could call it and pass the selected index?

```suggestion
  renderLabel: PropTypes.func,
```

src/components/DropDown/DropDown.js
```diff
@@ -119,7 +122,7 @@ const DropDown = React.memo(function DropDown({
             ${textStyle('body2')};
           `}
         >
-          {selectedLabel}
+          {renderLabel || selectedLabel}
```
bpierre: If we move `renderLabel` to be a function, then we could do, earlier:

`const Label = renderLabel`

And here:

```suggestion
          {Label? <Label selectedIndex={selectedIndex} /> : selectedLabel}
```

That way, users will be passing a function that might be a React component, or not.

src/components/DropDown/DropDown.js
AquiGorka: And as a default prop value: `index => items[index]`

bpierre: Yes good idea, having a default value can also serve as an example when looking at the props in the code. :+1:

src/components/DropDown/DropDown.js
```diff
@@ -185,17 +195,19 @@ const DropDown = React.memo(function DropDown({
...
 
   // deprecated
   active: PropTypes.number,
+  label: PropTypes.string,
 }
 
 DropDown.defaultProps = {
-  label: 'Select an item',
+  placeholder: 'Select an item',
 }
```
AquiGorka: We're missing the default prop for `renderLabel` no?

bpierre: @deamme @AquiGorka I added it 👉👉👉 https://github.com/aragon/aragon-ui/pull/436/commits/d518936db44fe205b15659e0f9f83e220811d308

deamme:
@bpierre Looks good to me!

src/components/DropDown/DropDown.js
```diff
@@ -91,36 +128,49 @@ const DropDown = React.memo(function DropDown({
...
+    // Update the button width every time the reference updates
+    setButtonWidth(el.clientWidth)
+  })
+
+  // And every time the viewport resizes
+  useWindowResize(() => {
+    if (buttonRef.current) {
+      setButtonWidth(buttonRef.current.clientWidth)
+    }
+  })
```
sohkai: Not sure I fully understand this; is this to keep the popover's width the same as the dropdown button's width?

src/components/DropDown/DropDown.js
```diff
@@ -55,13 +63,40 @@ function useDropDown({ items, selected, onChange, label, placeholder }) {
...
+    [cb]
+  )
+
+  return {
+    buttonRef,
+    refCallback,
+  }
+}
+
+function useWindowResize(onResize) {
```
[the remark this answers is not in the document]
bpierre: Ah yes, I was thinking of doing that but I felt bad using `useViewport` just for the event haha. But yes you’re right, `useViewport` is throttled already, so it makes sense to use it.

Actually we should even use `useLayout()`, which uses `useViewport()` internally, and will also give us resizes outside of the viewport, like when the menu panel opens / closes with internal apps.

src/components/DropDown/DropDown.js
bpierre: Yes, Popover is rendered at the root level, so we have to pass the width of the button to it.

devbox/apps/DropDown.js
```diff
@@ -20,7 +20,7 @@ class App extends React.Component {
             header="Fruits"
             selected={active}
             onChange={this.handleChange}
-            width="400px"
```
AquiGorka: If you remove `width` the `Popover` will wrap to its contents' width, maybe keep it to showcase how to modify it?

AquiGorka: Ahh, I see this is kept in sync with `js` via the `buttonRef`.

bpierre: Yes, using `width` shouldn’t be as needed now that we have `wide`, we usually want it to adapt to its content, or to be full width (side panel). :+1:

src/components/DropDown/DropDown.js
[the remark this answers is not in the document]
bpierre: Done: https://github.com/aragon/aragon-ui/pull/436/commits/56615f30b6c833b916143965ca1eb9e0e788b549

I ended up using `useViewport()` for now, as `useLayout()` only updates between breakpoints (and we probably want to keep it this way!).

PR #437, opened 2019-08-06
Fix CardLayout on medium and large layouts

bpierre:
Still looking for a solution on small layouts, where cards would have a minimum width and a maximum width, and the minimum width would be preferred by the layout algorithm (`minmax()` prefers the max value since `fr` can not be used as a first argument). See https://github.com/w3c/csswg-drafts/issues/2611

Other solutions:

- [`min()`](https://developer.mozilla.org/en-US/docs/Web/CSS/min) or [`clamp()`](https://developer.mozilla.org/en-US/docs/Web/CSS/clamp), which are not supported yet.
- Using Flexbox on the small layout (which would allow `min-width` / `max-width`). Since we are using fixed heights, this might be the easiest solution.
- JS calculations using `Viewport` and `Layout`, passing the calculated dimensions to the children `Card` using context.

### DIFF
--- a/devbox/apps/Card.js
+++ b/devbox/apps/Card.js
@@ -1,0 +1,48 @@
+import React, { useState } from 'react'
+import {
+  Header,
+  Layout,
+  Bar,
+  Box,
+  Button,
+  BackButton,
+  Card,
+  CardLayout,
+  Split,
+  Tabs,
+  Theme,
+  useLayout,
+  IconPlus,
+} from '@aragon/ui'
+import { ToggleThemeButton } from '../components/toggle-theme'
+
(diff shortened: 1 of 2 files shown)


AquiGorka:
Small layout is not respecting `Layout` padding:

image

What if we used `1fr` for small layouts? No `min-max`

bpierre:
@AquiGorka We can’t, using `repeat(auto-fit)` / `repeat(auto-fill)` with `1fr` (without `minmax`) is invalid :-(

bpierre:
Merging this for now, until we figure out a solution for the small layout.

PR #440, opened 2019-08-09
Distribution: add renderFullLegendItem()

bpierre:
This change makes it possible to deprecate `PartitionBar`, which is now using `Distribution` internally.


bpierre:
@sohkai Yes, I added some spacing between the left part and the percentage, which required some tweaks to keep `text-overflow: ellipsis` working.

PR #441, opened 2019-08-09
Dropdown fix

AquiGorka:
cc @deamme

### DIFF
--- a/src/components/DropDown/DropDown.js
+++ b/src/components/DropDown/DropDown.js
@@ -53,7 +53,7 @@ function useDropDown({
       return
     }
     setSelectedLabel(items[selected])
-  }, [items, selected, setSelectedLabel, label])
+  }, [items, selected, setSelectedLabel, label, placeholder])
 
   return {
     handleChange,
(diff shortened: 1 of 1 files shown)


src/components/DropDown/DropDown.js
```diff
@@ -47,8 +47,8 @@ function useDropDown({
 
   useEffect(() => {
     if (selected === -1 || !items[selected]) {
-      if (label) {
-        setSelectedLabel(label)
+      if (placeholder || label) {
```
bpierre: Shouldn’t `label` take priority over `placeholder`?

bpierre: `placeholder` should be added to the dependencies too :-)

AquiGorka: `label` is deprecated, I think it will eventually fade away...

AquiGorka: @deamme I think you did this on your PR no?

bpierre: Oh right! I was mistaking with `renderLabel`. All good then :-)

bpierre: No it’s not there (`placeholder` wasn’t in the `useEffect()` before) => https://github.com/aragon/aragon-ui/blob/newstyle-0.8/dropdown-fix/src/components/DropDown/DropDown.js#L56

AquiGorka: ahhhh you meant the effect deps, jajaja 😄 sure

PR #450, opened 2019-08-12
Add Link

bpierre:
Link combines the features of `ExternalLink` and `SafeLink` into a single component.


src/components/Link/Link.js
```diff
@@ -0,0 +1,33 @@
...
+const SAFE_REL = 'noopener noreferrer'
+
+function Link({ children, external, unsafe, ...props }) {
+  const passedProps = { ...props }
+
+  if (external) {
+    passedProps._target = 'blank'
+  }
+
+  if (unsafe) {
```
AquiGorka: Do we really need this?

bpierre: Good point yes! Users can set `rel` themselves to achieve that, especially because it’s not really recommended. :+1:

PR #451, opened 2019-08-12
Add ButtonText

bpierre:
image


src/components/Button/ButtonText.js
```diff
@@ -0,0 +1,34 @@
...
+import PropTypes from 'prop-types'
+import { ButtonBase } from './ButtonBase'
+import { GU } from '../../style'
+import { useTheme } from '../../theme'
+
+function ButtonText({ horizontalPadding, ...props }) {
+  const theme = useTheme()
+
+  const leftPadding =
+    Number(horizontalPadding === 'left' || horizontalPadding === 'normal') * GU
```
sohkai:
```suggestion
    Number(horizontalPadding === 'left' || horizontalPadding === 'both') * GU
```

src/components/Button/ButtonText.js
```diff
@@ -0,0 +1,34 @@
...
+
+  const rightPadding =
+    Number(horizontalPadding === 'right' || horizontalPadding === 'normal') * GU
```
sohkai:
```suggestion
    Number(horizontalPadding === 'right' || horizontalPadding === 'both') * GU
```

src/components/Button/Button.js
```diff
@@ -237,7 +237,7 @@ const ButtonWithRef = React.forwardRef((props, ref) => (
 ))
 
 ButtonWithRef.Anchor = React.forwardRef((props, ref) => (
-  <ButtonWithRef ref={ref} as={SafeLink} {...props} />
+  <ButtonWithRef ref={ref} as={Link} {...props} />
```
sohkai: Should we also provide a deprecation warning for `Button.Anchor` given that you can now just do `<Button href="">` to get the same behaviour?

src/components/Button/ButtonText.js
```diff
@@ -0,0 +1,34 @@
...
+    />
+  )
+}
+
+ButtonText.propTypes = {
+  ...ButtonBase.propTypes,
+  horizontalPadding: PropTypes.oneOf(['both', 'left', 'right', 'none']),
+}
+
+export default ButtonText
```
sohkai: Would it make sense to make this a mode in `Button`? Given that it's just applying a few styles and could leverage the same icon / etc. functionality?

src/components/Button/Button.js
bpierre: Yes!

src/components/Button/ButtonText.js
bpierre: That’s how it used to be before, but I thought it could be nice to separate them now, because of the amount of exclusive props they have:

- `ButtonText` only: `horizontalPadding`.
- `Button` only: `display`, `icon`, `size`, `wide`.

src/components/Button/Button.js
bpierre: Done https://github.com/aragon/aragon-ui/pull/451/commits/a240d3afce5e63903f06d3f922f561b8b44b667d

src/components/Button/ButtonText.js
sohkai: Sounds good, I'm not _really_ convinced we need `ButtonIcon` either but let's go with this for now and see if there are any pains in the future.

bpierre: > Sounds good, I'm not really convinced we need `ButtonIcon` either

Now that we have `display`, I agree it would make sense to have `ButtonIcon` in button yes (with a new `surface` mode). Which might mean that it would be reasonable to have `ButtonText` in there as well (with a `text` mode), even if it doesn’t fit perfectly :thinking:

PR #453, opened 2019-08-12
Add IconStarFilled

bpierre:
<p align=center>

</p>


bpierre:
> Our tooling doesn't really support it, but I imagine it would have been nice to add this as a "filled" mode to `IconStar` :man_shrugging: 

It’s something we could do, but I also like the idea of having a common API for all the `Icon` prefixed components, rather than having to explain that some are special.

PR #454, opened 2019-08-12
Inside: allow additional data to be passed

bpierre:
This is useful to pass some data from the parent component to the consumers.

Consumer:

```jsx
// like before
const insideSomething = useInside('Something')

// pass true as a second parameter to get the optional data
const [insideSomething, somethingData] = useInside('Something', true)
```

Provider:

```jsx
// data can take any value
function Something({ children }) {
  return (
    <Inside name="Something" data={{ hey: 'hey hey' }}>
      {children}
    </Inside>
  )
}
```

### DIFF
--- a/src/utils/inside.js
+++ b/src/utils/inside.js
@@ -33,9 +33,9 @@ Inside.propTypes = {
 
 // Use this hook to know if a given component is somewhere
 // in the tree of an <Inside> declared with the same name.
-function useInside(name, withData) {
+function useInside(name) {
   const { inside, data } = useContext(getContext(name))
-  return withData ? [inside, data] : inside
+  return [inside, data]
 }
 
 export { Inside, useInside }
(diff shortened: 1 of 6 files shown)


src/utils/inside.js
```diff
@@ -10,26 +10,32 @@ const insideContexts = new Map()
...
+  name: PropTypes.string.isRequired,
 }
 
 // Use this hook to know if a given component is somewhere
 // in the tree of an <Inside> declared with the same name.
-function useInside(name) {
-  return useContext(getContext(name))
+function useInside(name, withData) {
+  const { inside, data } = useContext(getContext(name))
+  return withData ? [inside, data] : inside
```
sohkai: I would not mind simplifying this so it always returns the array, with most uses of `useInside()` just destructuring the first index.

bpierre: It was an attempt to maintain compatibility, but it’s probably not much used outside of aragonUI yet. Let’s return the same thing every time :+1:

bpierre: https://github.com/aragon/aragon-ui/pull/454/commits/e2c4487f57275d1b4da88c84ceb71f9a575e8ac0

PR #455, opened 2019-08-12
Allow Bar in Box

bpierre:
Builds on https://github.com/aragon/aragon-ui/pull/454


bpierre:
> But I had the impression that both Bar and Box were meant as primary surfaces, and to me it doesn't make much sense to put one in the other (and especially because it wouldn't work the other way around).

Yes I agree it’s a bit weird… this is what it’s for:

image

Originally I wanted to pass `<BackButton />` this way:

```jsx
<Box
  heading={<BackButton />}
/>
```

But I wanted to avoid detecting its presence (using `contains-component`) only for that… but maybe it’s fine.

What about having a new prop, that would be mutually exclusive with `heading`?

```jsx
<Box
  actions={<BackButton />}
/>
```

src/utils/inside.js
```diff
@@ -10,26 +10,32 @@ const insideContexts = new Map()
...
+  name: PropTypes.string.isRequired,
 }
 
 // Use this hook to know if a given component is somewhere
 // in the tree of an <Inside> declared with the same name.
-function useInside(name) {
-  return useContext(getContext(name))
+function useInside(name, withData) {
+  const { inside, data } = useContext(getContext(name))
+  return withData ? [inside, data] : inside
```
AquiGorka: Could it happen that _falsy_ `withData` wants to be passed down?

bpierre: `withData` was only to request the return value, but it has been removed :point_right:  https://github.com/aragon/aragon-ui/pull/454#discussion_r313308639

sohkai:
I don't know how common the pattern will be, but if it's not going to be that common (most apps would prefer to have separate `Bar`s and `Box`s?), we can also just apply the negative margin directly to the `BackButton`.

I just don't see this being common enough for us to design a case around it right now.

bpierre:
> I just don't see this being common enough for us to design a case around it right now.

It was kind of the idea: make this possible, nothing more. Or do you mean by not having any support for it in aragonUI? That could be an option too, yes.

bpierre:
Closing as it can be done manually until it becomes a pattern.

PR #457, opened 2019-08-12
Field: use new theme / text styles

bpierre:


src/components/Field/Field.js
```diff
@@ -1,40 +1,50 @@
...
+          `}
+        >
           {label}
-          {isRequired && <StyledAsterisk title="Required">*</StyledAsterisk>}
-        </StyledTextBlock>
+          {isRequired && (
+            <span
+              css={`
+                color: ${theme.accent};
+                margin-left: auto;
```
AquiGorka: From what I've seen in the designs, the asterisks now are placed next to the label.
cc @dizzypaty

src/components/Field/Field.js
```diff
@@ -1,40 +1,50 @@
...
-const Field = ({ children, label, ...props }) => {
   const isRequired = React.Children.toArray(children).some(
     ({ props }) => props && props.required
   )
+
   return (
-    <StyledField {...props}>
+    <div
+      css={`
+        margin-bottom: ${3 * GU}px;
```
sohkai: Not sure on this bottom margin; the designs _feel_ like they may be smaller than before (we previously had 20px), so maybe 16px? cc @dizzypaty 

This change is also a bit scary since we don't always use `Field` directly and will have to update various self-set margins in the client and apps.

bpierre: From what I can see, the new margin is `24px`:



> This change is also a bit scary since we don't always use Field directly and will have to update various self-set margins in the client and apps.

I don’t think this is more of a problem here than in other places? We have been fixing the spacing as we go in various places, from using multiples of 10 to multiples of 8, and I think we should continue to do so. If a space look wrong, it probably means that we need to update it to GU units :smile:

src/components/Field/Field.js
bpierre: Done https://github.com/aragon/aragon-ui/pull/457/commits/592a1a8079437c58cd4ea32a709e944c446bd654

src/components/Field/Field.js
owisixseven: The margin here is 24px!

sohkai: > I don’t think this is more of a problem here than in other places?

Yes... it's just that we haven't used `Field` consistently and thus have a lot of places where we've used `20px` that we should update (and haven't yet).

bpierre: Yes I think we should do some grepping for `5px`, `10px`, `15px`, `20px` etc. in the different projects at some point, to make sure everything is up to date.

PR #458, opened 2019-08-12
Update CircleGraph

deamme:
## Intro
Some changes that I needed for the presale functionality of the Fundraising app.
You might or might not want these changes, but here they are 😃 

## Changes
- Font size is now dynamic based on the `size`
- ~Changed font color to `#25314d`~
- Border look has been changed to a grey outline
- Added `color` prop
- ~Added `width` prop to change the stroke width~ now `strokeWidth`

## Preview
CircleGraph

<hr />

## April 2020 changes

- Consolidated this version with the versions from [aragon/aragon](https://github.com/aragon/aragon/blob/c31c31f7df5b4ad32297e9ab8b466d24fc4cd59b/src/onboarding/CircleGraph.js) and [aragon/court-dashboard](https://github.com/aragon/court-dashboard/blob/57a4000e3a7cff114a47d95e8a3ec9f26aa6ed02/src/components/CircleGraph.js) (@rperez89).
- The label can now have a prefix, a suffix, and a secondary line.
- `color` is now optionally dynamic.
- A secondary line can be added.
- Updated documentation.


src/components/CircleGraph/README.md
```diff
@@ -50,3 +50,29 @@ A function to generate the label displayed on the graph.
...
+```jsx
+<CircleGraph color="#FF6969" />
+```
+
+### `width`
+
+* Type: `Number`
+* Default: `4`
+
+Width of the circle.
```
sohkai:
```suggestion
Stroke width of the circle.
```

I'd also consider renaming this prop to `stroke` or etc. "Width" of a circle usually means diameter which is definitely not what we want to convey 😄.

src/components/CircleGraph/CircleGraph.js
```diff
@@ -7,11 +7,12 @@ const BORDER_WIDTH = 4
 
 const VALUE_DEFAULT = 1
 const SIZE_DEFAULT = 80
-const LABEL_DEFAULT = value => `${Math.round(value * 100)}%`
+const LABEL_DEFAULT = value => `${Math.round(value * 100)}`
```
sohkai: I'd leave the % here; a user might want to display a non-percentage label (e.g. step 1 out of 3).

src/components/CircleGraph/CircleGraph.js
```diff
@@ -7,11 +7,12 @@ const BORDER_WIDTH = 4
...
+const COLOR_DEFAULT = '#21c1e7'
 
-const CircleGraph = ({ value, label, size }) => {
-  const length = Math.PI * 2 * (size - BORDER_WIDTH)
-  const radius = (size - BORDER_WIDTH) / 2
+const CircleGraph = ({ value, label, size, color, width }) => {
+  const length = Math.PI * 2 * (size - width)
+  const radius = (size - width) / 2
```
sohkai: Small nitpick; perhaps we can re-order this so it's more math-like:

```
const radius = (size - width) / 2
const circumference = Math.PI * 2 * (2 * r)
```

src/components/CircleGraph/CircleGraph.js
```diff
@@ -25,23 +26,40 @@ const CircleGraph = ({ value, label, size }) => {
...
             {progressValue.interpolate(t => label(Math.min(1, Math.max(0, t))))}
           </Label>
+          <Label
+            size={size / 6}
+            css={`
+              margin-top: ${size / 8}px;
+            `}
+          >
+            %
+          </Label>
```
sohkai: Rather than using two labels, perhaps we can use a `renderLabel` prop or etc. to let the user plug whatever label they'd like.

We can keep the simple previous one as a default.

src/components/CircleGraph/CircleGraph.js
```diff
@@ -72,21 +94,20 @@ const CircleSvg = styled.svg`
 
 const CircleBase = styled.circle`
   fill: none;
-  stroke: #6d777b;
-  opacity: 0.3;
+  stroke: #ecedf1;
```
sohkai: We probably want `theme.accent` for this instead?

src/components/CircleGraph/CircleGraph.js
```diff
@@ -72,21 +94,20 @@ const CircleSvg = styled.svg`
...
-  stroke: #21c1e7;
+  stroke: ${({ color }) => color};
 `
 
 const Label = styled(animated.div)`
-  font-size: 16px;
+  font-size: ${({ size }) => size}px;
   font-weight: 400;
-  color: #000;
+  color: #25314d;
```
sohkai: We should make this a theme colour; perhaps `theme.surfaceContent`?

src/components/CircleGraph/CircleGraph.js
deamme: This would make `%` as big as the text.

src/components/CircleGraph/CircleGraph.js
deamme: The stroke color in that case is the gray outline.

src/components/CircleGraph/CircleGraph.js
deamme: There's already a `theme.surfaceContent` which is gray.

src/components/CircleGraph/CircleGraph.js
deamme: Regarding `step 1 out of 3`, the way it works with the dynamic sizing that kind of text wouldn't work..

src/components/CircleGraph/README.md
```diff
@@ -7,46 +7,62 @@ A circular graph component.
...
 
-```jsx
-<CircleGraph size={40} />
-```
+| Type                 | Default value |
+| -------------------- | ------------- |
+| `Function` or `Node` | None          |
 
-### `value`
+When a React node is passed, it gets rendered directly.
```
sohkai, line 29:
```suggestion
If a React node is passed, it gets rendered directly.
```

src/components/CircleGraph/README.md
```diff
@@ -7,46 +7,62 @@ A circular graph component.
...
 
-* Type: `Number`
-* Default: `1`
+A function that gets called to generate the label displayed on the graph. It takes the current value as a parameter and should return an object containing the following nodes:
```
sohkai, line 31:
```suggestion
If a function is passed, it gets called to generate the label displayed on the graph. It takes the current value as a parameter and should return an object containing the following nodes:
```

src/components/CircleGraph/README.md
```diff
@@ -7,46 +7,62 @@ A circular graph component.
...
 
-The value represented by the graph, between 0 and 1.
+- `value`: the main number (defaults to `Math.round(value * 100)`).
+- `suffix`: the suffix displayed after (defaults to `%`).
```
sohkai, line 34:
```suggestion
- `suffix`: the suffix to display (defaults to `%`).
```

Suffixes by definition are display afterwards :)

src/components/CircleGraph/README.md
```diff
@@ -7,46 +7,62 @@ A circular graph component.
...
+- `prefix`: the optional prefix to display (defaults to `<` below 1%).
```
sohkai, line 35:
```suggestion
- `prefix`: the optional prefix to display (defaults to `<` when `value` is below 1).
```

src/components/CircleGraph/CircleGraph.js
```diff
@@ -1,92 +1,179 @@
...
+  }
+
+  // We assume label is a node if not a function.
+  const isLabelNode = typeof label !== 'function'
+
+  const { prefix, suffix, secondary } =
+    label && !isLabelNode ? label(value) : {}
+
+  let colorFn = color || (() => theme.accent)
+  if (typeof colorFn !== 'function') {
```
sohkai, line 38: Not sure I get this, wonder if we could do something like:

```js
const colorFn =
  typeof colorFn === 'function'
    ? colorFn
    : (() => typeof colorFn === 'string' : theme.accent)
```

devbox/apps/CircleGraph.js
```diff
@@ -0,0 +1,110 @@
...
+              size={lerp(size, SIZES[0], SIZES[1])}
+              color={value =>
+                `rgb(
+                  ${lerp(value, COLOR_START[0], COLOR_END[0])},
+                  ${lerp(value, COLOR_START[1], COLOR_END[1])},
+                  ${lerp(value, COLOR_START[2], COLOR_END[2])}
+                )`
+              }
+              secondary={
+                displaySecondary ? `${Math.floor(value / 10)}/10` : null
```
sohkai, line 56: Even though I was setting this, I don't think I saw a difference in the label (e.g. 50% was still 50%, rather than 5%).

devbox/apps/CircleGraph.js
```diff
@@ -0,0 +1,110 @@
...
+        >
+          <Card
+            css={`
+              width: 400px;
+              height: 400px;
+            `}
+          >
+            <CircleGraph
+              value={value / 100}
+              size={lerp(size, SIZES[0], SIZES[1])}
```
sohkai, line 47: The progress bar seems to take a life of its own when changing the size:

Screen Shot 2020-04-10 at 12 17 55 PM

bpierre: Yes this is a Chrome bug, it only happens when we dynamically change the size without re-rendering the SVG, so it shouldn’t be a problem outside of this demo.

src/components/CircleGraph/CircleGraph.js
bpierre: Ah yes, I adapted a bit but it’s nicer this way :+1: 

```jsx
  const colorFn =
    typeof color === 'function' ? color : () => color || theme.accent
```

We should avoid checking if `typeof` returns `string`, because we might receive `String` instances.

devbox/apps/CircleGraph.js
bpierre: I forgot to update the demo for this, it was using a previous version of the API. Updated: https://github.com/aragon/aragon-ui/pull/458/commits/7e411982feca6a44d35854fb29d4c06788fdc7a2

bpierre:
@sohkai I just made some changes to align the way all the label parts are treated: https://github.com/aragon/aragon-ui/pull/458/commits/8c337a1efce8e1a619920d5fbb0d48ac69c6e839

We might want to memoize `label()` since it gets called four times every frame, but I don’t think it should be a big issue.

sohkai:
> We might want to memoize label() since it gets called four times every frame, but I don’t think it should be a big issue.

We can also leave this as a user-problem for now, and see if it becomes an issue we have to consistently solve outside.

bpierre:
@deamme We finally got this one merged! :tada: :laughing:

deamme:
@bpierre Haha, awesome ! 😄 Hope you're doing great ❤️

PR #459, opened 2019-08-12
Input: new style

bpierre:
image

- Use the new theme colors.
- Use the new text styles.
- No shadow anymore.
- “Selected” (focused) state.

### DIFF
--- a/src/components/Input/TextInput.js
+++ b/src/components/Input/TextInput.js
@@ -5,11 +5,12 @@ import { warnOnce } from '../../utils'
 import { textStyle, GU, RADIUS } from '../../style'
 
 // Simple text input
-function TextInput({ multiline, ...props }) {
+function TextInput({ multiline, type, ...props }) {
   const theme = useTheme()
   return (
     <input
       as={multiline ? 'textarea' : 'input'}
+      type={multiline ? undefined : type}
       {...props}
       css={`
         width: ${({ wide }) => (wide ? '100%' : 'auto')};
--- a/src/components/Input/TextInput.js
+++ b/src/components/Input/TextInput.js
@@ -47,11 +47,13 @@ function TextInput({ multiline, type, ...props }) {
 TextInput.propTypes = {
   required: PropTypes.bool,
   type: PropTypes.string,
+  multiline: PropTypes.bool,
 }
 
 TextInput.defaultProps = {
   required: false,
   type: 'text',
+  multiline: false,
 }
 
 // Text input wrapped to allow adornments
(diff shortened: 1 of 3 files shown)


src/components/Input/TextInput.js
```diff
@@ -1,34 +1,47 @@
...
+          color: transparent;
+          text-shadow: 0 0 0 ${theme.surfaceContentSecondary};
+        }
+      `}
+    />
+  )
+}
 
 TextInput.propTypes = {
   required: PropTypes.bool,
```
sohkai: Perhaps we don't even need the `type` props anymore, I think React will complain for us if it tries to set a type that isn't valid?

bpierre: mmm are you thinking of only having it in the documentation? Or maybe using `PropType.string`, to not limit it to the existing list of types we have. But I consider it to be part of the API otherwise.

> They especially wouldn't be valid with multiline

Yes it shouldn’t be passed to `textarea`. :+1:

sohkai: > Or maybe using PropType.string, to not limit it to the existing list of types we have

This might make the most sense? Does react error if you try to give an `<input>` an unknown type (I thought it started doing so)?

bpierre: > Does react error if you try to give an <input> an unknown type (I thought it started doing so)?

I’m not sure, but I guess it would be the user’s responsibility at that point?

bpierre: I added a few commits :point_down:

src/components/Input/TextInput.md
```diff
@@ -1,76 +1,85 @@
 # TextInput
 
-A basic text input component. Not meant to be used for anything other than
-`type=email|number|password|search|tel|text|url`.
+A basic text input component that can use a single or multiple lines.
```
sohkai:
```suggestion
A basic text input component that can be used in single or multi-line modes.
```

src/components/Input/TextInput.md
```diff
@@ -1,76 +1,85 @@
...
 
 Set this to change the position of the adornment. `'start'` is displayed on the left, and `'end'` on the right. Note: in the future, the position will be reversed in right-to-left languages.
 
 ### `adornmentSettings.width`
 
-- Type: `Number`
-- Default: `34`
+| Type     | Default value |
+| -------- | ------------- |
+| `Number` | `34`          |
```
sohkai: I think this is 36 now?

bpierre: Yes :pray:

PR #461, opened 2019-08-12
Dropdown tweaks

bpierre:
image

Changes:

- Fix height.
- Vertically center the menu.
- Prefer `selected` over `active`.
- naming + hook dependencies fixes.

### DIFF
--- a/src/components/DropDown/DropDown.js
+++ b/src/components/DropDown/DropDown.js
@@ -164,7 +164,7 @@ const DropDown = React.memo(function DropDown({
           display: ${wide ? 'flex' : 'inline-flex'};
           justify-content: space-between;
           align-items: center;
-          height: ${6 * GU}px;
+          height: ${5 * GU}px;
           padding: 0 ${2 * GU}px;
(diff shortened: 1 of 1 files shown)


src/components/DropDown/DropDown.js
```diff
@@ -186,19 +182,11 @@ const DropDown = React.memo(function DropDown({
...
         />
       </ButtonBase>
-      <Popover
-        visible={opened}
-        onClose={handleClose}
-        opener={buttonRef.current}
-        placement="bottom-start"
-        scaleEffect={false}
-      >
+      <Popover visible={opened} onClose={close} opener={buttonRef.current}>
```
sohkai: Do we not need the `placement="bottom-start"` anymore? Otherwise I get the `DropDown` coming on top of the button, which is probably not what we want?

Screen Shot 2019-08-14 at 2 23 05 PM

bpierre: No it’s intended, that’s how it should behave now! (similar to macOS, Android and others)

image

PR #462, opened 2019-08-12
DropDown: disabled mode

bpierre:
image

@dizzypaty @owisixseven This is a proposition for a disabled state on DropDown, following the same style we use for disabled buttons (https://github.com/aragon/aragon-ui/pull/432). Please let me know if it works!


bpierre:
image

Done! Thank you :heart:

2color:
Nice! Thanks for the addition @bpierre

PR #463, opened 2019-08-13
Dataview tweaks

bpierre:
Changes:

- Pass the current mode to the render functions.
- In table mode, let cells use the full width.

### DIFF
--- a/src/components/DataView/TableView.js
+++ b/src/components/DataView/TableView.js
@@ -274,10 +274,11 @@ function EntryRow({ cells, selected, rowHeight }) {
             <div
               css={`
                 display: flex;
+                width: 100%;
                 justify-content: ${`flex-${align}`};
               `}
             >
-              <div css="display: flex">{content}</div>
+              {content}
             </div>
           </td>
         )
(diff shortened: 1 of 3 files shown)


src/components/DataView/TableView.js
```diff
@@ -274,10 +274,11 @@ function EntryRow({ cells, selected, rowHeight }) {
             <div
               css={`
                 display: flex;
+                width: 100%;
                 justify-content: ${`flex-${align}`};
               `}
             >
-              <div css="display: flex">{content}</div>
```
sohkai: Hope this doesn't break any of the current apps 🤞

bpierre: It should be fine ® :crossed_fingers:

AquiGorka: For that sake we could add `flex-direction: column` to mimic a _traditional_ block container.

bpierre: The idea of using `flex-direction: row` is because DataView’s table rows have a fixed height, making the vertical space is quite limited. It reinforces the recommended way to lay out things (horizontally).

PR #470, opened 2019-08-14
Box: adjust default padding if in primary slot of Split

AquiGorka:


bpierre:
This is quite the change! Is it because of the padding of the large box in a split?

AquiGorka:
I recalled we agreed this would be the default padding no? It is the padding for App Center and Voting's Boxes

bpierre:
But only for the big ones right? In the `primary` slot of a `Split`. I think we can change this default only in that case.

src/components/Split/Split.js
```diff
@@ -1,31 +1,38 @@
...
   const oneColumn = layout === 'small' || layout === 'medium'
   return (
-    <div
-      css={`
-        display: ${oneColumn ? 'block' : 'flex'};
-        padding-bottom: ${3 * GU}px;
-      `}
-    >
-      <div css="flex-grow: 1">{primary}</div>
+    <Inside name="Split">
```
sohkai: Wasn't really sure if we needed all these levels, but I thought we might as well for consistency's sake?

bpierre: Yes it can be useful!

bpierre: Yes it can be useful!

PR #472, opened 2019-08-14
Popover: ignore onClose() if focus is on opener

### LINKED ISSUE #477: DropDown doesn't close when trigger is re-clicked
I observed this behavior with 1.0.0.alpha.9:

dropdown click to close

When clicking away, the dropdown closes correctly. When re-clicking the trigger, it closes and then quickly re-opens.

sohkai:
There's a lot of cases where we set toggle functionality on the opener, and this doesn't play nicely with the blur handler on the popper (resulting in difficult-to-fix race conditions with the handlers). For an example, the current [DropDown](https://devbox.dweb.cool/#DropDown) has this problem.

Perhaps this should be protected behind a prop, since it could be a bit surprising if you expected the `onClose()` to always be fired when the Popover closes.

Fixes #477

### DIFF
--- a/src/components/Popover/Popover.js
+++ b/src/components/Popover/Popover.js
@@ -127,7 +127,11 @@ class PopoverBase extends React.Component {
   handleBlur = event => {
     const { opener, onClose } = this.props
     const focusedElement = event.relatedTarget
-    if (this._cardElement.current.contains(focusedElement)) {
+    if (
+      (this._cardElement.current &&
+        this._cardElement.current.contains(focusedElement)) ||
+      (opener && opener.contains(focusedElement))
+    ) {
       return
     }
 
--- a/src/components/Popover/Popover.js
+++ b/src/components/Popover/Popover.js
@@ -9,6 +9,7 @@ import RootPortal from '../RootPortal/RootPortal'
 
 class PopoverBase extends React.Component {
   static propTypes = {
+    closeOnOpenerFocus: PropTypes.bool,
     opener: PropTypes.instanceOf(Element),
     placement: PropTypes.oneOf(
       // "center" is a value that doesn’t exist in Popper, but we are using it
@@ -29,6 +30,7 @@ class PopoverBase extends React.Component {
   }
 
   static defaultProps = {
+    closeOnOpenerFocus: false,
     opener: null,
     placement: 'center',
     onClose: noop,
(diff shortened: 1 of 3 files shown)


bpierre:
That’s a really neat solution, I didn’t think of using the opener for this, super clean! :tada: 

> Perhaps this should be protected behind a prop, since it could be a bit surprising if you expected the onClose() to always be fired when the Popover closes.

Yes I agree, but still can’t imagine a valid case for it. Perhaps we could not have the prop until it becomes needed?

sohkai:
> Yes I agree, but still can’t imagine a valid case for it.

`AutoComplete` doesn't use the `Popover`, but if it ever did we may want the ability to turn off that behaviour 🤷‍♂.

But in any case it's more about the API contract we give to developers; `onClose()` sounds like it should always be triggered if the Popover is going to be closed, although in these particular cases with the opener being a toggle it's _not exactly_ the Popover itself that's closing (rather, we're controlling it from outside).

src/components/Popover/Popover.js
```diff
@@ -125,12 +127,16 @@ class PopoverBase extends React.Component {
...
     const focusedElement = event.relatedTarget
+
+    // Do not close if:
+    // - The blur event is emitted from an element inside of the popover.
+    // - The focused target is the opener and closeOnOpenerFocus is true.
     if (
       (this._cardElement.current &&
         this._cardElement.current.contains(focusedElement)) ||
-      (opener && opener.contains(focusedElement))
+      (closeOnOpenerFocus && opener && opener.contains(focusedElement))
```
sohkai: We probably want `!closeOnOpenerFocus`, since its name implies that it should close on opener focus?

bpierre: Done https://github.com/aragon/aragon-ui/pull/472/commits/581f6f570ac47e8c7574a454d281f8b89c882689

src/components/Popover/README.md
```diff
@@ -40,6 +40,14 @@ function App() {
 
 Reference to the "opener"which is what the Popover is positioning relative to.
 
+### `closeOnOpenerFocus`
+
+| Type      | Default value |
+| --------- | ------------- |
+| `Boolean` | `false`       |
+
+Do not immediatly reopen the Popover when the opener gets focused while being opened, making it possible to toggle its visibility from the opener, if its placement allows it.
```
sohkai:
```suggestion
While the Popover is opened, do not immediately close the Popover when the opener gets focused. This makes it possible to toggle its visibility from the opener, if its placement allows it.
```

PR #474, opened 2019-08-14
New style − address comments from Brett’s review

bpierre:


src/components/Button/ButtonIcon.js
```diff
@@ -43,13 +38,11 @@ function ButtonIcon({ label, children, mode, ...props }) {
...
 ButtonIcon.propTypes = {
-  mode: PropTypes.oneOf(['flat', 'button']),
   label: PropTypes.string.isRequired,
   children: PropTypes.node.isRequired,
-}
 
-ButtonIcon.defaultProps = {
-  mode: 'flat',
+  // deprecated
+  mode: PropTypes.oneOf(['flat', 'button']),
```
sohkai: I don't think we use 'flat' at all?

bpierre: No, I left it because it was there when `mode` was supported, but this is a new component, so let’s remove it yes :+1:

bpierre: I only removed this `flat` for now, and let’s remove `mode` right after the next alpha release.

PR #481, opened 2019-08-15
devbox: add a DataView demo to sort entries

bpierre:


devbox/apps/DataView.js
```diff
@@ -12,33 +15,128 @@ import {
...
+  //         { account: addr() },
+  //         { account: addr() },
+  //         { account: addr() },
+  //       ]}
+  //       renderEntry={({ account, amount }) => {
+  //         return [<IdentityBadge entity={account} />]
+  //       }}
+  //     />
+  //   ),
+  // ],
```
sohkai: Can we still have this demo?

bpierre: Ah yes, I commented it to have some space in the menu, otherwise the last one was getting cut in the medium layout. We’ll probably want more demos for DataView too. Let me try to find another way to choose from a list.

sohkai: Sounds like a dropdown 🤷‍♂

bpierre: Done in https://github.com/aragon/aragon-ui/pull/481/commits/ca45d02bc2946f84203f071b7c9cbda2c4d8754f, good idea! :bulb: 

Note: it feels nice to move from `Tabs` to `DropDown`, the props are exactly the same 🥰

PR #272, opened 2018-12-13
DatePicker - DateInput - DateRangeInput

lmcorbalan:
Closes #51 
Depends on #265 

This PR adds 3 new components:
 - DatePicker
 - DateInput
 - DateRangeInput

Thanks to @sistemico for creating DatePicker and DateInput
Thanks to @sgobotta for implementing DateRangeInput

PR #482, opened 2019-08-16
Add experimental DateRange

AquiGorka:
Experimental phase  🧪🧪🧪

Special thanks to @lmcorbalan, @sistemico, and @sgobotta for their prior work in #272 that was used and refined in the Aragon apps, eventually resulting in this component ❤️🎊


src/components/DateRange/TextInput.js
```diff
@@ -0,0 +1,144 @@
+import React from 'react'
```
sohkai: I'm pretty sure we can remove this file now, and just use the default `TextInput`.

package.json
```diff
@@ -85,6 +85,7 @@
     "airbnb-prop-types": "^2.9.0",
     "arg": "^2.0.0",
     "command-exists": "^1.2.6",
+    "date-fns": "^2.0.0-alpha.22",
```
sohkai: Perhaps we can try to remove this altogether (everything should be supported via `dayjs`?), but otherwise we should pin it.

src/components/DateRange/TextInput.js
sohkai: Done in e9c0959.

package.json
sohkai: Done in ebd596f.

sohkai:
I've updated the components to rely on dayjs instead and lightly updated their styles.

cc @dizzypaty The component current looks like:

Screen Shot 2019-08-17 at 6 55 16 PM

(I'm not sure, but did we previously decide to not show the dates of the prior / next month?

PR #483, opened 2019-08-16
Badge: fix deprecated modes and remove extra instance of BadgeNumber

sohkai:
Removes an extra instance of `BadgeNumber` and consolidates it into `Badge`.

Also fixes usage of outdated props for `Tag`.


src/components/IdentityBadge/IdentityBadgePopover.js
```diff
@@ -78,15 +78,13 @@ class IdentityBadgePopover extends React.PureComponent {
...
                 css={`
                   margin: -1px 0 0 5px;
-                  text-transform: uppercase;
-                  ${font({ size: 'xxsmall' })};
                 `}
+                size="small"
                 title="This is your Ethereum address"
               >
-                You
+                you
```
sohkai: Modified this to be the same as the apps' "You" tags.

src/components/Badge/Badge.js
```diff
@@ -63,7 +63,7 @@ function BadgeIdentity(props) {
 }
 
 function BadgeApp(props) {
-  return <Badge {...props} mode="identifier" />
+  return <Badge {...props} />
```
bpierre: This should still be using `identifier` (it was the purple badge)?

sohkai: Good catch, I missed the difference between `identifier` and `identity` 😅.

Done in fa6e453.

PR #484, opened 2019-08-16
SidePanel: update for newstyles

sohkai:
In the devbox:

Screen Shot 2019-08-16 at 1 08 42 PM

In aragon/aragon's upgrade org panel:

Screen Shot 2019-08-16 at 12 51 31 PM

Note that I've taken out the `SidePanelSplit`'s top and bottom borders, since it could collide with the header's own bottom border. Didn't think it was worth it trying to detect if the `SidePanelSplit` was the first component being rendered inside a `SidePanel` to remove the top border.


src/components/SidePanel/SidePanel.js
```diff
@@ -170,26 +190,24 @@ const PanelContent = styled.div`
...
 
-const PanelCloseButton = styled.button`
-  ${PanelHeader} & {
-    position: absolute;
-    padding: 20px;
-    top: 0;
-    right: 0;
-    cursor: pointer;
-    background: none;
+const PanelCloseButton = styled.button.attrs({
```
sohkai: @bpierre Perhaps we should add a focus ring around this too? Not sure what the best way of doing so would be (probably similar to how `Tabs` does it)?

bpierre: Yes! You can have it by doing:

```jsx
<ButtonIcon>
  <IconCross />
</ButtonIcon>
```

sohkai: Done in 58d6234.

I removed the pressed state on the button since it felt a bit odd (due to the positioning):
Screen Shot 2019-08-16 at 2 32 48 PM


But alternatively we could make it look something like (but this feels quite large and awkward):
Screen Shot 2019-08-16 at 2 32 24 PM

bpierre: I think the first one is ok? It only appear on press, and that’s what we do everywhere else.

AquiGorka: For mobile it makes sense to have the 2nd one

sohkai: Reverted to normal behaviour in 6c7b56e.

bpierre: > For mobile it makes sense to have the 2nd one

Yes :+1: I think we are doing the switch between the two in some places already?

AquiGorka: Yep, all around, you taught me that, for mobile, make bigger areas for button press :)

bpierre: I’m thinking that it could be nice to also provide a `hasTouchCapabilty()` utility, to know if the screen has touch capabilities (rather than assuming that small viewports are touch). That way we could can have these big touch targets on large touch screens as well (e.g. the iPad pro) and stop having them on non-touch screens that happen to have a small viewport (e.g. a browser window not fullscreen).

sohkai: Switched to larger button in 3e87126.

AquiGorka: > That way we could can have these big touch targets on large touch screens as well 
Some desktops have touch enabled screens too, although if the designs are in sync, it should work out of the box even with the larger area buttons.

bpierre: > Some desktops have touch enabled screens too

Exactly, mine is haha :-)

sohkai:
@bpierre I've updated this to add the `Inside` scope to the SidePanel and detect in `Tabs` to avoid rendering an extra top border.

PR #486, opened 2019-08-16
ContextMenu: adapt for new styles

sohkai:
Applies some more changes to align the `ContextMenu` closer to the designs in the apps (e.g. [Tokens](https://www.figma.com/file/M9AbLkUkeRCFKQt6qdFUsCks/Tokens?node-id=1%3A2)).

@dizzypaty Wasn't sure what the behaviour should be in the pressed mode, but it felt most appropriate to make the entire item "pressed":

contextmenu


src/components/ContextMenu/ContextMenu.js
```diff
@@ -11,7 +11,7 @@ import { ButtonBase } from '../Button/ButtonBase'
 const BASE_WIDTH = 46
 const BASE_HEIGHT = 32
 
-function ContextMenu({ children }) {
+function ContextMenu({ children, zIndex }) {
```
sohkai: Not sure if there's a better way to achieve this D:

devbox/apps/ContextMenu.js
```diff
@@ -0,0 +1,76 @@
...
+          <p
+            css={`
+              ${textStyle('body2')}
+            `}
+          >
+            Actions
+          </p>
+          <ContextMenu>
+            {actions.map(([onClick, Icon, label], index) => (
+              <ContextMenuItem onClick={onClick} key={index}>
```
bpierre: While we are at refreshing this component: what would you think of moving to an `items` prop, like `Tabs` and `DropDown` do? We could have a `onSelect` or `onActivate` on `ContextMenu` itself, with the corresponding index. We could keep the compatibility with `ContextMenuItem` for a while (and use `warnOnce`).

src/components/ContextMenu/ContextMenu.js
bpierre: Yes, everything that should be rendered on top of should be using `RootPortal`… and perhaps `Popover` directly in that case? I think we should avoid to ask app authors to set a `z-index` on our components,  they should work out of the box − in normal scenarios.

PR #488, opened 2019-08-16
Style: prefer ternary declarations in dynamic styled-components

sohkai:
I've since learned that passing around booleans in declaring styled-components can lead to these errors:

Screen Shot 2019-08-16 at 12 59 44 PM

As you said @AquiGorka, it is preferable to use the ternary syntax :).

These components generally don't error, but we may as well use this syntax consistently when defining dynamically styled styled-components.

### DIFF
--- a/src/components/DropDown/DropDown.js
+++ b/src/components/DropDown/DropDown.js
@@ -169,13 +169,14 @@ const DropDown = React.memo(function DropDown({
           border: ${disabled ? 0 : 1}px solid
             ${closedWithChanges ? theme.selected : theme.border};
           ${textStyle('body2')};
-          ${disabled && 'font-weight: 600;'}
-          ${!disabled &&
-            `
+          ${disabled ? 'font-weight: 600;' : ''}
+          ${!disabled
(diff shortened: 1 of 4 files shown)


AquiGorka:
Yep, this is the error I mentioned.

src/components/DropDown/DropDown.js
```diff
@@ -169,13 +169,14 @@ const DropDown = React.memo(function DropDown({
           border: ${disabled ? 0 : 1}px solid
             ${closedWithChanges ? theme.selected : theme.border};
           ${textStyle('body2')};
-          ${disabled && 'font-weight: 600;'}
-          ${!disabled &&
-            `
+          ${disabled ? 'font-weight: 600;' : ''}
```
AquiGorka: When there is more than one style that depends on the condition these could be grouped together no?

sohkai: Yes, missed that while making these changes! Good eye :).

Fixed in https://github.com/aragon/aragon-ui/pull/503.

PR #492, opened 2019-08-18
Tabs: update border colour when in full width mode

sohkai:
Updates the `Tabs`'s full width version to use a lighter border-colour.

Not too sure, but this seems consistent in most of the design files I've seen (e.g. [this one for Permissions](https://www.figma.com/file/VmfsSnRMDlxcmxTmHP1WwiLR/Permissions?node-id=4%3A12009)).

### DIFF
--- a/src/components/Tabs/TabsFullWidth.js
+++ b/src/components/Tabs/TabsFullWidth.js
@@ -72,7 +72,7 @@ function TabsFullWidth({ items, selected, onChange }) {
             height: 100%;
             background: ${theme.surface};
             border-style: solid;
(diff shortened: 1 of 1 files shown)


bpierre:
Not sure we should respect this, I don’t remember but I might have asked @dizzypaty before deciding to go for the normal `border`?

dizzypaty:
This must have been a mistake in one of the instances that we copied / reused in other files. All the surfaces & components should use color `border` at 100% opacity. Really sorry for the inconvenience @sohkai!

PR #494, opened 2019-08-18
Modal: add a prop to display a close button

bpierre:


src/components/Modal/Modal.js
```diff
@@ -1,112 +1,122 @@
...
+          (({ opacity, scale }) => (
+            <animated.div
+              css={`
+                position: absolute;
+                z-index: 1;
+                top: 0;
+                left: 0;
+                right: 0;
+                bottom: 0;
+                background: ${theme.overlay.alpha(0.9)};
```
sohkai: Should we do this for other overlays, like the `SidePanel` too?

bpierre: Yes, doing it :+1:

bpierre:
> Let's add it to the README as well :).

Done https://github.com/aragon/aragon-ui/pull/494/commits/85d69870dd4115abddb50117fb0e5216dac1ef76

PR #498, opened 2019-08-18
Text styles: add “Title 0”

bpierre:


bpierre:
haha yes :laughing: Maybe we should have gone for `title100`, `title200` etc. (like Google does for colors) to give us some space :thinking: 

I like `banner`, but I don’t like that it introduces a new category on its own… it’s really a title, just bigger than the others. One thing I like about `title0` is that it keeps `title1` as the main title app authors should use.

I wonder if we want to add it at all, as the onboarding is the only part to use it, and it has been designed before the work on the text styles. What would you think about closing this until we discuss with @owisixseven and @dizzypaty, and have this text style being specific to aragon/aragon for now?

sohkai:
Sounds good to me!

Issue #333, opened 2019-03-01
Grid unit utility


bpierre:
Historically, we’ve been using a `10px` base grid for most of the things, then started to introduce a `8px` grid for some things. In the future, we want to be using the `8px` grid only.

I think having a utility for this would help to make the transition to `8px` faster and easier to follow. There are of course many cases where we don’t want grid units. It is not about doing everything with it, but more as a convenience tool to make sure we use multiples of the grid unit when we want it. It would also help making a semantic distinction between e.g. `8px` when we want exactly `8px`, and `gu(1)` because we want to use one grid unit that happens to be `8px` too.

In the Aragon’s [App Center PR](https://github.com/aragon/aragon/pull/625), I started experimenting with using a constant to reinforce the grid unit we use for spacing:

```jsx
import { GU } from '../utils'
const App = () => <div css={`margin: ${GU}px ${4 * GU}px`} />
```

The problem is that it’s pretty verbose.

Another option would be to expose a function:

```jsx
import { gu } from '../utils'

const App = () => <div css={`margin: ${gu(1)}px ${gu(4)}px`} />

// We could even add `px` by default:

const App = () => <div css={`margin: ${gu(1)} ${gu(4)}`} />
```

But it’s still quite verbose. It would also add a lot of function calls, but it probably doesn’t have any significant impact on modern JS engines.

Another solution to make it less verbose could be to [extend](https://github.com/Losant/styled-transform-proxy) the way styled-components transforms styles to add a custom syntax:

```jsx
// As a CSS unit
const App = () => <div css="margin: 1gu 4gu" />

// As a CSS function
const App = () => <div css="margin: gu(1) gu(4)" />
```

Another solution could have been to use a [CSS variable](https://developer.mozilla.org/en-US/docs/Web/CSS/Using_CSS_variables) for it, but they are too verbose:

```jsx
const App = () => <div css="margin: calc(1 * var(--gu)) calc(4 * var(--gu))" />
```

## Proposal 1

As a first step:

- Make the `GU` constant available (`GU = 8`).
- Export a function named `g()` rather than `gu()` (shorter).
- `g()` would add `px` to the value, e.g. `g(2)` would return `16px`.

That way, we could use both GU and g() depending on what is needed. I don’t like the naming much, but they would be so common that I think conciseness is the most important aspect here. We could also have long name equivalents (`GRID_UNIT` and `gridUnit()`), but I don’t think they would be used at all.

It would means having this kind of syntax for now:

```jsx
[...]

PR #500, opened 2019-08-18
Add a babel plugin to handle the gu unit

bpierre:
See discussion: https://github.com/aragon/aragon-ui/issues/333

# babel-plugin-aragon-ui

Plugin greatly inspired from [babel-plugin-styled-components](https://github.com/styled-components/babel-plugin-styled-components), and made to work with it.

## What it does

Transforms this:

```jsx
<div
  css={`
    width: 4gu;
    height: 4gu;
  `}
/>
```

Into this:

```jsx
<div
  css={`
    width: 32px;
    height: 32px;
  `}
/>
```

## Features

- Adds the `gu` unit to styled component’s CSS blocks (in normal styled
  components, in the `css` prop, or using the `css` helper).

## Installation

In your .babelrc, declare this plugin before styled-components:

```json
{
  "plugins": [
    "@aragon/babel-plugin-aragon-ui",
    "styled-components"
  ]
}
```

## To do

- Support for text styles: `-ui-text-style: title1`
- Support for theme values: `-ui-theme(surfaceContent)`

Note: both are using the [CSS syntax for proprietary extensions](https://www.w3.org/TR/CSS22/syndata.html#vendor-keywords), but we might as well decide to drop them, or not follow it exactly (e.g. by removing the initial `-`).

### DIFF
--- a/babel-plugin-aragon-ui/README.md
+++ b/babel-plugin-aragon-ui/README.md
@@ -1,0 +1,22 @@
+# babel-plugin-aragon-ui
+
+Plugin greatly inspired from [babel-plugin-styled-components](https://github.com/styled-components/babel-plugin-styled-components), and made to work with it.
+
+## Features
+
+- Adds the `gu` unit to styled component’s CSS blocks (in normal styled
+  components, in the `css` prop, or using the `css` helper).
+- That’s all for now.
+
+## Installation
+
+In your .babelrc, declare this plugin before styled-components:
+
+```json
+{
+  "plugins": [
+    "@aragon/babel-plugin-aragon-ui",
+    "styled-components"
+  ]
+}
+```
--- a/babel-plugin-aragon-ui/src/gu.js
+++ b/babel-plugin-aragon-ui/src/gu.js
@@ -1,0 +1,53 @@
+import { isStyled, isHelper } from './utils/detectors'
+
+const GU = 8
+const GU_REGEX = /([^0-9a-zA-Z])((-?[0-9]*\.?[0-9]+)gu)/gm
+
+function guReplacer(match, p1, p2, p3) {
+  const value = parseInt(p3, 10)
+  if (isNaN(value)) {
+    return match
+  }
+  return `${p1}${value * GU}px`
+}
+
+function replaceGus(value) {
+  return value.replace(GU_REGEX, guReplacer)
+}
+
+// Transform a styled component or helper (e.g. css``)
+export function guStyledComponent(t, path, state) {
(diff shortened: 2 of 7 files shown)


babel-plugin-aragon-ui/src/gu.js
```diff
@@ -0,0 +1,53 @@
+import { isStyled, isHelper } from './utils/detectors'
+
+const GU = 8
```
AquiGorka: Would it make sense to bring this from `aragonUI`?
`import { GU } from '@aragon/ui'`

babel-plugin-aragon-ui/README.md
```diff
@@ -0,0 +1,22 @@
+# babel-plugin-aragon-ui
+
+Plugin greatly inspired from [babel-plugin-styled-components](https://github.com/styled-components/babel-plugin-styled-components), and made to work with it.
+
+## Features
+
+- Adds the `gu` unit to styled component’s CSS blocks (in normal styled
```
AquiGorka: This paragraph sounds like the description, maybe it should go right after the h1?
Should we add some usage?

babel-plugin-aragon-ui/src/gu.js
bpierre: Yes I thought about it, but it would mean either having a specific aragonUI as a dependency, or having it as a peerDependency, which doesn’t work since we also want to use it from aragonUI itself :laughing: 

So I thought it’s just simpler to have it hardcoded. Another option would be to transform to `x * GU`, and add the import dynamically, but it seems a bit too much for something that should not change often, if ever.

AquiGorka: Yeah, makes sense, tough circular dependency, we'll need to keep this in sync or we might end up with two different values for GU

babel-plugin-aragon-ui/README.md
bpierre: Yes I need to make this a bit better, with some examples to. I also added some “To do” items in the PR description that I’ll add later.

Evalir:
Should we pick this up again or just close? A babel plugin for `GU` is quite nice but maybe doesn't align with current priorities. cc @bpierre

bpierre:
@Evalir I think we could merge it in its current state, since it is functional and separated from aragonUI. The idea of leaving this open was to have other people testing it a little bit more, but we or someone else can iterate on it in separate PRs. What do you think?

Evalir:
Sounds good to me! Let's merge and then iterate on it. 😬

PR #501, opened 2019-08-18
Sidepanel tweaks

bpierre:
- Move to `css`.
- Indentation / formatting.
- Fix the overlay color.
- Fix the shadow color.

### DIFF
--- a/src/components/SidePanel/SidePanel.js
+++ b/src/components/SidePanel/SidePanel.js
@@ -61,7 +61,6 @@ function SidePanel({
         >
           {({ progress }) => (
             <div
-              opened={opened}
               css={`
                 position: fixed;
                 z-index: 3;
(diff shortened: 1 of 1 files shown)


src/components/SidePanel/SidePanel.js
```diff
@@ -191,27 +205,14 @@ const Panel = React.memo(function Panel({ compact, ...props }) {
         width: 100vw;
         height: 100vh;
         background: ${theme.surface};
-        box-shadow: -2px 0px 4px rgba(118, 137, 173, 0.2);
-
+        box-shadow: -2px 0px 4px rgba(0, 0, 0, 0.1);
```
sohkai: Not sure, but all the [designs](https://www.figma.com/file/M9AbLkUkeRCFKQt6qdFUsCks/Tokens?node-id=200%3A0) have it at the previous value.

bpierre: Ah yes I assume it’s a mistake. It’s important to use black (#000) as a base color for shadows, so they look correct on dark colors (and not like they are glowing). /cc @dizzypaty @owisixseven

bpierre: Confirmed with @owisixseven offline: shadows should always be black.

PR #506, opened 2019-08-19
DateRange: adjust input field styling

sohkai:
Uses `theme.hint` as the colour, and adjusts its text size to be inline with other text inputs.

Previously:

Screen Shot 2019-08-19 at 12 28 49 PM

Now:

Screen Shot 2019-08-19 at 12 25 46 PM


bpierre:
This component just keeps looking better haha :heart:

PR #507, opened 2019-08-19
Button + icon adjustments

bpierre:
The Button geometry is now based on the new specification by @owisixseven:

Also, all the icons have been updated to fix their alignment on buttons.

Changes:
- New horizontal padding.
- New size: `mini`.
- Deprecated sizes: `large`, `normal` (which becomes `medium` like the other size-related props).
- Add `body4` to the text styles.
- Adapt the icon size to the button size.
- Add `Inside` slots to `Button`.
- Update the devbox demo.
- Reorganize a bit the way styles are generated in `Button`.
- Add `QuestionIcon`.


devbox/apps/Button.js
```diff
@@ -13,7 +13,7 @@ import {
 } from '@aragon/ui'
 
 const MODES = ['normal', 'strong', 'positive', 'negative']
-const SIZES = ['normal', 'small']
+const SIZES = ['normal', 'small', 'mini']
```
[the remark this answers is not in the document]
bpierre: Yes!

src/components/Button/Button.js
```diff
@@ -6,8 +6,82 @@ import { warn, warnOnce, useInside } from '../../utils'
...
-function buttonStyles(theme, { mode, disabled }) {
+// Base styles related to every size.
+// See src/icons/icon-size.js for the corresponding icon sizes.
+const SIZE_STYLES = {
+  medium: {
+    textStyleName: 'body2',
+    height: 5 * GU,
+    padding: 3 * GU,
+    iconPadding: 2 * GU,
+    minWidth: 14.5 * GU,
```
sohkai: Hmm, the specs have the icon + label version with a smaller min-width than just the label version, should this value change based on that?

bpierre: Do you mean having a different smaller minimum width when there is no icon displayed?

I was wondering if it does make sense to apply the minimum width on buttons that are not text only? IIRC it’s mainly to fix how buttons like “OK” and “Cancel” look? At the moment, it can sometimes look a bit odd when a minimum width is applied to icon+text buttons, and the minimum width is only slightly larger than the natural one. /cc @owisixseven @dizzypaty

bpierre:
> we should also update the README with these new modes :).

Done https://github.com/aragon/aragon-ui/pull/507/commits/b11bd623a2e5b8cf3e385dc504c00213f8316f36

src/components/Button/Button.js
dizzypaty: It's ok for me to remove the min width for buttons containing icon + label.

PR #508, opened 2019-08-19
Box: align heading's padding with content

sohkai:


bpierre:
Ah I tried this at some point, but it was looking really weird… I think we want a fixed padding for the header?

sohkai:
I think we should apply it as a default, since we let users customize it if they provide their own component.

Otherwise we have slight misalignment by default if we provide a padding override to Box:

Screen Shot 2019-08-19 at 7 48 20 PM

And fixed with this change:

Screen Shot 2019-08-19 at 7 48 07 PM

bpierre:
Yes that’s why I added it in the first place :-)

But then I noticed it was looking really weird in some cases:

- When the padding is even more important. The header in your second image is already a bit unbalanced, and the more padding you add, the more it accentuates this effect.
- When two boxes are following each other and don’t have the same padding (because they display different content), the titles are misaligned.

sohkai:
> When two boxes are following each other and don’t have the same padding (because they display different content), the titles are misaligned.

Hmm yes, this would look awkward. Especially in cases like Split, although we usually never use a heading in the Box when it's in the primarily slot of Split.

Although now I'm realizing a lot of the designs we have use a 3GU padding around the box and we've generally kept it at 2GU.

src/components/Box/Box.js
```diff
@@ -84,7 +84,7 @@ function HeaderContent({ heading }) {
   return (
     <h1
       css={`
-        padding-left: ${2 * GU}px;
+        padding-left: ${padding}px;
```
bpierre: Should we apply it to both sides? I’m thinking of the App Center case with a label on each side.

sohkai: Done in 1efe0f6.

PR #509, opened 2019-08-20
Modal updates to fix some z-index issues

AquiGorka:


bpierre:
Could we pass `...props` to the element, so we can apply a `z-index` using `css`, but without having it part of the official API? I’d really like to avoid asking authors to manage `z-index`es.

PR #512, opened 2019-08-20
Header: prevent text overflow / wrap

bpierre:
<p align=center>

### DIFF
--- a/src/components/Header/Header.js
+++ b/src/components/Header/Header.js
@@ -32,18 +32,32 @@ function Header({ primary, secondary, children, ...props }) {
         >
           {children || (
             <>
-              <div>
-                <Inside name="Header:primary">
-                  {typeof primary === 'string' ? (
+              <Inside name="Header:primary">
+                <div
+                  css={`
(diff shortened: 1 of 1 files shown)


src/components/Header/Header.js
```diff
@@ -65,9 +79,10 @@ Header.Title = function HeaderTitle({ children, ...props }) {
   return (
     <h1
       css={`
-        display: inline-flex;
-        align-items: center;
         color: ${theme.content};
+        overflow: hidden;
+        text-overflow: ellipsis;
+        white-space: nowrap;
```
sohkai: Does this need a `text-align: left`? I was having an issue today where applying `overflow: hidden` would change the text alignment to center 🤷‍♂

bpierre: It shouldn’t be, maybe it was because of the parent?

sohkai: Ahh yes, these were buttons and the usual browser default for text-align on buttons is center 🤦‍♂

PR #523, opened 2019-08-21
Table: new style update: useTheme() + GU

bpierre:
More `useTheme()`

### DIFF
--- a/src/components/Table/TableHeader.js
+++ b/src/components/Table/TableHeader.js
@@ -1,26 +1,28 @@
 import React from 'react'
 import PropTypes from 'prop-types'
-import styled from 'styled-components'
-import { theme } from '../../theme-legacy'
-import Text from '../Text/Text'
+import { GU, textStyle } from '../../style'
+import { useTheme } from '../../theme'
 
-const StyledTableHeader = styled.th`
-  padding: 0;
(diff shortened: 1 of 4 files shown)


src/components/Table/TableHeader.js
```diff
@@ -1,26 +1,28 @@
...
-  <StyledTableHeader align={align} {...props}>
-    <Text.Block color={theme.textSecondary} smallcaps>
+function TableHeader({ title, align, ...props }) {
+  const theme = useTheme()
+  return (
+    <th
+      css={`
+        padding: 0;
+        padding-left: ${align === 'left' ? 2.5 * GU + 1 : 0}px;
+        padding-right: ${align === 'right' ? 2.5 * GU + 1 : 0}px;
```
sohkai: This feels like it's cheating a bit 😄

bpierre:
> Are we planning on using this internally or is it just meant as compatibility?

Just as compatibility for now, but it might be nice to keep it in case we need really simple tables one day.

src/components/Table/TableHeader.js
bpierre: haha yes, the `1` is for the border-width applied to the `td` below :o

PR #525, opened 2019-08-22
Link: add styles, move the core features to LinkBase

bpierre:
This is the first step of moving away from `ButtonText`.

Unlike the `ButtonText` component, `Link` can be used in the middle of any inline content: it inherits the parent font size, can be selected, and doesn’t have any specific padding or height.

The next step will be to add a `text` mode to `Button`, that will look very similar to the current `ButtonText`.


src/components/Link/Link.js
```diff
@@ -1,25 +1,42 @@
...
-  rel: PropTypes.string,
+  return (
+    <LinkBase
+      href={href}
+      onClick={onClick}
+      onKeyDown={handleKeyDown}
+      {...props}
+      css={`
+        color: ${theme.link};
+        text-decoration: none;
```
AquiGorka: Should we add the `text-decoration` on hover?

src/components/Link/Link.js
```diff
@@ -1,25 +1,42 @@
...
-  external: PropTypes.bool,
-  rel: PropTypes.string,
+  return (
+    <LinkBase
+      href={href}
+      onClick={onClick}
+      onKeyDown={handleKeyDown}
+      {...props}
+      css={`
+        color: ${theme.link};
```
AquiGorka: Ja! :blue-links:!

src/components/Link/Link.js
bpierre: I don’t think we have any hover state on the links? /cc @dizzypaty @owisixseven 

It could be nice to have a pressed state though, as we have on other interactive components.

AquiGorka: Isn't hover:underline a known UX behaviour for links?

bpierre: I don’t know about general UX recommendations, but from an accessibility point of view, links shouldn’t only be differentiated only by color (not only on hover, but always).

From a UI point of view, any hover state would only be visible by users of pointer devices (no hover on touch screens), so it’s really something that is optional that give a very specific feel: for example, Apple doesn’t use any hover state on macOS, while other platforms do. Now we don’t use much hover states in aragonUI components, so I tend to find it inconsistent if we only use them for specific components and not others: I would rather use them everywhere or not at all.

Pressed states are really important though, as they give an immediate feedback when interacting with the element.

AquiGorka: Good rationale. Yet, I'd be in favour of underlinying (?) on hover for links - unless some specific reason not to, links are links, I don't disagree with having our own considerations but overriding well known behaviour I feel is a hard call.

bpierre: Something else to consider is that we these are most of the time going to be used as buttons rather than actual links. We are also going to have a `<Button mode="text">`, with the same text styles, but the dimensions of a normal button. Having underlines on these two components would work, but we need to ensure it’s consistent with the rest, and if the having underlines don’t make them look too much like HTTP links.

Another possibility could be to have a prop to let the user of the component decide on whether or not an underline should be displayed (since a `onClick` could end up opening a URL, and a `href` could be just changing the hash in an app, it’s hard to know what is a link vs. a button).

But again, if we decide to go for using underlines, I think they should be used for the normal state (like HTML links are), not only on hover for pointer devices.

bpierre: @dizzypaty @owisixseven @AquiGorka Getting this PR merged until we decide what to do, so we can start using this component in the meantime.

src/components/LinkBase/README.md
```diff
@@ -0,0 +1,67 @@
...
+import { LinkBase } from '@aragon/ui'
+
+function App() {
+  return <LinkBase href="https://example.com/">Example</LinkBase>
+}
+```
+
+## Props
+
+### onClick
```
sohkai: Does LinkBase support `onClick`? It's only Link, no?

src/components/Link/Link.js
```diff
@@ -1,25 +1,42 @@
...
+    />
+  )
 }
 
-Link.defaultProps = {
-  external: false,
-  rel: 'noopener noreferrer',
+Link.propTypes = {
+  ...LinkBase.propTypes,
+  href: PropTypes.string,
```
sohkai: Should this belong in LinkBase instead?

bpierre: Yes :+1:

src/components/LinkBase/README.md
bpierre: Yes you’re right, and it doesn’t make sense to have it on `LinkBase` since we can use `ButtonBase` in these cases.

bpierre: Thinking more about it, it might be better to move the `onClick` handler to `LinkBase`, and keep `Link` for stylistic purposes only.

PR #526, opened 2019-08-22
Keydown fixes

bpierre:
-  `TabsFullWidth`: only listen to keydown events when needed.
-  `useKeyDown()`: prevent the event callback to be recreated on every render.

### DIFF
--- a/src/components/Tabs/TabsFullWidth.js
+++ b/src/components/Tabs/TabsFullWidth.js
@@ -106,13 +106,18 @@ function TabsFullWidth({ items, selected, onChange }) {
             css={`
               display: flex;
               align-items: center;
-              padding: 0 ${2 * GU}px;
+              justify-content: center;
+              width: ${7 * GU}px;
+              height: 100%;
               color: ${theme.surfaceIcon};
-              transition: transform 150ms ease-in-out;
-              transform: rotate3d(0, 0, 1, ${opened ? 180 : 0}deg);
             `}
           >
-            <IconDown />
+            <IconDown
+              css={`
+                transition: transform 150ms ease-in-out;
+                transform: rotate3d(0, 0, 1, ${opened ? 180 : 0}deg);
+              `}
+            />
           </div>
         </ButtonBase>
         <Transition
--- a/src/components/Tabs/TabsFullWidth.js
+++ b/src/components/Tabs/TabsFullWidth.js
@@ -1,13 +1,13 @@
 import React, { useCallback, useEffect, useState } from 'react'
 import { Transition, animated } from 'react-spring'
 import { GU, textStyle, springs } from '../../style'
-import { useOnBlur, useKeyDown } from '../../hooks'
+import { useOnBlur } from '../../hooks'
 import { IconDown } from '../../icons'
 import { useTheme } from '../../theme'
 import { useInside } from '../../utils'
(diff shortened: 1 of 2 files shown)


src/components/Tabs/TabsFullWidth.js
```diff
@@ -41,9 +41,23 @@ function TabsFullWidth({ items, selected, onChange }) {
...
+
+    const onKeyDown = event => {
+      if (event.keyCode === KEY_ESC) {
+        close()
+      }
+    }
+    window.addEventListener('keydown', onKeyDown)
+
+    return () => {
+      window.removeEventListener('keydown', onKeyDown)
```
sohkai: This shouldn't be a problem in the case we don't register an event listener (when the tabs menu is closed), right?

bpierre: No it can only get called after `addEventListener()`, but even if it was the case, calling `removeEventListener` with a function that hasn’t been added does nothing.

Note: I removed this part to use `onKeyDown` on the menu directly.

src/components/Tabs/TabsFullWidth.js
```diff
@@ -1,34 +1,37 @@
...
 import { GU, textStyle, springs } from '../../style'
-import { useOnBlur, useKeyDown } from '../../hooks'
+import { useOnBlur } from '../../hooks'
 import { IconDown } from '../../icons'
 import { useTheme } from '../../theme'
 import { useInside } from '../../utils'
 import { ButtonBase } from '../Button/ButtonBase'
 
-const ESC_CODE = 27
+const KEY_ESC = 27
```
AquiGorka: Am I wrong to think there is a _keycodes_ object with constants somewhere?

src/components/Tabs/TabsFullWidth.js
```diff
@@ -37,13 +40,24 @@ function TabsFullWidth({ items, selected, onChange }) {
...
-  useKeyDown(ESC_CODE, () => {
+  // close on escape
+  const handleMenuKeyDown = useCallback(
+    event => {
+      if (event.keyCode === KEY_ESC) {
+        close()
+        focusButton()
+      }
+    },
+    [close]
```
AquiGorka: `focusButton` has no deps, so it won't change, so it does not have to be a dep here?

src/components/Tabs/TabsFullWidth.js
bpierre: [We have one](https://github.com/aragon/aragon/blob/master/src/keycodes.js) on `aragon/aragon`, but we could have one on aragonUI yes :-)

src/components/Tabs/TabsFullWidth.js
bpierre: Ah no I forgot this one! I wonder what is the consensus in the React community, but for now I always add user-created callbacks / memoized values in the dependencies.

src/components/Tabs/TabsFullWidth.js
bpierre: (I’ll do this as a separate PR)

src/components/Tabs/TabsFullWidth.js
sohkai: Agreed, we should do to protect ourselves in case a user-created callback / memoized value in the future gets changed to have dependencies.

PR #527, opened 2019-08-23
Move back to a local scroll

bpierre:
This fixes the issue when the main scrollbar was displayed on top of other components like SidePanel.

- Add a ScrollView component (injected by Main).
- Both the layout and the scrollview can be disabled in Main.
- Update Modal, SidePanel and Popover.


src/components/Layout/Layout.js
```diff
@@ -78,18 +85,20 @@ function Layout({ children, parentWidth, breakpoints, ...props }) {
...
     large: PropTypes.number,
   }),
+  children: PropTypes.node,
+  paddingBottom: PropTypes.oneOfType([PropTypes.number, PropTypes.string]),
+  parentWidth: PropTypes.number,
 }
 
 Layout.defaultProps = {
   breakpoints: {},
+  paddingBottom: 3 * GU,
```
AquiGorka: Room to breathe at the end or soemthing else?

src/components/Main/Main.js
```diff
@@ -16,23 +17,35 @@ const [
...
+  if (scrollView === undefined) {
+    scrollView = !hasAppView
+  }
+
+  // Optionally wrap `children` with Layout and/or ScrollView
+  let content = layout ? <Layout>{children}</Layout> : children
+  content = scrollView ? (
+    // The main ScrollView is set to 100vh by default (best for Aragon apps)
+    // Disable `scrollView` and insert your own if needed.
+    <ScrollView css="height: 100vh">{content}</ScrollView>
```
AquiGorka: Would it make sense that `ScrollView` injects the correct paddings/spacing for the different views too?

src/components/Layout/Layout.js
bpierre: Yes, it is the padding at the end of every app. It was already here, but it is now configurable :-)

src/components/Main/Main.js
bpierre: I’d like to let `Layout` handle that, so that `ScrollView` stays a pure scroll component (and the two are generally used together).

AquiGorka: `Layout` is not part of main no?

bpierre: It is, one line above this one! :point_up:

AquiGorka: So apps should have the appropriate spacing towards edge of screen out of the box? cc @chadoh

bpierre: Oh yes they already do! `Layout` is taking care of that :smile:

PR #528, opened 2019-08-23
SidePanel: update title color

AquiGorka:


src/components/SidePanel/SidePanel.js
```diff
@@ -116,6 +116,7 @@ function SidePanel({
                   <h1
                     css={`
                       ${textStyle('body1')}
+                      color: ${theme.content};
```
bpierre: mmm we should have `surfaceContent` here (since we are on a `surface` background), is it because `content` matches the color in the design screens?

AquiGorka: Done. Both had the same code yea.

bpierre: Oh yes I forgot that! :laughing:

PR #530, opened 2019-08-23
Various fixes

bpierre:
- Update devbox Tabs.
- FocusVisible is now a non-pure component.
- Fix LinkBase import.
- Improve Modal transition.
- Ensure Split is always 100% wide.
- Components index order.
- Use `body2` as a default text style.
- Update package-lock.json.


sohkai:
> Use body2 as a default text style.

🙌

src/components/index.js
```diff
@@ -1,20 +1,22 @@
 export { Bar } from './Bar/Bar'
 export { ButtonBase } from './Button/ButtonBase'
 export { DataView } from './DataView/DataView'
+export { DateRange as _DateRange } from './DateRange/DateRange'
```
sohkai: I moved this to be after the `_AutoComplete` export because it's an experimental component

bpierre: Oh right, I’ll move it back.

PR #533, opened 2019-08-24
Card: use ButtonBase when interactive

bpierre:
To make the keyboard activation work.

@AquiGorka Thanks for suggesting this! I couldn’t notice any change in the cards layout for Permissions, Voting, or App Center, so it seems to be working just fine.

### DIFF
--- a/src/components/Card/Card.js
+++ b/src/components/Card/Card.js
@@ -4,7 +4,7 @@ import { css } from 'styled-components'
 import { GU, RADIUS } from '../../style'
 import { useTheme } from '../../theme'
 import { unselectable, useInside } from '../../utils'
-import FocusVisible from '../FocusVisible/FocusVisible'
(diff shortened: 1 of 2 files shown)


src/components/Card/Card.js
```diff
@@ -25,61 +25,51 @@ function Card({ children, width, height, onClick, ...props }) {
...
+            }
 
-                &:focus {
-                  outline: 0;
-                  .focus-ring {
-                    display: block !important;
-                  }
-                }
-              `}
+            &:focus {
```
sohkai: Do we still need this css, now that ButtonBase is handling the focus?

bpierre: We don’t, I realized this afterwards :s https://github.com/aragon/aragon-ui/pull/391/commits/e88fce1a7dbde98a6aee6123a2ec6cb40d26e34b

PR #535, opened 2019-08-27
DataView child content: dynamic height + rename it to “expansion”

bpierre:
@sembrestels @deamme let me know if it does the job for the conviction voting app! :smile:

### DIFF
--- a/src/components/DataView/DataView.js
+++ b/src/components/DataView/DataView.js
@@ -1,6 +1,6 @@
 import React, { useCallback, useMemo, useState } from 'react'
 import PropTypes from 'prop-types'
-import { noop } from '../../utils'
+import { noop, warnOnce } from '../../utils'
 import { textStyle, GU } from '../../style'
 import { useTheme } from '../../theme'
 import { Box } from '../../components/Box/Box'
@@ -35,28 +35,31 @@ function prepareFields(fields) {
   })
 }
 
-function entryChildrenFromChild(child) {
-  if (!child) {
+function entryExpansion(content) {
+  // Rows
+  if (Array.isArray(content) && content.length > 0) {
     return {
-      children: null,
-      singleNodeChildren: false,
+      content,
+      freeLayout: false,
     }
   }
-  if (Array.isArray(child)) {
+  // Free layout
+  if (content && !Array.isArray(content)) {
     return {
-      children: child,
-      singleNodeChildren: false,
+      content: [content],
+      freeLayout: true,
     }
   }
+  // No expansion
   return {
-    children: [child],
-    singleNodeChildren: true,
+    content: [],
+    freeLayout: false,
   }
 }
 
 function renderEntries(
   entries,
-  { fields, renderEntry, renderEntryActions, renderEntryChild, mode }
+  { fields, renderEntry, renderEntryActions, renderEntryExpansion, mode }
 ) {
   return entries.map(entry => {
     const { value, index, selected } = entry
@@ -72,9 +75,9 @@ function renderEntries(
       entryNodes.push(null)
     }
 
-    const { children, singleNodeChildren } = entryChildrenFromChild(
-      renderEntryChild
-        ? renderEntryChild(value, index, { selected, mode })
+    const expansion = entryExpansion(
+      renderEntryExpansion
+        ? renderEntryExpansion(value, index, { selected, mode })
         : null
     )
 
@@ -84,11 +87,10 @@ function renderEntries(
 
     return {
       actions,
-      children,
       entryNodes,
+      expansion,
       index,
       selected,
-      singleNodeChildren,
     }
   })
 }
@@ -167,16 +169,30 @@ const DataView = React.memo(function DataView({
   renderEntry,
   renderEntryActions,
   renderEntryChild,
+  renderEntryExpansion,
   renderSelectionCount,
   mode,
   selection,
   tableRowHeight,
 }) {
-  if (renderEntryChild && onSelectEntries) {
-    throw new Error(
-      'A DataView cannot be made selectable if its entries have children. ' +
-        'Please use only one of renderEntryChild or onSelectEntries at a time.'
+  if (renderEntryChild && !renderEntryExpansion) {
+    warnOnce(
+      'DataView:renderEntryChild',
+      'DataView: the renderEntryChild prop has been renamed ' +
+        'renderEntryExpansion and will be removed in a future version. ' +
+        'Please use the new name.'
     )
+    renderEntryExpansion = renderEntryChild
+  }
+
+  if (renderEntryExpansion && onSelectEntries) {
+    warnOnce(
+      'DataView: renderEntryExpansion + onSelectEntries',
+      'A DataView cannot be made selectable with some of its entries ' +
+        'expandable. Please use only one of renderEntryExpansion or ' +
+        'onSelectEntries at a time. Ignoring renderEntryExpansion.'
+    )
+    renderEntryExpansion = undefined
   }
 
   // Only used if `page` is not passed. The pagination supports both a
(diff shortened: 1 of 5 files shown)


src/components/DataView/DataView.js
```diff
@@ -35,13 +35,6 @@ function prepareFields(fields) {
   })
 }
 
-function entryChildrenFromChild(child) {
```
sohkai: It seems like we're doing this transformation more often in the children, would it make sense to move it back here?

bpierre: I’m not sure I understand this one, we are doing about the same thing than before?

bpierre:
> Perhaps we may also want to note this in the docs, that if an entry renders a single child, it can have variable width (but not if it renders more than one).

Done in https://github.com/aragon/aragon-ui/pull/535/commits/2a1f682bd12c57842902f00efe555151db1eb118

src/components/DataView/DataView.js
sohkai: We are, but now we duplicate this array wrapping in both the lower components when we could do it here and always send down an array.

bpierre: Oh right, you are suggesting to move the “single node mode” check here. Yes it makes sense, let’s do this.

bpierre: Done in https://github.com/aragon/aragon-ui/pull/535/commits/9a827e848a68a4251e0c7dc0e0a66ae27030560f

src/components/DataView/DataView.js
```diff
@@ -35,6 +35,25 @@ function prepareFields(fields) {
   })
 }
 
+function entryChildrenFromChild(child) {
+  if (!child) {
+    return {
+      children: null,
```
sohkai: What if we just returned an empty array here? Then we can just check in the child components if it's a single child by doing an array length check

src/components/DataView/DataView.js
```diff
@@ -35,6 +35,25 @@ function prepareFields(fields) {
...
+      singleNodeChildren: false,
+    }
+  }
+  if (Array.isArray(child)) {
+    return {
+      children: child,
+      singleNodeChildren: false,
```
sohkai: If we do keep `singleNodeChildren`, we should do a length check on child.

src/components/DataView/DataView.js
bpierre: I guess we could do that for the empty child, checking `const isEmpty = Array.isArray(children) && children.length === 0`, but it would be more or less the same than we were doing before? Also for the single child, there would be no way to know if it’s an actual “single child” (a React element with custom content) or an array of rows containing only one row.

src/components/DataView/DataView.js
bpierre: Oh yes, you mean to also make empty arrays behave the same than `null`? But I don’t see how it relates to `singleNodeChildren` here?

src/components/DataView/DataView.js
sohkai: I saw that we generally use the `children` via either `children.length` or `children.map()` (e.g. in [`ListView`](https://github.com/aragon/aragon-ui/blob/9a827e848a68a4251e0c7dc0e0a66ae27030560f/src/components/DataView/ListView.js#L200)) and so thought an empty array would behave the same as logic checking that it wasn't an array.

src/components/DataView/DataView.js
sohkai: If `renderEntryChild()` returned an array but only one child, would we consider it a single node?

bpierre: Nope, it would be a “single row” (taking the height of the table row above, and having its content aligned on one of the cells). What I called “single node children” is when a React node is returned, and the layout is then totally free.

These names are only internal but they are probably quite misleading yeah… I’m opened to suggestions if you have any idea. I also dislike the term “entry child” that is exposed in the API, since children are already a React concept, but I couldn’t come with anything better yet.

src/components/DataView/DataView.js
bpierre: Oh right, I thought it was about returning the array directly (rather than an object also containing `singleNodeChildren`. `EntryChildren` is not rendered when there is no child content, but I guess we could return empty arrays for consistency yes, it would remove the `Array.isArray` check.

src/components/DataView/DataView.js
sohkai: Ahh I see what you mean now. Yeah, the mode names are a bit confusing...

Perhaps `renderEntryExpansion()`? And then we can use `singleExpansionNode` for the mode name?

bpierre: That’s much better :pray:

bpierre: Done :point_right: https://github.com/aragon/aragon-ui/pull/535/commits/92914a6385990128662bdc61c793fdc05f606e64

I ended up using the term “free layout” internally, since that’s how I described it in the comment above haha.

sohkai: Yes! I was thinking about "free" vs. "controlled" mode but "controlled" is also a confusing term that React has used already haha

src/components/DataView/README.md
```diff
@@ -260,17 +255,25 @@ function App() {
...
+It should return one of these:
 
-The child content can also be aligned on a specific column in table mode, see `field.childStart`.
+- A non empty array, to be displayed as individual rows.
+- A React node, to remove any specific layout constraint.
+- `null` or an empty array, to make the entry not expandable.
+
+When a non-empty array is returned, each of its entries will be displayed as a row, whose height is determined by the value of `tableRowHeight` − even in list view mode. The alignment of these rows can also start from a specific column in table mode: see `field.childStart`.
+
+When a React node is returned there is no layout constraints anymore, and the height depends on the content.
```
sohkai:
```suggestion
When a React node is returned, there are no layout constraints anymore and the expansion's height depends on the returned content.
```

src/components/DataView/ListView.js
```diff
@@ -30,7 +31,9 @@ function ListView({
   return (
     <React.Fragment>
       {entries.map((entry, index) => {
-        const hasChildren = entry.children.length > 0
+        const hasChildren = Array.isArray(entry.children)
```
AquiGorka: I think there is a native React method that can be used here: https://reactjs.org/docs/react-api.html#reactchildrencount

src/components/DataView/ListView.js
```diff
@@ -180,11 +147,78 @@ ListView.propTypes = {
...
+  rowHeight: PropTypes.number.isRequired,
   selectable: PropTypes.bool.isRequired,
 }
 
+// Disable prop types check for internal components
 /* eslint-disable react/prop-types */
 
+function EntryChildren({ entry, opened, rowHeight }) {
+  const theme = useTheme()
+  const singleNode = !Array.isArray(entry.children)
```
AquiGorka: Same as comment from above on `React.children` to use those methods to jump over the extra data added to children and figure out if it is an array or not.

src/components/DataView/TableView.js
```diff
@@ -42,7 +42,9 @@ function rowFromEntry(
     toggleChildContent,
   }
 ) {
-  const hasChildren = entry.children.length > 0
+  const hasChildren = Array.isArray(entry.children)
+    ? entry.children.length > 0
```
AquiGorka: `React.children.count` definitely here

src/components/DataView/TableView.js
```diff
@@ -358,21 +373,17 @@ function EntryChildren({
...
-                  will-change: height;
-                `}
-                style={{
-                  height: totalHeight.interpolate(v => `${v}px`),
-                }}
+                css="overflow: hidden"
+                style={{ height: totalHeight }}
               >
-                {entry.children.map((child, i) => (
+                {children.map((child, i) => (
```
AquiGorka: `React.children.map`

src/components/DataView/ListView.js
bpierre: They just just normal arrays, not children in the React sense (that confirms why it was a good idea [to rename these](https://github.com/aragon/aragon-ui/pull/535/commits/92914a6385990128662bdc61c793fdc05f606e64) :smile: )

AquiGorka: There you go 😄

sembrestels:
Thank you so much @bpierre! It's already in use in https://github.com/1Hive/conviction-voting/commit/5894f9cdd2877a1f03ad1bb0557f619c9e6cf61a :wink:

PR #536, opened 2019-08-27
Move LinkBase features into ButtonBase

bpierre:
`LinkBase` is now deprecated, and `ButtonBase` handles its different cases. `Button`, `ButtonText`, and `Link` have been updated accordingly.

The `external` prop of both `Link` and `ButtonBase` defaults to `true` if `href` is present, `false` otherwise.

Also:

- Add documentation for `ButtonBase`.
- In the gallery, “providers” is now “advanced”, since it contains different things (providers, HOC functions, hooks, render props components).


src/components/Link/Link.js
```diff
@@ -1,42 +1,37 @@
...
       href={href}
       onClick={onClick}
-      onKeyDown={handleKeyDown}
+      external={external}
+      focusRingSpacing={6}
       {...props}
       css={`
         color: ${theme.link};
-        text-decoration: none;
+        text-decoration: ${external ? 'underline' : 'none'};
```
sohkai: More of a note for @dizzypaty, but we should probably note this behaviour in the design guidelines somewhere?

src/components/ButtonBase/README.md
```diff
@@ -0,0 +1,79 @@
...
+
+## Props
+
+### disabled
+
+| Type      | Default value |
+| --------- | ------------- |
+| `Boolean` | false         |
+
+When set to true, the `disabled` prop will be passed to the button element. And `onClick` will get ignored. If an `href` is present, it won’t get passed to the `<a>` element underneath.
```
sohkai:
```suggestion
When set to true, the `disabled` prop will be passed to the button element. And `onClick` will get ignored. If an `href` is present and `disabled` is true, the `href` won’t get passed to the `<a>` element underneath.
```

src/components/Link/Link.js
dizzypaty: Yes, I am aware! We discussed the rational and made the decision together with Pierre.

PR #540, opened 2019-08-28
Card: unset text-align when interactive

sohkai:
User agents normally apply `text-align: center` to buttons, and we don't override this behaviour.

It's a bit odd for Card to changes its text alignment when being interactive.


bpierre:
Closing as https://github.com/aragon/aragon-ui/pull/541 addressed this issue.

PR #541, opened 2019-08-28
Card: fix the layout differences in interactive mode

bpierre:
Move to an inner `<div>` element to avoid any layout specificity with `<button>`.

### DIFF
--- a/src/components/ButtonBase/ButtonBase.js
+++ b/src/components/ButtonBase/ButtonBase.js
@@ -1,14 +1,13 @@
-import React from 'react'
+import React, { useCallback } from 'react'
 import PropTypes from 'prop-types'
 import FocusVisible from '../FocusVisible/FocusVisible'
 import { useTheme } from '../../theme'
 import { RADIUS, textStyle } from '../../style'
-import { warnOnce } from '../../utils'
-
-const ANCHOR_EXTERNAL_PROPS = { rel: 'noopener noreferrer', target: '_blank' }
+import { warnOnce, KEY_ENTER } from '../../utils'
 
 function ButtonBase({
   disabled,
+  element,
   external,
   focusRingRadius,
   focusRingSpacing,
@@ -16,6 +15,7 @@ function ButtonBase({
   href,
   innerRef,
   onClick,
+  onKeyDown,
   showFocusRing,
   ...props
 }) {
@@ -26,21 +26,55 @@ function ButtonBase({
     external = Boolean(href)
   }
 
-  // element-specific props
-  const elementProps = href
-    ? // <a href>
-      {
-        as: 'a',
-        href: disabled ? undefined : href,
-        ...(external ? ANCHOR_EXTERNAL_PROPS : {}),
+  if (!element) {
+    element = href ? 'a' : 'button'
+  }
+
+  const elementProps = { as: element }
+
+  // <button> (handle key events)
+  if (element === 'button') {
+    elementProps.type = 'button'
+    elementProps.disabled = disabled
+  }
+
+  // <a href=""> (handle key events)
+  if (element === 'a' && href && !disabled) {
+    elementProps.href = href
+    elementProps.rel = 'noopener noreferrer'
+    elementProps.target = '_blank'
+  }
+
+  // <a> or <div> (doesn’t handle key events)
+  if ((element === 'a' && !href) || element === 'div') {
+    elementProps.role = 'button'
+    elementProps.tabIndex = '0'
+  }
+
+  const handleKeyDown = useCallback(
+    event => {
+      // Only applies to cases where the enter key is not handled already
+      if (
+        ((element === 'a' && !href) || element === 'div') &&
+        event.keyCode === KEY_ENTER &&
+        onClick
+      ) {
+        onClick()
       }
-    : // <button>
-      { as: 'button', type: 'button', disabled }
+
+      // Pass the event up
+      if (onKeyDown) {
+        onKeyDown(event)
+      }
+    },
+    [onClick, element]
+  )
 
   return (
     <button
       ref={innerRef}
       onClick={disabled ? undefined : onClick}
+      onKeyDown={disabled ? undefined : handleKeyDown}
       {...elementProps}
       {...props}
       css={`
@@ -87,7 +121,9 @@ ButtonBase.propTypes = {
   href: PropTypes.string,
   innerRef: PropTypes.any,
   onClick: PropTypes.func,
+  onKeyDown: PropTypes.func,
   showFocusRing: PropTypes.bool,
+  element: PropTypes.oneOf(['button', 'div', 'a']),
 }
 
 ButtonBase.defaultProps = {
--- a/src/components/ButtonBase/ButtonBase.js
+++ b/src/components/ButtonBase/ButtonBase.js
@@ -5,6 +5,42 @@ import { useTheme } from '../../theme'
 import { RADIUS, textStyle } from '../../style'
 import { warnOnce, KEY_ENTER } from '../../utils'
 
+function getElementProps({ element, href, disabled }) {
+  // <button> (handles key events)
+  if (element === 'button') {
+    return [
+      'button',
+      {
+        type: 'button',
+        disabled,
+      },
+    ]
+  }
+
+  // <a href=""> (handles key events)
+  if (element === 'a' && href) {
+    return [
+      'anchor',
+      disabled
+        ? {}
+        : {
+            href: href,
+            rel: 'noopener noreferrer',
+            target: '_blank',
+          },
+    ]
+  }
+
+  // <a> or <div> (doesn’t handle key events)
+  return [
+    'basic',
+    {
+      role: 'button',
+      tabIndex: disabled ? '-1' : '0',
+    },
+  ]
+}
+
 function ButtonBase({
   disabled,
   element,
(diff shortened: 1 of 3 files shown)


src/components/Card/Card.js
```diff
@@ -50,6 +50,7 @@ function Card({ children, width, height, onClick, ...props }) {
             transition-property: top, box-shadow;
             transition-duration: 50ms;
             transition-timing-function: ease-in-out;
+            white-space: normal;
```
sohkai: Ahh, that's what was causing https://github.com/aragon/aragon-apps/pull/954#discussion_r318561939.

src/components/Card/Card.js
```diff
@@ -60,7 +61,18 @@ function Card({ children, width, height, onClick, ...props }) {
       `}
       {...props}
     >
-      {children}
+      <div
+        css={`
+          position: relative;
+          display: flex;
+          align-items: center;
```
sohkai: Should this have `flex-direction: column` set by default?

src/components/Card/Card.js
```diff
@@ -60,7 +61,18 @@ function Card({ children, width, height, onClick, ...props }) {
       `}
       {...props}
```
sohkai: This does mean that we'll have to migrate any custom styling of Card to an inner child, now that we're not directly styling the parent.

src/components/Card/Card.js
bpierre: Yes definitely!

src/components/Card/Card.js
sohkai: We still need to reset `text-align`, the inner div doesn't reset it.

src/components/Card/Card.js
bpierre: Oh yes, I didn’t think about styling the inner card. OK, I think it’s enough arguments to move back to a `div` :smile:

src/components/Card/Card.js
AquiGorka: Wait, I remember suggesting this, but can't recall why you said it could not work.

bpierre: I think your suggestion was to remove the `display: flex` entirely (and I’d like to keep it centered by default).

AquiGorka: That too, but the first way I fixed this was by setting the direction which is why I asked, then a whole bunch of other problems spawned, but anyhow, div it is.

bpierre:
I added a new `element` prop on `ButtonBase`, that can be used to force the element to be set, if really needed. I wonder if it’s worth having a `<button>` mode at all in `ButtonBase`, now that we handle the keyboard activation for the `<div>` mode.

src/components/Card/Card.js
```diff
@@ -25,35 +25,43 @@ function Card({ children, width, height, onClick, ...props }) {
...
-            border: 0;
             box-shadow: 0px 1px 3px rgba(51, 77, 117, 0.15);
-            transition-property: top, box-shadow;
+            transition-property: transform, box-shadow;
             transition-duration: 50ms;
             transition-timing-function: ease-in-out;
-            ${unselectable};
+
+            // Reset some styles applied by ButtonBase
+            text-align: left;
```
sohkai: We might not need to reset `text-align` anymore if we're using a div

src/components/ButtonBase/ButtonBase.js
```diff
@@ -26,21 +26,55 @@ function ButtonBase({
...
+  if ((element === 'a' && !href) || element === 'div') {
+    elementProps.role = 'button'
+    elementProps.tabIndex = '0'
+  }
+
+  const handleKeyDown = useCallback(
+    event => {
+      // Only applies to cases where the enter key is not handled already
+      if (
+        ((element === 'a' && !href) || element === 'div') &&
```
sohkai: Perhaps we can lift this condition to a variable

src/components/Card/Card.js
```diff
@@ -25,35 +25,43 @@ function Card({ children, width, height, onClick, ...props }) {
...
 
         ${interactive &&
           css`
-            border: 0;
             box-shadow: 0px 1px 3px rgba(51, 77, 117, 0.15);
-            transition-property: top, box-shadow;
+            transition-property: transform, box-shadow;
             transition-duration: 50ms;
             transition-timing-function: ease-in-out;
-            ${unselectable};
```
sohkai: Maybe we can keep this? Or are users supposed to apply this themselves now?

bpierre:
> LGTM, it's nice that we also fixed the tabIndex for just `<a>` but this is getting a little bit overengineered as something that was meant to be used for buttons.

I’m not sure I follow about `tabIndex`: we need `tabIndex=0` if we want it to be focusable?

When you say overengineered, do you mean that you would remove the `div` mode from `BaseButton`, and move the focus / keyboard handling part back in `Card` instead of using `BaseButton`? That could work too, I think both approaches are valid. I ended up adding a way to use a `<div>` in `BaseButton` because I anticipate other cases of large buttons where a layout inside of `BaseButton` would be needed, and the way `<button>` behaves in CSS is too problematic for that.

> Perhaps we should rename this InteractiveBase or something like that, since it's really starting to be more than just a button :shrug:

Yes I almost went for this (or `InteractiveSurface` more exactly), but I felt it was a bit too abstract (and I want to avoid the term “surface” that we use for something else already).

I like that “button base”, while maybe not being strictly correct, is easy to understand as “a clickable area mostly used for buttons”, which can take many different shapes: a `Button`, a tab, a `Link`, the handle of a `Slider`, an interactive `Card`… but it should also be ok to not use it when it doesn’t fit well.

`InteractiveBase` is better than `InteractiveSurface`, but the term “interactive” is still a bit too broad IMO: we have many components that won’t use it, but are interactive in other ways. iOS is having something a bit similar to it, that they call [`UIControl`](https://developer.apple.com/documentation/uikit/uicontrol), I guess that could work too… but it would mean making it work with `Checkbox`, `Radio` etc. and I think it’s a bit too early for that.

Maybe we could keep the name `BaseButton` until it really becomes generic? For now I don’t think it is really the case yet, I think interactive cards are looking a lot like something we could describe as “big buttons” :smile:

src/components/Card/Card.js
bpierre: We have better in `ButtonBase` now :-) https://github.com/aragon/aragon-ui/blob/newstyle/src/components/ButtonBase/ButtonBase.js#L52

It was added for `Link`, but it is useful in other components as well: it keeps them non selectable when being interacted with, but their content can still get selected when they are in the middle of a selection (e.g. in the middle of a paragraph) − the best of both worlds. This is basically how `<a href="">` behave since forever in browsers, but now we can do it in CSS too.

bpierre: …and now I am confused, because I can’t reproduce it anymore. `user-select: text` selects elements when double clicking on them, which is something we don’t want. I’ll move back to `user-select: none` for now (which means that `Link` buttons won’t be selectable :disappointed: ).

src/components/ButtonBase/ButtonBase.js
bpierre: I moved it all in its own function https://github.com/aragon/aragon-ui/pull/541/commits/43a9f7d263618b7b660ee331557cfcb4e94df9c0

src/components/Card/Card.js
bpierre: Ah yes I forgot to do it in `ButtonBase`, but I wanted to make it explicit since it it doesn’t only use `<button>`.

bpierre: Done in https://github.com/aragon/aragon-ui/pull/541/commits/d349eaefc853accf8bf240dfabaab7c6617b7e25

sohkai:
> Maybe we could keep the name BaseButton until it really becomes generic? For now I don’t think it is really the case yet, I think interactive cards are looking a lot like something we could describe as “big buttons”

Let's do this 👍.

> I’m not sure I follow about tabIndex: we need tabIndex=0 if we want it to be focusable?

Yes! I don't think we had the `tabIndex` on `<a>`s before, and it's good we fixed the "button" mode for `<a>`s.

> When you say overengineered, do you mean that you would remove the div mode from BaseButton, and move the focus / keyboard handling part back in Card instead of using BaseButton?

It was mostly about how I perceive ButtonBase; it's at the point now where it feels too generic to be really a "button" but not generic / re-usable enough to be used elsewhere. Ideally I would've liked an abstraction that would allow components to "mixin" the interactive handling and styling (favouring composition over the current component pseudo-inheritance), but perhaps we're just handicapped by the APIs we're given.

Extrapolating ahead, we'll probably end up adding more and more elements to the list (and then we might even ask, "what about components?") and slowly ButtonBase will become unwieldy to figure out how / when to use. But maybe that's just my anxiety.

src/components/Card/Card.js
sohkai: Perhaps this is something we can override in `Link` (since usually we want inline text to be selectable)? What do you think about only applying `user-select: none` when ButtonText isn't rendering an `<a>`?

bpierre:
> Yes! I don't think we had the tabIndex on <a>s before, and it's good we fixed the "button" mode for <a>s.

Oh right! I don’t think it was possible to use it with a `<a>` without a `href` before `element`: it was determined with the presence of `href` (which makes it work with the keyboard automatically).

> It was mostly about how I perceive ButtonBase; it's at the point now where it feels too generic to be really a "button" but not generic / re-usable enough to be used elsewhere.

Yes that’s an accurate description haha.

> Ideally I would've liked an abstraction that would allow components to "mixin" the interactive handling and styling (favouring composition over the current component pseudo-inheritance), but perhaps we're just handicapped by the APIs we're given.

Yes I’d totally agree with something like that, maybe we could try to do it progressively? But yes, there are so many little things to take care of with HTML / DOM that I’m not sure how manageable that would be − like how `<button>` behave weirdly when used as a Flex or Grid container.

> Extrapolating ahead, we'll probably end up adding more and more elements to the list (and then we might even ask, "what about components?") and slowly ButtonBase will become unwieldy to figure out how / when to use. But maybe that's just my anxiety.

I can’t see any reason to add any other element than the basic `button`, `a` or `div` here, actually we could even get rid of `div` as an `a` without `href` also acts as a “neutral HTML element”. It’s really about having a neutral button that can also benefit from some HTML features when possible.

src/components/Card/Card.js
bpierre: Yes we could at least make the text selectable when it is a normal link, let’s do this for now. :+1:

bpierre: Yes let’s do it for anchors at least since they behave properly, but I’d like to avoid having the other buttons selecting themselves when being clicked.

PR #544, opened 2019-08-29
Box: update padding behavior

bpierre:
As seen with @dizzypaty, we’ll keep a padding of 2gu all around in full width mode: any extra spacing can be added, but the line height of our text styles is enough in most cases to make it like the design screens.

- ~It is now possible to provide the header and content padding separately, by using an array.~
- Fix the default values (always 2gu in full width).
- ~When padding is 0 or false, the header stop following the content padding (it is still possible to do this by setting it independently using the array).~
- The header stop following the content padding.
- Coding style: remove the internal HeaderContent component.

### DIFF
--- a/src/components/Box/Box.js
+++ b/src/components/Box/Box.js
@@ -5,11 +5,23 @@ import { useTheme } from '../../theme/Theme'
 import { useLayout } from '../Layout/Layout'
 import { useInside } from '../../utils'
 
-function getPaddingValue(padding, insideSplitPrimary) {
-  if (typeof padding === 'boolean') {
-    return padding ? (insideSplitPrimary ? 5 : 2) * GU : 0
+// Padding values for the heading and the content.
+function getPaddingValues(padding, { insideSplitPrimary, fullWidth }) {
+  const defaultPadding = (fullWidth ? 2 : insideSplitPrimary ? 5 : 3) * GU
+
+  // Separate values for the header and content
+  if (Array.isArray(padding)) {
+    return padding
   }
-  return padding
+
+  // Default value if true, disable padding on the content otherwise
+  if (typeof padding === 'boolean') {
+    return [defaultPadding, padding ? defaultPadding : 0]
+  }
+
+  // The heading follows the content padding except when 0,
+  // in which case it will follow the default value.
+  return [padding === 0 ? defaultPadding : padding, padding]
 }
 
 function Box({ heading, children, padding, ...props }) {
@@ -18,7 +30,10 @@ function Box({ heading, children, padding, ...props }) {
   const { layoutName } = useLayout()
   const fullWidth = layoutName === 'small'
 
-  const paddingValue = getPaddingValue(padding, insideSplitPrimary)
+  const [headerPadding, contentPadding] = getPaddingValues(padding, {
+    insideSplitPrimary,
+    fullWidth,
+  })
 
   return (
     <div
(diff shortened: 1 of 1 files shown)


src/components/Box/Box.js
```diff
@@ -5,11 +5,30 @@ import { useTheme } from '../../theme/Theme'
...
-function getPaddingValue(padding, insideSplitPrimary) {
+// Padding values for the heading and the content.
+function getPaddingValues(padding, { insideSplitPrimary, fullWidth }) {
+  if (Array.isArray(padding)) {
+    // User declared separate values for the header and content;
+    // just use their values
+    return padding
+  }
+
+  const defaultPadding = (fullWidth ? 2 : insideSplitPrimary ? 5 : 3) * GU
```
AquiGorka: What about mobile? Is this consistent with mobile spacing too? What about items that go to the full edge of the screen?

bpierre:
@sohkai I started these changes because an author couldn’t remove the padding on the content without also removing it on the header, which is problematic IMO. The independent paddings can be considered as an advanced feature, and shouldn’t be needed in most cases, but it ensures it can be achieved.

Now I am wondering if we shouldn’t use an object syntax, and only use the two-entries array syntax when we need want to define separate horizontal and vertical spacings:

```jsx
<Box
  padding={{ heading: 3 * GU, content: 3 * GU }}
/>
```

src/components/Box/Box.js
bpierre: This is `fullWidth` :-)

AquiGorka: That's why I ask, if `fullWidth` then there's a `2px` padding, which means no bleeding edge no?

bpierre: It’s `2 * GU` => `16px`, which is correct I think?

AquiGorka: True `GU` not `px`, anyway I may have been confused, I was talking about the outer margin instead of the inner padding.

sohkai:
> Now I am wondering if we shouldn’t use an object syntax, and only use the two-entries array syntax when we need want to define separate horizontal and vertical spacings:

Agreed! We could also `'default'` to specify it should follow the default behaviour.

sohkai:
@bpierre @dizzypaty This is the only misalignment I could find with these new paddings (save for `DataView` woes):

Screen Shot 2019-08-29 at 3 19 27 PM

The designs for voting have the Box at 3GU padding on fullwidth when it's in the primary slot of a split, and hence the back button looks misaligned.

But from what I understand, we should be changing the BackButton so that it has 2GU when in fullwidth?

bpierre:
> But from what I understand, we should be changing the BackButton so that it has 2GU when in fullwidth?

We might want 3gu - 4px (so 20px), to make the icon glyph aligned on 3gu (icons have more or less 4px of horizontal spacing).

bpierre:
@sohkai done https://github.com/aragon/aragon-ui/pull/544/commits/674c4ac64a39c1b83bb1497e380f850eddb08bd4

I didn’t use `"default"` in the end because we are already using `true` for that.

src/components/Box/Box.js
```diff
@@ -48,28 +48,19 @@ function Box({ heading, children, padding, ...props }) {
...
-        <div
+        <h1
           css={`
             display: flex;
             align-items: center;
             height: ${4 * GU}px;
+            padding: 0 ${defaultPadding}px;
             border-bottom: 1px solid ${theme.border};
+            color: ${theme.surfaceContentSecondary};
+            ${textStyle('label2')};
```
sohkai: Small things:

- Should we only apply this if the heading is a string?
- We need to set this heading to have full width so things like the [repo versions](https://www.figma.com/file/J2dUh6DtHJileEjmuZxzgBU1/App-center?node-id=2%3A0) header works.

bpierre: It should have both of these things? If the heading wasn’t full width, we wouldn’t have the border.

bpierre: And `heading` can be any node, string or not.

sohkai: > It should have both of these things? If the heading wasn’t full width, we wouldn’t have the border.

Hmm yes, good point!

> And heading can be any node, string or not.

But would it be weird if this style was automatically applied to nodes you passed in (as opposed to a string)?

bpierre: > But would it be weird if this style was automatically applied to nodes you passed in (as opposed to a string)?

It should be the case already, these styles apply to anything inside?

So we can do this:

```jsx
<Box
  heading={
    <>
      My title
      <Help />
    </>
  }
/>
```

bpierre: Added a comment https://github.com/aragon/aragon-ui/pull/544/commits/93da71f2f76a8f7974a52e79ab733efb87d0a24d

PR #547, opened 2019-08-29
Popover: use nearest root as boundary

sohkai:
Sets the Popover's boundary to be the nearest Root element, defaulting back to `window` otherwise.

This avoids awkward cases where a Popover would not consider the nearest Root when calculating its position, leading to cases like

Screen Shot 2019-08-29 at 5 31 28 PM

-------

Also exposes `useRoot()` for convenience.


bpierre:
Ah no, it’s only because of the demo container:



All good! :smile:

sohkai:
Yeah... I couldn't figure out how to make the demo container nicer and didn't want to spend too much time on it haha.

PR #554, opened 2019-08-30
DataView improvements

bpierre:
- `<Box padding={false}>` => `<Box padding={0}>`
- Skip extra renders in table mode.
- Allow empty header in table mode.
- Allow to disable the pagination.

### DIFF
--- a/src/components/DataView/TableView.js
+++ b/src/components/DataView/TableView.js
@@ -8,8 +8,8 @@ import { Checkbox } from '../Input/Checkbox'
 import { ToggleButton } from './ToggleButton'
 import { OpenedSurfaceBorder } from './OpenedSurfaceBorder'
 
-// Table head row
-function cellsFromFields(
+// Table heading cells
+function headingsFromFields(
   fields,
   { hasAnyActions, hasAnyExpansion, selectContent, selectable }
 ) {
@@ -92,7 +92,7 @@ function TableView({
 
   const headCells = useMemo(
     () =>
-      cellsFromFields(fields, {
+      headingsFromFields(fields, {
         hasAnyActions,
         hasAnyExpansion,
         selectContent: (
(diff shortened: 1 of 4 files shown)


src/components/DataView/TableView.js
```diff
@@ -26,24 +26,26 @@ function cellsFromFields(
     cells.push([null, 'left'])
   }
 
+  // Return null if all the fields are empty
+  if (cells.every(cell => !cell[0])) {
+    return null
+  }
+
   return cells
 }
```
sohkai: Perhaps this function could be renamed to `headingsFromFields` or similar?

src/components/DataView/TableView.js
```diff
@@ -160,31 +119,30 @@ function TableView({
...
+            entry={entry}
+            fields={fields}
+            firstRow={index === 0}
+            hasAnyActions={hasAnyActions}
+            hasAnyExpansion={hasAnyExpansion}
+            onToggle={toggleEntry}
+            opened={opened === entry.index}
+            rowHeight={rowHeight}
+            selectable={selectable}
+            alignChildOnField={alignChildOnField}
```
sohkai: Can we move `alignChildOnField` to the top?

src/components/DataView/TableView.js
```diff
@@ -160,31 +119,30 @@ function TableView({
...
+        {entries.map((entry, index) => (
+          <Entry
+            key={entry.index}
+            entry={entry}
+            fields={fields}
+            firstRow={index === 0}
+            hasAnyActions={hasAnyActions}
+            hasAnyExpansion={hasAnyExpansion}
+            onToggle={toggleEntry}
+            opened={opened === entry.index}
```
sohkai: Should we passing down an `onSelect()` as well?

src/components/DataView/TableView.js
bpierre: Yes!

src/components/DataView/TableView.js
bpierre: Oh yes :man_facepalming:

PR #555, opened 2019-08-30
New component: Accordion

bpierre:
Builds on https://github.com/aragon/aragon-ui/pull/554

Also:

- Fix `createRandomInt(seed)` in devbox, that was not actually using the seed.
- In components/index, move back the experimental components at the end.
- Fix an issue where “compact cells” (the ones containing the checkbox, the expand buttons, or the actions) where getting larger than their content on Chrome.


bpierre:
@sohkai We could make it a render function maybe, but they are not actually mounted until we open a single expand, so it shouldn’t be a big issue anyway.

PR #556, opened 2019-09-02
chore: add react-hooks ESLint plugin

sohkai:
Thought we already had this enabled in this repo D:.

Caught a few bugs, but a few of them I'm not 100% sure. @bpierre can you double check? I don't think any of them should cause regressions, since they're mostly based on props that don't change.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -20,13 +20,16 @@
     },
     "sourceType": "module"
   },
-  "plugins": ["prettier", "react", "import", "promise", "jest"],
+  "plugins": ["prettier", "react", "react-hooks", "import", "promise", "jest"],
   "rules": {
+    "react/jsx-boolean-value": "off",
+    "react-hooks/rules-of-hooks": "error",
+    "react-hooks/exhaustive-deps": "warn",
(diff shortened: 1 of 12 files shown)


.eslintrc
```diff
@@ -20,13 +20,16 @@
     },
     "sourceType": "module"
   },
-  "plugins": ["prettier", "react", "import", "promise", "jest"],
+  "plugins": ["prettier", "react", "react-hooks", "import", "promise", "jest"],
   "rules": {
+    "react/jsx-boolean-value": "off",
```
bpierre: I’m starting to think that the `true` shortcut looks a bit like a historical mistake in JSX. It’s really weird to do `<Component prop />` to set `prop` to `true` and `<Component prop={false} />` to set it to false.

I also noticed that the React documentation started to recommend not using it:




Nothing urgent but if we agree on this, maybe we could start using `"always"` for this rule.

sohkai: Let's do this when we upgrade our configs altogether :).

It is a good note though, and I do actually like the shorthand.

PR #557, opened 2019-09-02
Update AddressField, IdentityBadge, TokenBadge, and TransactionBadge to new styles

sohkai:
As seen in [Figma](https://www.figma.com/file/bZ1j7omfhdMCbtFahP3nEJ9B/Client?node-id=598%3A910).

AddressBadge:

Screen Shot 2019-09-02 at 3 07 08 PM

IdentityBadge:

Screen Shot 2019-09-02 at 3 07 31 PM

Screen Shot 2019-09-02 at 3 07 36 PM

TokenBadge:

Screen Shot 2019-09-02 at 3 06 59 PM

Screen Shot 2019-09-02 at 3 07 02 PM

### DIFF
--- a/src/components/AddressField/AddressField.js
+++ b/src/components/AddressField/AddressField.js
@@ -1,9 +1,9 @@
 import React, { useCallback, useEffect, useRef } from 'react'
 import PropTypes from 'prop-types'
 import { IconCopy } from '../../icons/'
-import { GU, RADIUS } from '../../style'
+import { GU, RADIUS, textStyle } from '../../style'
 import { useTheme } from '../../theme'
-import { monospace, noop, warn } from '../../utils'
+import { noop, warn } from '../../utils'
 import TextInput from '../Input/TextInput'
 import { ButtonIcon } from '../Button/ButtonIcon'
 import EthIdenticon from '../EthIdenticon/EthIdenticon'
@@ -113,7 +113,7 @@ const AddressFieldBase = React.memo(function AddressFieldBase({
           max-width: 100%;
           border: 0;
           ${icon ? `padding-left: ${1 * GU}px` : ''};
-          ${monospace};
+          ${textStyle('address2')};
           &:read-only {
             color: ${theme.surfaceContent};
             text-shadow: none;
--- a/src/components/AddressField/AddressField.js
+++ b/src/components/AddressField/AddressField.js
@@ -1,124 +1,130 @@
-import React from 'react'
+import React, { useCallback, useEffect, useRef } from 'react'
 import PropTypes from 'prop-types'
-import { IconCopy } from '../../icons/components'
+import { IconCopy } from '../../icons/'
+import { GU, RADIUS } from '../../style'
+import { useTheme } from '../../theme'
+import { monospace, noop, warn } from '../../utils'
 import TextInput from '../Input/TextInput'
 import { ButtonIcon } from '../Button/ButtonIcon'
 import EthIdenticon from '../EthIdenticon/EthIdenticon'
 import { Toast } from '../ToastHub/ToastHub'
-import { theme } from '../../theme-legacy'
-import { font, noop, warn } from '../../utils'
 
-class AddressFieldBase extends React.PureComponent {
-  static propTypes = {
-    address: PropTypes.string.isRequired,
-    onCopy: PropTypes.func,
-    icon: PropTypes.node,
-  }
-  _input = React.createRef()
-  componentDidMount() {
-    setTimeout(() => {
-      this._input.current.focus()
-    }, 0)
-  }
-  handleFocus = () => {
-    this._input.current.select()
-  }
-  handleCopy = () => {
-    const { onCopy } = this.props
-    this._input.current.focus()
-    this._input.current.select()
+const HEIGHT_ADJUSTED_FOR_BORDER = 5 * GU - 2
 
-    try {
-      document.execCommand('copy')
-      onCopy('Address copied')
-    } catch (err) {
-      warn(err)
+const AddressFieldBase = React.memo(function AddressFieldBase({
+  address,
+  icon,
+  onCopy,
+}) {
+  const theme = useTheme()
+  const inputRef = useRef(null)
+
+  const handleFocus = useCallback(() => {
+    inputRef.current && inputRef.current.select()
+  }, [])
+  const handleCopy = useCallback(() => {
+    if (inputRef.current) {
+      inputRef.current.focus()
+      inputRef.current.select()
+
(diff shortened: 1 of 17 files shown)


src/theme/theme-light.js
```diff
@@ -50,8 +50,7 @@ export default {
   positiveSurface: colors.MintGreenLight,
   positiveSurfaceContent: colors.MintGreenDark,
 
-  badge: colors.Blue,
-  badgeContent: colors.Blue,
+  badge: colors.AragonBlueLighter,
```
sohkai: As discussed offline with @dizzypaty, we weren't using `theme.badge` anymore anyway, and we took over this colour with the new background for the identity badges.

src/components/IdentityBadge/IdentityBadge.js
```diff
@@ -1,136 +1,155 @@
...
+              white-space: nowrap;
+              text-overflow: ellipsis;
               overflow: hidden;
-              display: flex;
-              align-items: center;
-              ${compact
-                ? 'padding-left: 8px; border-radius: 2px;'
-                : 'background: #daeaef; border-radius: 3px;'};
-              text-decoration: none;
+              ${textStyle('body2')}
```
sohkai: @dizzypaty This how an app would look with the current 16px font-size:

Screen Shot 2019-09-02 at 3 11 13 PM

And how it would look with the previous 15px:

Screen Shot 2019-09-02 at 3 12 05 PM

src/theme/theme-light.js
bpierre: We should still define and use `badgeContent` for the text on `badge`. Also I’m not sure what the decision is for the pressed state, but if we don’t want to add a shadow on this component, then it would be nice to add a color for the pressed state (`badgePressed`).

src/utils/css.js
```diff
@@ -1,3 +1,9 @@
+import { MONOSPACE_FONT_FAMILY } from '../components/BaseStyles/BaseStyles'
+
+export const monospace = () => `
+  font-family: ${MONOSPACE_FONT_FAMILY}, monospace;
```
bpierre: This should really be in the text styles, maybe we can name these and add them now? What about `address1` (for the address displayed on badges and inside text fields)? /cc @owisixseven

src/theme/theme-light.js
sohkai: > Also I’m not sure what the decision is for the pressed state, but if we don’t want to add a shadow on this component, then it would be nice to add a color for the pressed state (badgePressed).

Yes... @dizzypaty and I discussed briefly and decided to use `surfacePressed` since it was consistent with the other pressed surfaces, but I do agree it doesn't look very distinct (`surfacePressed` is very close to white).

bpierre: Ah yes, we should probably still create a new entry even if it uses the same color than `surfacePressed`, because we can’t guarantee they will have the same value in the dark theme or others.

src/components/IdentityBadge/IdentityBadge.js
dizzypaty: I think it works well with both font sizes, so let's use `Body 2` (16px) to keep it consistent with the theme text styles.

src/components/AddressField/AddressField.js
```diff
@@ -1,124 +1,132 @@
...
+  const theme = useTheme()
+  const inputRef = useRef(null)
+
+  const handleFocus = useCallback(() => {
+    inputRef.current && inputRef.current.select()
+  }, [])
+  const handleCopy = useCallback(() => {
+    if (inputRef.current) {
+      inputRef.current.focus()
+      inputRef.current.select()
```
bpierre: Just as a note for later (better not touch it now), I’m wondering if `focus()` is needed since `select()` probably always implies the focus being moved at the selection level :thinking:

src/theme/theme-light.js
dizzypaty: If possible we should add the badge styles to the light theme: 

`badge`: colors.AragonBlueLighter
`badgeContent`: colors.GreyDark
`badgePressed`: colors.ArcticBlueLight
`address1`: Overpass Mono Regular, 16px, line-height 1.5

dizzypaty: When `badge` displays a custom label, then text-style is `body2`

bpierre: Let’s also make sure we have these in `theme-dark` so it doesn’t error (even if the values are incorrect).

src/utils/css.js
sohkai: Done in 8e5843e.

src/theme/theme-light.js
sohkai: Done in ae53216.

src/style/text-styles.js
```diff
@@ -57,13 +59,26 @@ const TEXT_STYLES = {
     lineHeight: 1.5,
     transform: 'uppercase',
   },
+  address1: {
+    size: 16,
+    weight: 400,
+    lineHeight: 1.5,
+    monospace: true,
+  },
+  address2: {
```
sohkai: @bpierre @dizzypaty I've added `address2` for the AddressField's input text size. Unfortunately the text grows quite long when it's 16px, and 14 was the closest other value that we use (designs have this at 15px).

src/components/IdentityBadge/IdentityBadge.js
```diff
@@ -103,8 +105,8 @@ const IdentityBadge = React.memo(function IdentityBadge({
               text-overflow: ellipsis;
               overflow: hidden;
               ${textStyle(!customLabel && address ? 'address1' : 'body2')}
+              ${labelStyle}
```
sohkai: @bpierre Honestly not sure if this is the right way to do this, do we have other patterns or other ways of solving this (styling an inner component)?

bpierre: Nope, it’s something I try to avoid as much as possible, bu when needed, it doesn’t seem too bad to do it like that.

Another solution could be to export this as a component and let authors extend it + pass it, but it’s more verbose:

```jsx
<IdentityBadge
  label={
    <IdentityBadge.Label css="color: tomato">
      0xcafe
    </IdentityBadge.Label>
  }
/>
```

sohkai: Mmm yeah, what do you think about keeping this as-is? I think the only thing that people will override is the text style (to be similar to the old fontSize API).

bpierre: Yes agreed, let’s keep it this way :+1:

PR #558, opened 2019-09-03
Fix some component warnings

sohkai:
- Button: `warnOnce()` was used incorrectly
- Popover: try to warn if the `opener` was a `<button>` or `<input>`, and `closeOnOpenerFocus` is set

The Popover's a bit of a PITA. On macOS, both Firefox and Safari don't focus buttons or inputs on click, hence causing `focusevent.relatedTarget` to always be `null`. Users should only be using `closeOnOpenerFocus` if they expect the opener to be focused, so this adds a warning to let users know they should consider using a different focusable element instead.

Note that this check is **not** full-proof, since it's a PITA for us to know if the passed in element is actually focusable (we'd need to implement a tag whitelist, check for `tabIndex`, etc. which probably isn't worth it now).

### DIFF
--- a/src/components/Button/Button.js
+++ b/src/components/Button/Button.js
@@ -149,19 +149,24 @@ function Button({
 }) {
   // backward compatibility and deprecated props
   if (iconOnly) {
-    warnOnce('Button: "iconOnly" is deprecated, please use "display".')
+    warnOnce(
+      'Button:iconOnly',
+      'Button: "iconOnly" is deprecated, please use "display".'
+    )
     display = 'icon'
   }
   if (mode === 'outline' || mode === 'secondary') {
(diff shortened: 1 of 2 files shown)


src/components/Button/Button.js
```diff
@@ -149,19 +149,24 @@ function Button({
 }) {
   // backward compatibility and deprecated props
   if (iconOnly) {
-    warnOnce('Button: "iconOnly" is deprecated, please use "display".')
+    warnOnce(
+      'Button:iconOnly',
```
bpierre: Oops :see_no_evil: It just made me wonder why we have a key at all haha (instead of using the message directly).

sohkai: Haha yes, it's not like we ever re-use the keys ;).

PR #559, opened 2019-09-03
Design polish dropdown

AquiGorka:
The second commit (https://github.com/aragon/aragon-ui/commit/688ba3940d7fdc52192613e17261ac07233d7057) is a first attempt to get the `Popover`'s width and use that as the `min-width` for the container, but it only works after the first time the `Popover` appears.

Update: as discussed with @bpierre , followed the approach of having an invisible (out of viewport) content to get the width from.

### DIFF
--- a/devbox/apps/DropDown.js
+++ b/devbox/apps/DropDown.js
@@ -14,14 +14,7 @@ const items = [
       justifyContent: 'space-between',
     }}
   >
-    <span
-      css={`
-        /* Adjust for no descenders (due to textstyle: uppercase) */
-        padding-top: 4px;
-      `}
-    >
-      Test
-    </span>
+    <span>Test</span>
     <IdentityBadge entity="0xc41e4c10b37d3397a99d4a90e7d85508a69a5c4c" />
   </span>,
 ]
--- a/devbox/apps/DropDown.js
+++ b/devbox/apps/DropDown.js
@@ -14,7 +14,14 @@ const items = [
       justifyContent: 'space-between',
     }}
   >
-    <span>Test</span>
+    <span
+      css={`
+        /* Adjust for no descenders (due to textstyle: uppercase) */
+        padding-top: 4px;
+      `}
+    >
+      Test
+    </span>
     <IdentityBadge entity="0xc41e4c10b37d3397a99d4a90e7d85508a69a5c4c" />
   </span>,
 ]
--- a/src/components/DropDown/DropDown.js
+++ b/src/components/DropDown/DropDown.js
@@ -192,7 +192,7 @@ const DropDown = React.memo(function DropDown({
           color: ${disabled ? theme.disabledContent : theme.surfaceContent};
           border: ${disabled ? 0 : 1}px solid
             ${closedWithChanges ? theme.selected : theme.border};
-          ${textStyle('label2')};
+          ${textStyle('body2')};
           ${disabled
             ? 'font-weight: 600;'
             : `
--- a/src/components/DropDown/DropDown.js
+++ b/src/components/DropDown/DropDown.js
@@ -213,6 +213,8 @@ const DropDown = React.memo(function DropDown({
             position: absolute;
             top: -100vh;
             left: -100vw;
+            opacity: 0;
+            visibility: hidden;
           `}
         >
           <PopoverContent
--- a/src/components/DropDown/DropDown.js
+++ b/src/components/DropDown/DropDown.js
@@ -120,8 +120,10 @@ const DropDown = React.memo(function DropDown({
     return -1
   }, [active, selected])
 
+  const [getContentWidth, setGetContentWidth] = useState(true)
   const { refCallback: popoverRefCallback } = useButtonRef(el => {
     setPlaceholderMinWidth(el.clientWidth)
+    setGetContentWidth(false)
   })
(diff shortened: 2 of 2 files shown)


src/components/DropDown/DropDown.js
```diff
@@ -118,6 +120,15 @@ const DropDown = React.memo(function DropDown({
     return -1
   }, [active, selected])
 
+  const [getContentWidth, setGetContentWidth] = useState(true)
+  const { refCallback: popoverRefCallback } = useButtonRef(el => {
+    setPlaceholderMinWidth(el.clientWidth)
```
sohkai: We need to adjust this width by 4GU to add extra room on the right for the caret

src/components/DropDown/DropDown.js
```diff
@@ -118,6 +120,15 @@ const DropDown = React.memo(function DropDown({
...
+    setGetContentWidth(false)
```
sohkai: Perhaps we can move this logic to be closer to the other `useButtonRef`, since it's also doing a resize operation?

src/components/DropDown/DropDown.js
```diff
@@ -184,57 +202,40 @@ const DropDown = React.memo(function DropDown({
...
             color: ${closedWithChanges && !disabled ? theme.accent : 'inherit'};
           `}
         />
       </ButtonBase>
+      {getContentWidth && (
+        <div
+          css={`
+            position: absolute;
+            top: -100vh;
+            left: -100vw;
```
sohkai: Could we add an `opacity: 0` or `visibliity: hidden`?

src/components/DropDown/DropDown.js
```diff
@@ -118,6 +120,15 @@ const DropDown = React.memo(function DropDown({
     return -1
   }, [active, selected])
 
+  const [getContentWidth, setGetContentWidth] = useState(true)
```
sohkai: Should we also reset `getContentWidth` to true if the items changes?

src/components/DropDown/DropDown.js
AquiGorka: I don't think it is needed though

src/components/DropDown/DropDown.js
AquiGorka: Not sure if it can be closer, each callback has different references, or maybe I misunderstand, what do you mean by closer?

src/components/DropDown/DropDown.js
```diff
@@ -118,6 +120,15 @@ const DropDown = React.memo(function DropDown({
...
+  const { refCallback: popoverRefCallback } = useButtonRef(el => {
```
bpierre: We should rename `useButtonRef` if it’s not only used for the button ref anymore.

But perhaps we could use `useCallback()` directly here?

```jsx
const popoverRefCallback = useCallback(el => {
  if (el) {
    setPlaceholderMinWidth(el.clientWidth)
    setGetContentWidth(false)
  }
}, [])
```

AquiGorka: It is still used, this is another usage of the same hook.

src/components/DropDown/DropDown.js
sohkai: Hmm yes, didn't notice the `useState()` was dependent on this running; let's keep it here.

sohkai: Just moving it close to the other logic (e.g. `useEffect()`) that are about resizing.

src/components/DropDown/DropDown.js
sohkai: Fixed in 638b9a7.

If you change [this line](https://github.com/aragon/aragon-ui/blob/design%2Fdropdown/devbox/apps/DropDown.js#L45) to instead be `width="50px"` (or similarly small), this should be the correct result after selecting a long option:

Screen Shot 2019-09-04 at 5 21 22 PM

However, before 638b9a7, this is what would appear:

Screen Shot 2019-09-04 at 5 21 40 PM

devbox/apps/DropDown.js
```diff
@@ -14,7 +14,14 @@ const items = [
       justifyContent: 'space-between',
     }}
   >
-    <span>Test</span>
+    <span
+      css={`
+        /* Adjust for no descenders (due to textstyle: uppercase) */
+        padding-top: 4px;
```
bpierre: I don’t think we should use an uppercase text style in the DropDown?

src/components/DropDown/DropDown.js
```diff
@@ -163,13 +184,15 @@ const DropDown = React.memo(function DropDown({
...
+          padding-left: ${2 * GU}px;
+          padding-right: ${1.5 * GU}px;
           width: ${width || (wide ? '100%' : 'auto')};
+          min-width: ${placeholderMinWidth}px;
           background: ${disabled ? theme.disabled : theme.surface};
           color: ${disabled ? theme.disabledContent : theme.surfaceContent};
           border: ${disabled ? 0 : 1}px solid
             ${closedWithChanges ? theme.selected : theme.border};
-          ${textStyle('body2')};
+          ${textStyle('label2')};
```
bpierre: I don’t think this should be uppercase.

sohkai: Indeed, fixed in 8689c72.

PR #561, opened 2019-09-03
Button: fix the shadow of the pressed state

bpierre:
The Button shadow has been [changed](https://github.com/aragon/aragon-ui/commit/62b7a0fa0c7d351264f37829b91eb396340e9a65) at some point, but without adapting its pressed state. This change fixes that.

### DIFF
--- a/src/components/Button/Button.js
+++ b/src/components/Button/Button.js
@@ -253,7 +253,7 @@ function Button({
         transition-timing-function: ease-in-out;
         &:active {
           transform: ${disabled ? 'none' : 'translateY(1px)'};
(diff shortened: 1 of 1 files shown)


bpierre:
@AquiGorka Yes I’ve been resisting having shadows defined anywhere until we figure out what we need exactly, but we should start working with @dizzypaty @owisixseven to define the different values we use across the different components.

We should probably store them as number values, so they can be used in different contexts (box shadow, multiple box shadows, transitions, drop-shadow), and per level of elevation. We could also have a `box-shadow` CSS utility that would know the different names, because it is the most common use case:

```jsx
<div
  css={`
    // 100 would be the first elevation level,
    // 200 the next one etc. (like font weights)
    ${shadow('100')};
  `}
/>
```

Issue #448, opened 2019-08-11
IdentityBadge: avoid rendering as <button> if `badgeOnly` is set


sohkai:
We use `badgeOnly` in places already nested by a `<button>`, such as in `Autocomplete` usages.

In these cases, because `IdentityBadge` also renders a `<button>`, we get an invalid DOM nesting (`<button>`s cannot be nested in `<button>`s). Moreover, the `IdentityBadge` does not really act as a button when it is in `badgeOnly` mode.

PR #563, opened 2019-09-03
Dropdown + IdentityBadge fixes

bpierre:
-  DropDown: declare `Inside` slots.
- DropDown: stop using closeOnOpenerFocus (removes the warning).
-  IdentityBadge: set `badgeOnly` to `true` when inside of a `DropDown`. 
-  IdentityBadge: use `element="a"` when `badgeOnly=true` (fixes https://github.com/aragon/aragon-ui/issues/448).

### DIFF
--- a/devbox/apps/DropDown.js
+++ b/devbox/apps/DropDown.js
@@ -1,8 +1,23 @@
 import React from 'react'
 import styled from 'styled-components'
-import { DropDown, unselectable } from '@aragon/ui'
+import { IdentityBadge, DropDown, unselectable } from '@aragon/ui'
 
-const items = ['Strawberry', 'Banana', 'Apple', 'Cherry']
+const items = [
+  'Strawberry',
+  'Banana',
+  'Apple',
+  'Cherry',
+  <span
+    style={{
+      display: 'flex',
+      width: '100%',
+      justifyContent: 'space-between',
+    }}
+  >
+    <span>Test</span>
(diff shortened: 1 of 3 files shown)


AquiGorka:
image

bpierre:
@AquiGorka a bit of context would help, is it an issue caused by these changes?

AquiGorka:
Oh, sure, I assumed you would see it right away, if you execute the devbox, then choose the new option you included recently the contents of the DropDown selected item do not appear to be centered vertically.

bpierre:
Oh right! That’s really weird, are you testing on Chrome? Also, is it uppercase because you made it so, or is it rendering like that by default?

That’s what I get on Chromium 78:

AquiGorka:
This is your same screenshot:
image

I think it does look a bit off, the uppercase I've no idea where it is coming from.

bpierre:
Ah yes this is normal, there is no descender in this example. A text box centered without any descender always look a bit “up”, except if we fix it manually (that’s what we are doing in `Tag` when it’s guaranteed to be a number, or uppercase). This is the nature of dynamic text, there is not much to do until we have proper CSS API to say “I prefer this to be visually aligned, no matter where the text box is”.

Here is the “fixed” version (I just added characters with descenders):

image

And you can see the text box here:

image

AquiGorka:
Alright, thank you for clarifying.

PR #564, opened 2019-09-03
Add AppBadge

sohkai:
Adds an AppBadge.

See designs from [Figma](https://www.figma.com/file/bZ1j7omfhdMCbtFahP3nEJ9B/Client?node-id=600%3A233).

--------

Also does a few other things to the other badges, now that I've spent some more time / noticed them:

- Consolidated `disabled` and `badgeOnly` external props so it's just `badgeOnly`
- Adjusted spacing for IdentityBadge's 'you' tag, but discussing with @dizzypaty exactly how these tags should look due to sizing discrepancies
- Added a fallback timer to `useImageExists()`, to let users know if the image is taking a long time to load
- Adjusted the AddressField's icon slot so that there's no border applied by default (missed this in https://github.com/aragon/aragon-ui/pull/557)


src/components/AppBadge/README.md
```diff
@@ -0,0 +1,69 @@
...
+
+### `iconSrc`
+
+| Type     | Default value   |
+| -------- | --------------- |
+| `String` | None            |
+
+Source of app's icon asset, preferrably 24x24. A default app icon will be used if this prop is not provided or if the given src was not loadable.
+
+###  `identifier`
```
bpierre: I wonder if we need the `identifier` / `label` distinction at this level, would it be possible to only have `label`?

bpierre:
@sohkai LGTM!

src/components/AppBadge/README.md
```diff
@@ -0,0 +1,69 @@
...
+
+The app's contract address.
+
+### `iconSrc`
+
+| Type     | Default value   |
+| -------- | --------------- |
+| `String` | None            |
+
+Source of app's icon asset, preferrably 24x24. A default app icon will be used if this prop is not provided or if the given src was not loadable.
```
AquiGorka: What happens if a user does not want to render an icon but still use this badge? Can we make this boolean?

src/components/AppBadge/README.md
```diff
@@ -0,0 +1,69 @@
...
+
+```jsx
+import { AppBadge } from '@aragon/ui'
+
+function App() {
+  return (
+    <AppBadge
+      appAddress="0x281c36aee91…c31ef3fc115"
+      iconSrc="ipfs.io/ipfs/..."
+      label="50% 50%"
```
AquiGorka: I was thinking of a similar api, but then I figured it would make sense to make this consistent with `IdentityBadge` that uses `customLabel` and `entity` (my first implementation actually used `IdentityBadge` with a few modifications to show the icon.

src/components/AppBadge/README.md
sohkai: Hmm, yes, maybe we can just override the popover's title with a prop like we do in IdentityBadge.

In that case I'd prefer to keep the `identifier` instead, as that's inherent to an app (rather than a custom label / custom popover title, which is external).

src/components/AppBadge/README.md
sohkai: They should probably use a different badge; AFAIK we don't have a case where this badge is shown without the icon.

src/components/AppBadge/README.md
sohkai: I'll likely deprecate `customLabel` and settle on `label` as an API, since it'll be more obvious setting this prop is meant to override the badge's label.

AquiGorka: Hmm, I could not find the discussion but I recall my first implementation of the custom label identity badge used `label` and then it was updated to `customLabel`, I'm just trying to be consistent here since we've had this conversation before but ok, this is a different use case and `label` would have been my first option too (if `IdentityBadge` did not exist).

src/components/AppBadge/README.md
AquiGorka: Still, `IdentityBadge` has the option to not show an icon, wouldn't it make sense to allow for it here too?

sohkai: `IdentityBadge` only doesn't display the icon if you don't give it an address, at which point there's no way for it to display an icon. Apps should always have icons, or if not, use the default.

src/components/AppBadge/README.md
sohkai: I was looking back at the discussion in https://github.com/aragon/aragon-ui/pull/327 and AFAIK we didn't discuss `customLabel`; it was just for the popover action's naming.

AquiGorka: Yep, that's where I looked too and couldn't find it, but one of the first commits uses `label` instead of `customLabel` so, somewhere in there a discussion happened that got me to change the prop name - maybe it happened offline.

src/hooks/useImageExists.js
```diff
@@ -3,13 +3,18 @@ import cachedMap from '../utils/cached-map'
 
 const srcCache = cachedMap()
 
-// Check if a remote image exists.
-export function useImageExists(src) {
+// Check if a remote image exists and can be loaded within a specific amount of time.
+export function useImageExists(src, timeUntilFallback = 50) {
```
sohkai: @bpierre I've opted just to include the fallback timer here, since `RemoteImage` is pretty redundant if `useImageExists()` already has this built-in.

On success, I choose to keep the `displayFallback` state as-is, in case a user wanted to keep displaying the fallback even after success.

src/hooks/useImageExists.js
```diff
@@ -48,10 +54,16 @@ export function useImageExists(src) {
     init()
 
     return done
-  }, [src])
+  }, [src, timeUntilFallback])
 
-  return useMemo(() => ({ src, exists, loading }), [src, exists, loading])
+  return useMemo(() => ({ src, displayFallback, exists, loading }), [
+    src,
+    displayFallback,
```
bpierre: Should we set `exist` to `false` if `timeUntilFallback` is set to anything else than `-1`? If not we should probably move it at the very end, to not break the current API.

sohkai: Had to look at this a few times, maybe I should revise this a bit to make it more clear with a proper `return`.

I _think_ what you thought was that this returns an array, and that we shouldn't add items to displace the old API?

It actually returns an object with those fields so we can add more fields as necessary.

bpierre: aaah yes this is all correct, sorry for that :man_facepalming:

src/components/AppBadge/README.md
bpierre: Let’s keep them both for now, and see how we are using the component before deciding :+1:

sohkai:
I've just pushed a few final commits to adjust the styling for the designs:

- Popover action uses 2GU internal padding
- Use consistent tag sizing for 'you' and custom labels

PR #565, opened 2019-09-03
Add BadgeBase

sohkai:
@bpierre What do you think about consolidating all of IdentityBadge, TokenBadge, TransactionBadge, and AppBadge via this BadgeBase? It's really an internal component that sets up the base styling and behaviours for each Badge component so they don't have to do as much themselves.

The name conflicts a bit with the deprecated Badge, but I think this is OK (users should think of that as Tags now).

### DIFF
--- a/devbox/apps/IdentityBadge.js
+++ b/devbox/apps/IdentityBadge.js
@@ -16,12 +16,12 @@ class App extends React.Component {
         `}
       >
         <IdentityBadge
-          customLabel="Olive Oyl"
+          label="Olive Oyl"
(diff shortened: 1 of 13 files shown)


bpierre:
> @bpierre What do you think about consolidating all of IdentityBadge, TokenBadge, TransactionBadge, and AppBadge via this BadgeBase?

That’s a great idea yes!

> The name conflicts a bit with the deprecated Badge, but I think this is OK (users should think of that as Tags now).

Agreed. :+1:

PR #566, opened 2019-09-04
Help

bpierre:
Add the `Help` component.


src/components/Help/README.md
```diff
@@ -0,0 +1,34 @@
...
+
+## Usage
+
+```jsx
+    <Help hint="More info about something">
+      The Minimum Approval refers to the  percentage of the minimum amount of
+      tokens that need to vote YES for a vote to pass.
+    </Help>
+  )
+}
```
sohkai: We should fix this example

src/components/Help/README.md
```diff
@@ -0,0 +1,34 @@
+# Help
+
+Provide contextual help content.
+
+Note: when `Help` is used inside a `Box` heading, it will add the right margin automatically.
```
sohkai:
```suggestion
Note: when `Help` is used inside a `Box` heading, it will add the left margin automatically.
```

src/components/Help/README.md
bpierre: I yes, I wanted to put a realistic example, but this one is perhaps a bit too specific :D (and doesn’t match at all with the hint).

bpierre: https://github.com/aragon/aragon-ui/pull/566/commits/9cf40950401e22cd8d7d1e8b1e31c42ed4ce02ff

bpierre: …and the syntax, which might be what you were referring to! https://github.com/aragon/aragon-ui/pull/566/commits/104147790141104746e0c027024948e1ad480af5

PR #572, opened 2019-09-04
Integrate Help into Field

bpierre:
- Allow children to be a render function that will receive a unique ID, being assigned to the htmlFor prop of the label element (allowing to target a specific element rather than the first one found).
- Declare Inside slots.
- Center using flexbox rather than a custom line-height.
- Add white-space: nowrap.
- Accept React nodes for the label, rather than strings only.

Note: the render prop is a first step that will also be useful for advanced uses, but a solution (to be thought / discussed) will be added later so that `Help` can be used without having to use this feature.

### DIFF
--- a/src/components/Field/Field.js
+++ b/src/components/Field/Field.js
@@ -1,8 +1,10 @@
-import React from 'react'
+import React, { useMemo } from 'react'
 import PropTypes from 'prop-types'
 import { useTheme } from '../../theme'
 import { textStyle, GU } from '../../style'
-import { unselectable } from '../../utils'
+import { unselectable, Inside } from '../../utils'
+
+let fieldId = 1
 
 function Field({ children, label, required, ...props }) {
   const theme = useTheme()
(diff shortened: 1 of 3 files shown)


src/components/Field/Field.js
```diff
@@ -13,45 +15,60 @@ function Field({ children, label, required, ...props }) {
...
+          <div
+            css={`
+              display: flex;
+              align-items: center;
+              height: ${2 * GU}px;
+              margin-bottom: ${0.5 * GU}px;
+              color: ${theme.surfaceContentSecondary};
+              white-space: nowrap;
+              ${textStyle('label2')};
+              ${unselectable()};
```
sohkai:
```suggestion
              ${unselectable};
```

I _think_ we settled on this earlier 😄

src/components/Field/Field.js
```diff
@@ -13,45 +15,60 @@ function Field({ children, label, required, ...props }) {
...
+                  `}
+                  title="Required"
+                >
+                  {'\u00a0*'}
+                </span>
+              )}
+            </Inside>
+          </div>
+          <Inside name="Field:content">
+            {id === null ? children : children({ id })}
```
sohkai: Perhaps we should just do a more obvious check for if children is a function here (to avoid cascading issues if we ever update the implementation of how `id` is obtained).

src/components/Field/Field.js
```diff
@@ -1,8 +1,10 @@
-import React from 'react'
+import React, { useMemo } from 'react'
 import PropTypes from 'prop-types'
 import { useTheme } from '../../theme'
 import { textStyle, GU } from '../../style'
-import { unselectable } from '../../utils'
+import { unselectable, Inside } from '../../utils'
+
+let fieldId = 1
```
sohkai: Would be nice to attach some comments here :)

bpierre:
> does the height stay the same as before (going from line height to height)?

Yes :+1:

src/components/Field/Field.js
bpierre: Yes, I was hesitating and chose the shortest one, but it makes more sense to check again if it’s a function :+1:

src/components/Field/Field.js
bpierre: Done https://github.com/aragon/aragon-ui/pull/572/commits/2a62d319d1091627ec923b69185b77e489a1e1c9

src/components/Field/Field.js
bpierre: Done https://github.com/aragon/aragon-ui/pull/572/commits/5b5cf05e8441cb8298eda5c865dc2872ab9010bc

bpierre:
> Let's also add a few quick docs on the new render prop.

Done in https://github.com/aragon/aragon-ui/pull/572/commits/2a62d319d1091627ec923b69185b77e489a1e1c9

PR #573, opened 2019-09-04
DropDown: update caret colour to surfaceIcon

sohkai:
As discussed with @dizzypaty

### DIFF
--- a/src/components/DropDown/DropDown.js
+++ b/src/components/DropDown/DropDown.js
@@ -208,7 +208,9 @@ const DropDown = React.memo(function DropDown({
           size="tiny"
           css={`
             margin-left: ${1.5 * GU}px;
-            color: ${closedWithChanges && !disabled
+            color: ${disabled
+              ? theme.disabledContent
+              : closedWithChanges
               ? theme.accent
               : theme.surfaceIcon};
           `}
--- a/src/components/DropDown/DropDown.js
+++ b/src/components/DropDown/DropDown.js
@@ -209,7 +209,7 @@ const DropDown = React.memo(function DropDown({
           css={`
             margin-left: ${1.5 * GU}px;
             color: ${disabled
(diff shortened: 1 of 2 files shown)


src/components/DropDown/DropDown.js
```diff
@@ -208,7 +208,9 @@ const DropDown = React.memo(function DropDown({
           size="tiny"
           css={`
             margin-left: ${1.5 * GU}px;
-            color: ${closedWithChanges && !disabled ? theme.accent : 'inherit'};
+            color: ${closedWithChanges && !disabled
+              ? theme.accent
+              : theme.surfaceIcon};
```
bpierre: We should also use `theme.disabledIcon` when disabled (it might be different in other themes).

sohkai: Done in 362ff90, this is how it looks:

Screen Shot 2019-09-04 at 10 32 12 PM

src/components/DropDown/DropDown.js
```diff
@@ -208,7 +208,11 @@ const DropDown = React.memo(function DropDown({
           size="tiny"
           css={`
             margin-left: ${1.5 * GU}px;
-            color: ${closedWithChanges && !disabled ? theme.accent : 'inherit'};
+            color: ${disabled
+              ? theme.disabledContent
```
bpierre:
```suggestion
              ? theme.disabledIcon
```

src/components/DropDown/DropDown.js
bpierre: Just added a code suggestion to use `disabledIcon` rather than `disabledContent`, all good otherwise! :point_up:

sohkai: Ahh missed that I should be using `disabledIcon`! Thanks!

PR #576, opened 2019-09-04
TextInput: remove :invalid styling on some browsers

sohkai:
Removes the default `:invalid` styling that some browsers (e.g. Firefox) apply to inputs.

### DIFF
--- a/src/components/Input/TextInput.js
+++ b/src/components/Input/TextInput.js
@@ -5,53 +5,49 @@ import { warnOnce } from '../../utils'
 import { textStyle, GU, RADIUS } from '../../style'
 
 // Simple text input
-const TextInput = React.forwardRef(
-  ({ multiline, type, value, ...props }, ref) => {
-    const theme = useTheme()
-    return (
-      <input
-        as={multiline ? 'textarea' : 'input'}
-        type={multiline ? undefined : type}
-        value={value}
-        {...props}
-        ref={ref}
-        css={`
-          width: ${({ wide }) => (wide ? '100%' : 'auto')};
-          height: ${5 * GU}px;
-          padding: 0 ${1.5 * GU}px;
-          background: ${theme.surface};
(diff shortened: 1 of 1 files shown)


src/components/Input/TextInput.js
```diff
@@ -5,46 +5,53 @@ import { warnOnce } from '../../utils'
...
+            outline: none;
+            border-color: ${theme.selected};
+          }
+          &:read-only {
+            color: transparent;
+            text-shadow: 0 0 0 ${theme.surfaceContentSecondary};
+            border-color: ${theme.border};
+          }
+          &:invalid {
+            border-color: ${value ? theme.negative : theme.border};
```
sohkai: This isn't super nice... but it looks like an input is `invalid` if it's required and has no value (so it always starts by default in `invalid`).

bpierre: mmm yes, I think it’s a good idea to style these, but relying only on `:invalid` would probably cause problems. Every browser has its own validation mechanism, and doesn’t integrate really well with JavaScript / React where the validation is happening at another level. We could try using [`element.setCustomValidity()`](https://developer.mozilla.org/en-US/docs/Web/API/Constraint_validation#element.setCustomValidity(message))… but it would mean having the error messages reported in different ways depending on the browser, if supported at all.

What would you think of leaving this style for `:invalid` (so we stop having the default one when a field is invalid), but also adding an `invalid` prop, that will apply the same style? Or maybe we need different levels to be represented, in which case we could have something like `validity="success / warning / error / default"` /cc @dizzypaty

sohkai: That sounds good, I'd agree that moving the invalid styling up a level sounds like the least headache down the line.

sohkai: Added 00a251c to just remove the default `:invalid` styling.

Let's review the error states afterwards and add the `invalid` prop then.

PR #578, opened 2019-09-05
Added empty state to DataView

deamme:
## Preview
Empty state


dizzypaty:
Hey @deamme, the design above is meant for the 'No results found' state when users perform a search and no data matches their criteria. We do have designs for a generic a dataView empty state (not sure if part of the aragonUI component yet) where the copy can be customized to reflect the data type, in this case 'No orders yet'. It's on the Fundraising [Figma file](https://www.figma.com/file/bztwrkbhMwtEu4ARgpAdQy/Fundraising-project?node-id=94%3A77) on the Presale UI page. 

FUNDRAISING - Orders 01

deamme:
@dizzypaty The shown preview is only a custom placeholder. The default placeholder is the same illustration with the following text: "No entries found".

devbox/apps/DataView.js
```diff
@@ -115,7 +116,40 @@ function SortingDemo(props) {
...
+      css={`
+        font-size: 28px;
+        margin-top: 1rem;
+      `}
+    >
+      No results found.
+    </p>
+    <p>
+      We can’t find any item matching your search query.{' '}
+      <a css="color: blue;">Clear search</a>
```
bpierre: We could use [`theme.link`](https://github.com/aragon/aragon-ui/blob/newstyle/src/theme/theme-light.js#L70) here.

src/components/DataView/DataView.js
```diff
@@ -282,36 +295,164 @@ const DataView = React.memo(function DataView({
...
+                  />{' '}
+                  Loading data...
+                </p>
+              ))}
+
+            {status === 'empty-filters' && (
+              <>
+                <p
+                  css={`
+                    font-size: 28px;
```
bpierre: Not exactly the same size (26 vs. 28), but we should probably use a text style anyway:

```suggestion
                    ${textStyle('title2')};
```

src/components/DataView/README.md
```diff
@@ -290,3 +290,51 @@ Renders the label used to indicate the current selection. If not provided, “X
...
+
+Can be `default`, `loading`, `empty-filters` and `empty-search`. The correct illustration and text styles are used depending on the status.
+
+### statusEmpty
+
+| Type     | Default value |
+| -------- | ------------- |
+| `Node`   | None          |
+
+If you want to customize the `default` status content.
```
bpierre:
```suggestion
If you want to customize the `default` status content. This status is visible when there are no items.
```

src/components/DataView/DataView.js
```diff
@@ -282,36 +295,164 @@ const DataView = React.memo(function DataView({
...
+                    {'We can’t find any item matching your filter selection. '}
+                    <a
+                      css={`
+                        color: ${theme.link};
+                        &:hover {
+                          cursor: pointer;
+                        }
+                      `}
+                      onClick={onStatusEmptyClear}
+                    >
```
bpierre: We can use `<Link />` for these now :-)

src/components/DataView/DataView.js
```diff
@@ -282,36 +295,164 @@ const DataView = React.memo(function DataView({
...
+                    display: flex;
+                    align-items: center;
+                  `}
+                >
+                  <LoadingRing
+                    css={`
+                      margin-right: ${2 * GU}px;
+                    `}
+                  />{' '}
+                  Loading data...
```
bpierre:
```suggestion
                  Loading data…
```

src/components/DataView/DataView.js
```diff
@@ -282,36 +295,164 @@ const DataView = React.memo(function DataView({
...
+            onSelect={toggleEntrySelect}
+            onSelectAll={selectAll}
+            renderSelectionCount={renderSelectionCount}
+            rowHeight={tableRowHeight}
+            selectable={canSelect}
+            selectedCount={selectedIndexes.length}
+          />
+        ))}
+
+      {emptyEntries && (
```
bpierre: I was wondering if it would make sense to always display the empty state when `status=loading`, even when there are entries… but maybe it’s better to do it like you did, so that all the empty states work the same way (= only displayed when there are no entries). :+1:

src/components/DataView/DataView.js
```diff
@@ -282,36 +295,164 @@ const DataView = React.memo(function DataView({
...
+            align-items: center;
+          `}
+        >
+          <div
+            css={`
+              width: ${31 * GU}px;
+              text-align: center;
+            `}
+          >
+            {(!status || status === 'default' || status === 'loading') && (
```
bpierre: We could set `status=default` in the `defaultProps`, so that we don’t have to deal with `!status`.

src/components/DataView/DataView.js
```diff
@@ -282,36 +295,164 @@ const DataView = React.memo(function DataView({
...
+                  </p>
+                )}
+              </>
+            )}
+
+            {status === 'empty-search' && (
+              <>
+                <p
+                  css={`
+                    font-size: 28px;
```
bpierre:
```suggestion
                    ${textStyle('title2')};
```

src/components/DataView/DataView.js
```diff
@@ -282,36 +295,164 @@ const DataView = React.memo(function DataView({
...
+                    margin-top: ${2 * GU}px;
+                  `}
+                >
+                  No results found.
+                </p>
+                {statusEmptySearch || (
+                  <p>
+                    {'We can’t find any item matching your search query. '}
+                    <a
```
bpierre: `<Link />` can be used here too.

src/components/DataView/DataView.js
deamme: Hmm, you might run into this use case

src/components/DataView/DataView.js
deamme: Ahh, yes 👍

src/components/DataView/DataView.js
deamme: Let's leave it for now tho

PR #581, opened 2019-09-06
ToastHub: update raw values to use theme

AquiGorka:


src/components/ToastHub/ToastHub.js
```diff
@@ -201,11 +206,11 @@ const Message = styled(animated.div)`
 
 const Content = styled.div`
   ${textStyle('body3')}
-  color: white;
+  color: ${({ theme }) => theme.surface};
```
bpierre: `surface` is a color that should only be used for backgrounds. We should also ensure that the background color is using a theme color as well, instead of being hard coded one: if not, it would make the content almost invisible in the dark theme.

We probably need to introduce a new color for these in the theme, we were talking with @dizzypaty about maybe using `theme.floating` / `theme.floatingContent`.

AquiGorka: Yep. Ahead of you, asked in Figma for the combination, will update when I get a reply. 👍

dizzypaty: Hey Gorka, please create / use `theme.floating` & `theme.floatingContent`  : )

AquiGorka: Added them in this PR.

src/theme/theme-light.js
```diff
@@ -85,6 +85,9 @@ export default {
   accentEnd: colors.AragonTurquoise,
   accentContent: colors.White,
 
+  floating: 'rgba(48, 64, 79, 0.8)',
```
sohkai: We should add this to `aragon-colours.js` and remove the alpha

AquiGorka: As `Floating`?

Issue #338, opened 2019-03-26
Identity Badge: rounded corners not properly rendered on iOS WebKit


izqui:
Seems like both left corners of the Identity Badge are not getting properly rounded in mobile Safari and other apps that use iOS native WebKit webviews:

![image](https://user-images.githubusercontent.com/447328/54987330-b4f25a00-4fb4-11e9-9d8f-3d66db92cf03.png)

Tested on both Cipher and mobile Safari (iOS 12.0)

izqui:
Just tested on iOS 12.1.4 on an iPad and can't reproduce the issue. Updating my other device to the latest iOS version to see if it was only an issue on previous versions.

sohkai:
@izqui Could you retest on the latest version of the Aragon client?

PR #583, opened 2019-09-06
EthIdenticon + BadgeBase fixes

bpierre:
Fixes https://github.com/aragon/aragon-ui/issues/338

- EthIdenticon: fix border-radius on Chrome.
- AddressField: use a `<a>` when read only.

### DIFF
--- a/src/components/AddressField/AddressField.js
+++ b/src/components/AddressField/AddressField.js
@@ -61,8 +61,13 @@ const AddressFieldBase = React.memo(function AddressFieldBase({
         css={`
           position: relative;
           width: ${HEIGHT}px;
+          height: ${HEIGHT}px;
           overflow: hidden;
           border-radius: ${RADIUS}px 0 0 ${RADIUS}px;
+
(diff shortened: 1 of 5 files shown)


src/components/AddressField/AddressField.js
```diff
@@ -61,8 +61,13 @@ const AddressFieldBase = React.memo(function AddressFieldBase({
         css={`
           position: relative;
           width: ${HEIGHT}px;
+          height: ${HEIGHT}px;
           overflow: hidden;
           border-radius: ${RADIUS}px 0 0 ${RADIUS}px;
+
+          // Fix an issue where the border-radius wasn’t visible on Blink browsers.
+          // See https://gist.github.com/ayamflow/b602ab436ac9f05660d9c15190f4fd7b
+          mask-image: linear-gradient(red, red);
```
sohkai: Is this necessary if we do this in `EthIdenticon`?

bpierre: Yes, because the `border-radius` is applied at the container level in that case (only on one side + can be replaced by another component).

PR #588, opened 2019-09-09
DropDown: fix a prop type error

bpierre:


sohkai:
I think I may have made this change, as we started talking about the DropDown always having a header.

PR #601, opened 2019-09-17
Search input component

AquiGorka:

### DIFF
--- a/src/components/Input/SearchInput.js
+++ b/src/components/Input/SearchInput.js
@@ -1,5 +1,5 @@
-import React, { useRef, useState, useCallback, useMemo } from 'react'
-import ButtonIcon from '../Button/ButtonIcon'
+import React, { useRef, useState, useCallback } from 'react'
+import { ButtonIcon } from '../Button/ButtonIcon'
 import TextInput from './TextInput'
 import { IconSearch, IconCross } from '../../icons'
 import { useTheme } from '../../theme'
@@ -15,7 +15,7 @@ const SearchInput = React.forwardRef(({ onChange, ...props }, ref) => {
       setValue(ev.currentTarget.value)
       onChange(ev)
     },
-    [onChange, value]
+    [onChange]
   )
   const clear = useCallback(() => {
     // https://github.com/facebook/react/issues/11488
@@ -29,7 +29,7 @@ const SearchInput = React.forwardRef(({ onChange, ...props }, ref) => {
       tracker.setValue(value)
     }
     input.dispatchEvent(event)
-  }, [onChange, value, localRef])
+  }, [value, localRef])
 
   return (
     <TextInput
@@ -59,5 +59,9 @@ const SearchInput = React.forwardRef(({ onChange, ...props }, ref) => {
     />
   )
 })
+
+SearchInput.propTypes = {
+  ...TextInput.propTypes,
+}
 
 export { SearchInput }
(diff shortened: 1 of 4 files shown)


src/components/Input/SearchInput.js
```diff
@@ -0,0 +1,66 @@
...
+    // simulate an input change event onChange handler works as if
+    // user had clear with keyboard
+    const input = localRef.current
+    input.value = EMPTY
+    const event = new Event('input', { bubbles: true })
+    if (input._valueTracker) {
+      input._valueTracker.setValue(value)
+    }
+    input.dispatchEvent(event)
+  }, [value, localRef])
```
bpierre: Good find! But I would really like to avoid relying on React internals: it could either break anytime, or require us to move react and react-dom to be strict dependencies :-(

I plan to move move `TextInput` to pass the value directly like the other components, and move the event as an optional parameter:

```jsx
<TextInput
  onChange={((value, { event }) => {
    //
  })}
/>
```

The idea is to make the components pass the data as a first parameter (as most of them already do, so they fit better with the React flow), and the events as optional parameters depending on the internal DOM structure the component relies on.

For `SearchInput`, it could translate to something like that:

```jsx
<SearchInput
  onChange={((value, { inputChangeEvent, clearClickEvent }) => {
    //
  })}
/>
```

And `onChange` could be called directly with the emptied value on click. I’m not sure we would ever need to use these events (we really only use them to access `currentTarget.value`), but they would be available as an advanced part of the API, if needed.

What would you think of moving to pass the value for `SearchInput`? That way, using `dispatchEvent()` wouldn’t be needed as `onChange("")` could be called directly.

src/components/index.js
```diff
@@ -78,6 +78,7 @@ export { TabBarLegacyCompatibility as TabBar } from './Tabs/Tabs'
 export { default as Text } from './Text/Text'
 
 // Experimental components
+export { SearchInput as _SearchInput } from './Input/SearchInput'
```
bpierre: I think we can remove the experimental flag :smile: Could we also add a README?

AquiGorka: Sure, I figure submitting the first revision as experimental makes sense, and after the first review we can decide if we make it part of the whole bag or keep as experimental.

src/components/Input/SearchInput.js
AquiGorka: Done.

AquiGorka:
Hey @bpierre thanks.
Can you clarify on removing the uncontrolled mode? The way I understand this: the value is controlled at the `SearchInput` level and changes from the `TextInput` are in sync with such state, in a way that the source of truth is always the state at `SearchInput`.

bpierre:
Sorry I meant removing the optional internal state, removing the default value, and only rely on `value` passed from the props. The term “uncontrolled” comes from the [React docs](https://reactjs.org/docs/uncontrolled-components.html).

AquiGorka:
@bpierre I removed the `defaultValue` here: c0e5c8304a56d736f585bab1a795e346e5f10b64 and then remove the internal state here: https://github.com/aragon/aragon-ui/pull/601/commits/f39db3748307d066de67dafdc5bb5cd1ca44ac7f
Then updated the devbox example so that it behaves in a controlled manner here: https://github.com/aragon/aragon-ui/pull/601/commits/b74dfc4a2b7b8df88724b7e88511e7a3eb3e5a45

PR #606, opened 2019-09-20
Modal: fix the space around

bpierre:


src/components/Modal/Modal.js
```diff
@@ -141,7 +143,7 @@ Modal.defaultProps = {
   closeButton: true,
   onClose: noop,
   padding: 3 * GU,
```
AquiGorka: Does the padding have anything to do with the space around? I'm just trying to wrap this in my head and think maybe this value could be part of the equation.

bpierre: Nope, the space around is using `padding` in CSS, but it is not related to the `padding` prop which is only used for the internal padding.

PR #608, opened 2019-09-30
Dropdown & Popover fixes for dimensions and scrolling

AquiGorka:
These changes fix the following issues (from https://github.com/aragonone/product/issues/196)

- Finance Mobile
  - Fix DropDown in new deposit
  - New Transfer Deposit on Mobile: DropDown Popover does not scroll so it is not possible to select one of the first tokens nor close the list options without choosing one.

### DIFF
--- a/src/components/DropDown/DropDown.js
+++ b/src/components/DropDown/DropDown.js
@@ -113,7 +113,7 @@ const DropDown = React.memo(function DropDown({
 
   const [widthNoPx = MIN_WIDTH] = (width || '').split('px')
   const [buttonWidth, setButtonWidth] = useState(0)
-  const [getContentWidth, setGetContentWidth] = useState(true)
+  const [measureWidth, setMeasureWidth] = useState(true)
 
   // Adjust the button width if the item widths are larger than declared width
   const [ghostElement, setGhostElement] = useState(null)
@@ -134,7 +134,7 @@ const DropDown = React.memo(function DropDown({
   }, [ghostElement, widthNoPx])
 
   useEffect(() => {
-    setGetContentWidth(true)
+    setMeasureWidth(true)
   }, [vw, items])
 
   // Update the button width every time the reference updates
@@ -223,7 +223,7 @@ const DropDown = React.memo(function DropDown({
           `}
         />
       </ButtonBase>
-      {getContentWidth && (
+      {measureWidth && (
         <div
           css={`
             position: absolute;
--- a/src/components/DropDown/DropDown.js
+++ b/src/components/DropDown/DropDown.js
@@ -117,18 +117,21 @@ const DropDown = React.memo(function DropDown({
 
   // Adjust the button width if the item widths are larger than declared width
   const [ghostElement, setGhostElement] = useState(null)
-  const popoverRefCallback = useCallback(el => {
-    if (!el) {
-      return
-    }
-    setGhostElement(el)
-  })
-  const placeholderMinWidth = React.useMemo(() => {
+  const popoverRefCallback = useCallback(
+    el => {
+      if (!el) {
+        return
+      }
+      setGhostElement(el)
+    },
+    [setGhostElement]
+  )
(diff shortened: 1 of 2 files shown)


src/components/Popover/Popover.js
```diff
@@ -193,6 +193,9 @@ class PopoverBase extends React.Component {
               /* Having the popover visible already means that it focused. */
               outline: 0;
             }
+            max-height: calc(100vh - ${2 * GU}px);
+            max-width: calc(100vw - ${2 * GU}px);
```
bpierre: One thing to be aware of is that `Popover` [uses the root element](https://github.com/aragon/aragon-ui/pull/608/files#diff-8cb194f33719157dc1accb9f7904a92aR82) when available, which can be smaller than the window width / height. We should probably use the root dimensions here to make it work in nested root contexts.

src/components/DropDown/DropDown.js
```diff
@@ -298,8 +304,10 @@ const PopoverContent = React.memo(function PopoverContent({
         </div>
       )}
       <ul
+        ref={refCallback}
         css={`
           margin: 0;
+          padding: 0;
```
bpierre: Just to be sure, this is in case someone override the padding of `ul` elements and not to fix an actual issue, right? Asking because `BaseStyles` is supposed to take care of that already, but I think it’s ok to also have it here.

src/components/Popover/Popover.js
AquiGorka: Good suggestion, any tip on how to fetch the root element dimensions?

src/components/DropDown/DropDown.js
AquiGorka: Yep, just to make sure we are getting the correct "content" width of the elements with no extra spacing - I saw base styles does this but wanted to be explicit.

src/components/DropDown/DropDown.js
```diff
@@ -112,22 +112,27 @@ const DropDown = React.memo(function DropDown({
...
-      return
+  const [ghostElement, setGhostElement] = useState(null)
+  const popoverRefCallback = useCallback(
+    el => {
+      if (!el) {
+        return
+      }
+      setGhostElement(el)
+    },
+    [setGhostElement]
```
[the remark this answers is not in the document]
AquiGorka: > But we might also be able to store the element width directly, so we don’t have to keep a detached element around?

I'll double check but I think this is how it was set before and the updates were not happening so I refactored to use the element in state. i'll get back to you on this one.

src/components/Popover/Popover.js
bpierre: Yes, we already have it on the component! :-)

Something like this should work:

```jsx
boundaryDimensions() {
  const { rootBoundary } = this.props
  return rootBoundary
    ? [ rootBoundary.clientWidth, rootBoundary.clientHeight ]
    : [ window.innerWidth, window.innerHeight ]
}
```

src/components/Popover/Popover.js
```diff
@@ -193,8 +201,8 @@ class PopoverBase extends React.Component {
               /* Having the popover visible already means that it focused. */
               outline: 0;
             }
-            max-height: calc(100vh - ${2 * GU}px);
-            max-width: calc(100vw - ${2 * GU}px);
+            max-height: calc(${maxHeight}px - ${2 * GU}px);
+            max-width: calc(${maxWidth}px - ${2 * GU}px);
```
bpierre: One last thing: we should probably move these in `style`, now that they can update very often (so that styled-components doesn’t create tons of classes as we resize). We could also skip the `calc()` in CSS now that everything is in `px` :-)

src/components/DropDown/DropDown.js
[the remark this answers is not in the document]
AquiGorka: So yeah, was able to use a similar approach to the original that has a dependency on `vw` and `items` to get the new width.

PR #611, opened 2019-10-04
Split: allow to invert (visually) the two slots

bpierre:
Usage:

```jsx
<Split
  invert="horizontal"
  primary={
    <div>
      On large viewports, this will get displayed
      on the right side rather than the left.
    </div>
  }
  secondary={
    <div>
      On large viewports, this will get displayed
        on the left side rather than the right.
    </div>
  }
/>
```

Note: “left” and “right” will mean the opposite when right to left support will be added to aragonUI.

```jsx
<Split
  invert="vertical"
  primary={
    <div>
      On small viewports, this will get displayed second.
    </div>
  }
  secondary={
    <div>
      On small viewports, this will get displayed first.
    </div>
  }
/>
```

/cc @onbjerg @deamme

### DIFF
--- a/src/components/Split/Split.js
+++ b/src/components/Split/Split.js
@@ -4,9 +4,43 @@ import { useLayout } from '../Layout/Layout'
 import { GU } from '../../style'
 import { Inside } from '../../utils'
 
-function Split({ primary, secondary }) {
+function Split({ primary, secondary, invert }) {
   const { name: layout } = useLayout()
   const oneColumn = layout === 'small' || layout === 'medium'
+
+  const inverted =
+    (!oneColumn && invert === 'rows') || (oneColumn && invert === 'columns')
+
+  const primaryContent = (
+    <Inside name="Split:primary">
+      <div
+        css={`
+          flex-grow: 1;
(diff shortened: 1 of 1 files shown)


src/components/Split/Split.js
```diff
@@ -16,31 +51,22 @@ function Split({ primary, secondary }) {
...
-        </Inside>
+        {inverted ? secondaryContent : primaryContent}
+        {inverted ? primaryContent : secondaryContent}
       </div>
     </Inside>
   )
 }
 
 Split.propTypes = {
+  invert: PropTypes.oneOf(['none', 'horizontal', 'vertical']),
```
sohkai: Quick question, more so for @onbjerg and @deamme, would a `'both'` setting be useful?

If you've inverted the horizontal split, such that the right side is the primary element, wouldn't you also want that element to show first in the vertical mode?

bpierre: Yes, we were just discussing it yesterday with @deamme, [we need to add `both`](https://github.com/AragonBlack/fundraising/blob/next/apps/aragon-fundraising/app/src/screens/Presale.js#L50).

Issue #504, opened 2019-08-19
Add TextCopy


sohkai:
A `readonly` TextInput that attaches a button for copying the text.

See https://github.com/aragon/aragon/pull/909#discussion_r315004633.

bpierre:
Done https://github.com/aragon/aragon-ui/pull/612

PR #612, opened 2019-10-06
Add a TextCopy component #504

lcoenen:
This adds a TextCopy component as suggested in #504. 

It's almost finished. Todo: 

- Add a `galery/` page
- Document the properties in `TextCopy.md`

DEMO: https://youtu.be/DjUFLlEzddQ


bpierre:
Hi @lcoenen, thank you for contributing! :fire: 

We would like this component to take the shape of a read-only text field, with an button inside of it, to copy the text inside. Exactly like the text field inside of the [`IdentityBadge`](https://ui.aragon.org/identity-badge/) popup.

Let me know if that works for you!

lcoenen:
@bpierre Oh all right! I can just abstract the logic from `IdentityBadge` then?

image


Should it include the avatar image as well (the image on the left in the image.

bpierre:
> I can just abstract the logic from IdentityBadge then?

Yes, and maybe we can start using the component directly in [`AddressField`](https://github.com/aragon/aragon-ui/blob/newstyle/src/components/AddressField/AddressField.js)? Let me know if you want to do that as well, or we can also do it as a second step.

> Should it include the avatar image as well (the image on the left in the image).

That wouldn’t be generic enough, but we should have a way to insert it yes. What about using a prop named `adornment`, to follow what [`TextInput` does](https://github.com/aragon/aragon-ui/blob/newstyle/src/components/Input/TextInput.js#L68) (but without the positioning / settings, these are not needed here)?

lcoenen:
@bpierre It's easy enough. I decoupled `TextCopy` and `AddressField`, I will test it and push tonight.

bpierre:
Hey @lcoenen, I added a few changes, it’s ready good to be merged now :-)

src/components/TextCopy/README.md
```diff
@@ -0,0 +1,65 @@
...
+
+### `autofocus`
+
+| Type      | Default value |
+| --------- | ------------- |
+| `Boolean` | `false`       |
+
+Focus the text field when the component gets mounted.
+
+### `message`
```
sohkai: What do you think about renaming this "onCopyMessage" or something?

src/components/TextCopy/TextCopy.js
```diff
@@ -0,0 +1,156 @@
...
+    ref
+  ) {
+    const theme = useTheme()
+    const toast = useToast()
+    const inputRef = useRef(null)
+
+    // Allows to focus the component from the outside
+    useImperativeHandle(ref, () => ({
+      focus: () => {
+        inputRef.current.focus()
```
sohkai: Should we check that `inputRef.current` isn't null?

src/components/TextCopy/TextCopy.js
```diff
@@ -0,0 +1,156 @@
...
+          }}
+          autofocus={autofocus}
+          onFocus={handleFocus}
+          readOnly
+          value={value}
+          wide
+          css={`
+            text-overflow: ellipsis;
+            height: ${HEIGHT}px;
+            width: ${52.5 * GU}px;
```
sohkai: Do we want this width and max-width value set here, especially if we've already set `wide` as a prop?

src/components/TextCopy/TextCopy.js
```diff
@@ -0,0 +1,156 @@
...
+            }
+          `}
+        />
+      </div>
+    )
+  })
+)
+
+TextCopy.propTypes = {
+  adornment: PropTypes.node,
```
sohkai: Ahh... we may need to adjust the APIs of a few other components that may have similar logic for placing icons.

I just noticed that the `AppBadge` and `TokenBadge` popovers are using the default address identicon instead of the app / token icon.

src/components/TextCopy/TextCopy.js
```diff
@@ -0,0 +1,156 @@
...
+                color: ${theme.surfaceIcon};
+              `}
+            >
+              <IconCopy />
+            </ButtonIcon>
+          }
+          adornmentPosition="end"
+          adornmentSettings={{
+            // Keep the button square
+            width: HEIGHT - 2,
```
sohkai: I remember trying very hard to keep to the design behaviour, which has the adornment set "over" the TextInput's border.

See old:

Screen Shot 2019-11-05 at 1 39 55 AM

And new:

Screen Shot 2019-11-05 at 1 40 07 AM

You can see in the new one, the TextInput's border goes over the adornment. In particular, I think this looks a bit more "busy" and "enclosing" to the identicon / app icon / token icon.

bpierre: Yes, it has changed! See https://www.figma.com/file/EJpiv6QnJEjDEX3kxDQUlp/UI-kit-(v1)-Light?node-id=145%3A132

src/components/TextCopy/README.md
bpierre: I agree it would be better, but the `on` prefix could make it sound like our events… what about `messageOnCopy`? :laughing:

sohkai: `messageOnCopy` sounds good!

src/components/TextCopy/TextCopy.js
sohkai: Ahhh gotcha!

Should the AppBadge also change? Right now the app icons are above the border: https://www.figma.com/file/EJpiv6QnJEjDEX3kxDQUlp/UI-kit-(v1)-Light?node-id=145%3A132

src/components/TextCopy/TextCopy.js
bpierre: > Ahh... we may need to adjust the APIs of a few other components that may have similar logic for placing icons.

Yes, I’d like to rethink a little bit the way we handle this in different places, maybe we could start with [`TextInput`](https://github.com/aragon/aragon-ui/blob/newstyle/src/components/Input/TextInput.js#L146-L151).

The current version is a bit confusing (e.g. handling the padding and the width separately) and doesn’t support multiple adornments.

A newer version could support these cases:

```jsx
// Default: displayed on the right side,
// takes the width of an icon.
<TextInput
  adornment={<IconSearch />}
/>

// Specify a position.
<TextInput
  adornment={{
    start: <EthIdenticon />,
  }}
/>

// Specify two adornments.
// Note: the term adornment could stay
// singular, as in “adornment configuration”.
<TextInput
  adornment={{
    start: <EthIdenticon />,
    end: <IconSearch />,
  }}
/>

// Specify two adornments with custom widths.
<TextInput
  adornment={{
    start: <EthIdenticon />,
    startWidth: 4 * GU,
    end: <IconSearch />,
    endWidth: 4 * GU,
  }}
/>
```

This would also allow to maintain some level of compatibility with the old API.

> I just noticed that the AppBadge and TokenBadge popovers are using the default address identicon instead of the app / token icon.

I don’t remember if it was intended for `TokenBadge`, maybe as a way to display the associated identicon without having to go to Etherscan? AppBadge definitely has the app icon when opened. I wonder if we could let users toggle between e.g. an app icon and the identicon when clicking on it? @owisixseven @dizzypaty

src/components/TextCopy/TextCopy.js
bpierre: aha well spotted, maybe it hasn’t changed after all :laughing:

Let’s check again with @owisixseven and align everything.

PR #614, opened 2019-10-09
DataView empty state: make the height dynamic

bpierre:


bpierre:
@sohkai Tell me if I missed something, but it should already acts as a min-height for the empty states. If someone sets `statusLoading` to `<div />` (`null` would revert to the default), which is not recommended at all, then it would still have the 8 + 8 gu as padding.

PR #615, opened 2019-10-10
SearchInput: focus after clearing

bpierre:
Also:

- Fix icon colors.
- Let consumers extend the defined props.
- Allow spaces to remove the hint style.

### DIFF
--- a/src/components/Input/SearchInput.js
+++ b/src/components/Input/SearchInput.js
@@ -9,6 +9,7 @@ const EMPTY = ''
 const SearchInput = React.forwardRef(({ onChange, ...props }, ref) => {
   const theme = useTheme()
   const localRef = ref || useRef()
+
   const handleChange = useCallback(
     ev => {
       const { value } = ev.currentTarget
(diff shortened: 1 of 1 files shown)


src/components/Input/SearchInput.js
```diff
@@ -9,45 +9,50 @@ const EMPTY = ''
...
+
+  const handleClearClick = useCallback(
     ev => {
       onChange(EMPTY, { clearClickEvent: ev })
+      if (localRef.current) {
+        localRef.current.focus()
+      }
     },
-    [onChange]
+    [onChange, localRef]
```
sohkai: Curious: is it a good practice to add refs that come as props to the dependency list?

For local refs it's definitely not needed, but in this case, it could come as a random truthy value first and then turn into a ref?

bpierre: I was going to say that it is because a prop is never guaranteed while `useRef()` is… then I realized I was calling `useRef()` conditionally :man_facepalming:. Fixing that!

PR #616, opened 2019-10-10
SidePanel: status, focus, unmount children

bpierre:
## Changes

`SidePanel` is now using `Transition`, which means that children get properly unmounted when the panel closes :tada:. Two new hooks have been added:

### useSidePanel()

This hook should be used inside `SidePanel`, and returns an object with two entries:

- `status` represents the current state of the panel: `opened`, `closed`, `opening`, `closing`.
- `readyToFocus` can be used to focus an element as soon as possible inside the panel without affecting the transition.

About the need to have `readyToFocus`: browser engines tend to force newly focused elements to move inside the viewport, skipping current transitions in the process. `readyToFocus` detects that the transition is half way, which is generally a safe point to start focusing an element without any transition issue. If you still notice a transition issue, another option is to use `status === 'opened'`.

#### Example

This is how `readyToFocus` can be used:

```jsx
// MySidePanel.js

function MySidePanel() {
  const inputRef = useRef()
  const { readyToFocus } = useSidePanel()

  useEffect(() => {
    if (readyToFocus && inputRef.current) {
      inputRef.current.focus()
    }
  }, [readyToFocus, inputRef])

  // The TextInput will get focused as soon as possible.
  return <TextInput wide ref={inputRef} />
}

export default props => (
  <SidePanel title="Panel title" {...props}>
    <MySidePanel />
  </SidePanel>
)
```

And because calling `.focus()` on a reference is a very common use case, another hook is provided to do it easily: `useSidePanelFocusOnReady()`.

### useSidePanelFocusOnReady()

This hook can be used to focus a ref (usually a `TextInput`) when `readyToFocus` is set to `true`.

It takes a React ref as a parameter, or creates one if not provided. It returns this ref.

```jsx
// MySidePanel.js

function MySidePanel() {
  const inputRef = useSidePanelFocusOnReady()

  // The TextInput will get focused as soon as possible.
  return <TextInput wide ref={inputRef} />
}

export default props => (
  <SidePanel title="Panel title" {...props}>
    <MySidePanel />
  </SidePanel>
)
```

Other changes:

- Update README to the new format.
- `title` is now a node rather than a string.
- fix the transition jump near the end (spring precision).

Closes https://github.com/aragon/aragon-ui/issues/347.


src/components/SidePanel/README.md
```diff
@@ -5,77 +5,125 @@ Renders a side panel with an overlay over the app.
...
 
 ### `opened`
 
-- Type: `Boolean`
+| Type      | Default value |
+| --------- | ------------- |
+| `Boolean` | `true`        |
 
-This property toggles the visibility of your SidePanel.
+Toggles the visibility of the panel. Note: when the panel finishes closing, the children components will get unmounted.
```
sohkai: Just wondering if we've tried this unmounting behaviour in the client or some the apps, to see if it's a problem.

I think we generally reset the state of the panels on close anyway, so this shouldn't cause any weird behaviour?

src/components/SidePanel/SidePanel.js
```diff
@@ -32,159 +44,197 @@ function SidePanel({
...
+    [readyToFocus]
+  )
+
+  const handleTransitionDestroyed = useCallback(() => {
+    setReadyToFocus(false)
+  }, [])
+
   useEffect(() => {
-    document.addEventListener('keydown', handleEscape, false)
+    document.addEventListener('keydown', handleEscape)
```
sohkai: Just seems a bit weird we may not need the capturing behaviour anymore; did something change or was this just not needed anyway?

src/components/SidePanel/SidePanel.js
```diff
@@ -219,7 +269,41 @@ Panel.propTypes = {
   compact: PropTypes.bool,
 }
 
+function useSidePanel() {
+  const value = useContext(SidePanelContext)
+  if (value === null) {
+    throw new Error(
+      'useSidePanel() need to be part of the SidePanel tree, ' +
```
sohkai:
```suggestion
      'useSidePanel() was used without a SidePanel existing in the render tree' +
```
Perhaps this is more clear?

src/components/SidePanel/SidePanel.js
```diff
@@ -219,7 +269,41 @@ Panel.propTypes = {
...
+        'which has to be declared at an upper level!'
+    )
+  }
+  return value
+}
+
+function useSidePanelFocusOnReady(ref) {
+  const { readyToFocus } = useSidePanel()
+  const fallbackRef = useRef()
```
sohkai: I'm not convinced we need to make this implicit ref; are other hook APIs doing this?

src/components/SidePanel/SidePanel.js
bpierre: I like that it describes the current situation rather than only saying what to do, but I would like to say that the hook need to be inside of the `SidePanel` render tree, not the other way.

What about this?

```suggestion
      'useSidePanel() was used outside of the SidePanel render tree, ' +
```

bpierre: I like that it describes the current situation rather than only saying what to do, but I would like to say that the hook need to be inside of the `SidePanel` render tree, not the other way.

What about this?

```suggestion
      'useSidePanel() was used outside of the SidePanel render tree, ' +
```

src/components/SidePanel/SidePanel.js
bpierre: It used to be non-optional a loooong time ago (Firefox < 6), but it is fine now, except if you want it to be `true` :-)

src/components/SidePanel/SidePanel.js
sohkai: Yes, let's use this!

src/components/SidePanel/SidePanel.js
bpierre: I don’t know if this pattern is popular or not, but it’s something we’ve been doing in our codebase ([1](https://github.com/aragon/aragon/blob/658c5282ed6e1644245e0ee78a73727f4f7b2a01/src/hooks.js#L180) [2](https://github.com/aragon/aragon/blob/658c5282ed6e1644245e0ee78a73727f4f7b2a01/src/hooks.js#L202) [3](https://github.com/aragon/aragon/blob/658c5282ed6e1644245e0ee78a73727f4f7b2a01/src/hooks.js#L219)).

I don’t think we should avoid creating refs in our hooks if that’s needed − the same way we often create a local state directly in a hook, rather than having authors explicitly passing the value + the updater.

In this case, authors should do this 99% of the time:

```jsx
const ref = useSidePanelFocusOnReady()
```

But being able to pass a custom ref can be useful in some cases, for example if we want to use a ref created by a hook that doesn’t accept custom refs:

```jsx
const ref = useHookWithoutCustomRefSupport()
useSidePanelFocusOnReady(ref)
```

As a general rule, I think hooks using refs should always support this pattern, as it is a good compromise between conciseness and compatibility with non-flexible hooks.

Let me know what you think!

src/components/SidePanel/README.md
bpierre: Yes I was just checking Tokens app: we reset the state on close but it now gets unmounted at this stage, so we’ll be able to remove this code.

src/components/SidePanel/SidePanel.js
sohkai: > As a general rule, I think hooks using (element) refs should always support this pattern, as it is a good compromise between conciseness and compatibility with non-flexible hooks, kind of how React.createRef() creates an “implicit” ref for components if none is passed from the parent.

I agree with this. The analogy with creating internal state makes sense as well, and I hadn't considered refs as "internal" to a hook before.

Issue #596, opened 2019-09-10
Switch component


bpierre:
<p align=center><img width=50% src=https://user-images.githubusercontent.com/36158/64648319-93761500-d412-11e9-9c8c-1a91462a06d6.png>

PR #620, opened 2019-10-15
Switch component

adekbadek:
Fixes https://github.com/aragon/aragon-ui/issues/596

Adds a `Switch` component and updates theme's control-related colors.


src/components/Switch/Switch.js
```diff
@@ -0,0 +1,125 @@
...
+            position: relative;
+            display: inline-block;
+            width: ${WRAPPER_WIDTH}px;
+            height: ${BALL_WIDTH}px;
+            border: ${BORDER}px solid ${theme.border};
+            border-radius: ${BALL_WIDTH}px;
+            background-color: ${checked
+              ? colors.checkedBackground
+              : colors.unCheckedBackground};
+            transition: border-color 100ms, background-color 100ms;
```
bpierre: What would you think of using `50ms` here, to give it more of an “instant” feel on press?

src/components/Switch/Switch.js
```diff
@@ -0,0 +1,125 @@
...
+    unCheckedBackground: disabled ? theme.controlDisabled : theme.surfaceUnder,
+  }
+
+  const handleChange = disabled ? noop : () => onChange(!checked)
+
+  return (
+    <FocusVisible>
+      {({ focusVisible, onFocus }) => (
+        <span
+          onClick={handleChange}
```
bpierre: I’m wondering if we need a `role="checkbox"` here? Maybe not, as assistive tools will see the inner checkbox anyway?

src/components/Switch/Switch.js
adekbadek: yeah, feels a bit better

src/components/Switch/Switch.js
adekbadek: I think the underlying input provides a11y, and adding ARIA attributes may make it look like there are two checkboxes.

src/components/Switch/Switch.js
```diff
@@ -0,0 +1,125 @@
...
+import { Spring, animated } from 'react-spring'
+
+import { useTheme } from '../../theme'
+import { noop } from '../../utils'
+import { springs, GU } from '../../style'
+import FocusVisible from '../FocusVisible/FocusVisible'
+
+const BORDER = 1
+const WRAPPER_WIDTH = 5 * GU
+const BALL_WIDTH = GU * 2.25
```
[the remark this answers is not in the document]
bpierre: Ah yes, handle is probably more consistent with other components :+1: 

I am still wondering if it is not confusing to directly use the value [for the wrapper](https://github.com/aragon/aragon-ui/pull/620/commits/35da1203f1df0a489e12d6d92f740f23059ba7e6#diff-d7fe704bea926ddd248d59674f7cdd1eR34), while the actual handle is using [a value calculated from it](https://github.com/aragon/aragon-ui/pull/620/commits/35da1203f1df0a489e12d6d92f740f23059ba7e6#diff-d7fe704bea926ddd248d59674f7cdd1eR101)? Is it not more `WRAPPER_HEIGHT` than `HANDLE_WIDTH`?

This is assuming we are both using “handle” to talk about [this component](https://github.com/aragon/aragon-ui/pull/620/commits/35da1203f1df0a489e12d6d92f740f23059ba7e6#diff-d7fe704bea926ddd248d59674f7cdd1eR90) :laughing:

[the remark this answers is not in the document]
adekbadek: Yes, by handle I meant this `animated.span :D 
And you're right - it's the other way around indeed - `HANDLE_WIDTH` should be `WRAPPER_HEIGHT`

sohkai:
Btw @adekbadek, it would be really great to have you sign the CLA (https://github.com/aragon/aragon-ui/pull/620#issuecomment-542311208) :).

devbox/apps/Switch.js
```diff
@@ -0,0 +1,44 @@
...
+  min-width: 100px;
+`
+const OptionWrapper = styled.div`
+  display: flex;
+  align-items: center;
+  justify-content: space-between;
+`
+
+// eslint-disable-next-line react/prop-types
+const Option = ({ name, initialyChecked, ...passedProps }) => {
```
sohkai:
```suggestion
const Option = ({ name, initiallyChecked, ...passedProps }) => {
```

Spelling nitpick :)

src/components/Switch/README.md
```diff
@@ -0,0 +1,37 @@
...
+  return (
+    <Switch
+      checked={checked}
+      onChange={setChecked}
+    />
+  )
+}
+```
+
+## Props
```
sohkai: Ahh, we should format these in the new table format, e.g. https://github.com/aragon/aragon-ui/blob/newstyle/src/components/DropDown/README.md#props

src/components/Switch/README.md
```diff
@@ -0,0 +1,37 @@
...
+- Default: `false`
+
+### `disabled`
+
+- Type: `Boolean`
+- Default: `false`
+
+### `onChange`
+
+- Type: `Function`: `(checked: Boolean) -> *`
```
sohkai: Nice! We should definitely do this for our other argument props :)

src/components/Switch/Switch.js
```diff
@@ -0,0 +1,128 @@
...
+                css={`
+                  position: absolute;
+                  left: 0;
+                  z-index: 1;
+                  top: ${BORDER}px;
+                  width: ${WRAPPER_HEIGHT - BORDER * 4}px;
+                  height: ${WRAPPER_HEIGHT - BORDER * 4}px;
+                  border-radius: ${WRAPPER_HEIGHT - BORDER * 4}px;
+                  background-color: ${theme.surface};
+                  box-shadow: 0px 2px 2px rgba(0, 0, 0, 0.15);
```
sohkai: I think we've mostly tried to not use alphas in shadows, but need to double check with @bpierre.

src/theme/theme-light.js
```diff
@@ -76,9 +76,9 @@ export default {
   disabledContent: colors.GreyMedium,
   disabledIcon: colors.ArcticBlueDarker,
 
-  control: colors.ArcticBlue,
+  control: colors.ArcticBlueLight,
   controlBorder: colors.GreyBasic,
-  controlBorderPressed: colors.Grey,
+  controlBorderPressed: colors.ArcticBlueDark,
```
sohkai: Should we change any old control components to use these new colours?

src/components/Switch/Switch.js
```diff
@@ -0,0 +1,128 @@
...
+                  width: ${WRAPPER_WIDTH + BORDER * 2}px;
+                  height: ${WRAPPER_HEIGHT + BORDER * 2}px;
+                  border-radius: ${WRAPPER_HEIGHT}px;
+                  border: 2px solid ${theme.focus};
+                }
+              `
+              : ''};
+          `}
+        >
+          <input
```
sohkai: It looks like we're using a `button` in `Checkbox`, with an aria-role set. The nice thing about `button` is it gets us the 'enter' key functionality for free (as opposed to just 'space'), for toggling the state.

Was a bit surprised the `input` element doesn't respond on 'enter's, but we should probably add it to have the same behaviour as the other controls.

src/components/Switch/README.md
```diff
@@ -0,0 +1,37 @@
...
+      checked={checked}
+      onChange={setChecked}
+    />
+  )
+}
+```
+
+## Props
+
+### `checked`
```
sohkai: Small API naming nitpick: what do you think about `toggled` or etc.? Not sure how other Switch components are handled, but "checked" seems a bit misleading as there's not really anything to check.

src/components/Switch/Switch.js
bpierre: No alphas are fine :+1: The problem is with non-black colors ([to avoid this kind of things :laughing:](https://github.com/aragon/aragon/pull/92))

src/theme/theme-light.js
bpierre: They already do :-) https://github.com/aragon/aragon-ui/blob/newstyle/src/components/Input/Checkbox.js#L105

src/components/Switch/Switch.js
bpierre: Yes we were discussing it with @adekbadek… it made me realize that the behavior of `Checkbox` is wrong: the enter key should be used to submit forms (if any) when a checkbox is focused, not to toggle the checkbox itself ([reference](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/checkbox_role#Keyboard_interactions)).

The Switch should probably behave the same way as it is a checkbox with a different style. I think we could leave it like this, and fix it in `Checkbox` / `Radio` next (I also want to experiment with a hidden `<input type=checkbox>` on `Checkbox` to align them both, if possible).

src/theme/theme-light.js
adekbadek: No change needed, they all get colors from the theme and updating all control components was intended 👌

src/components/Switch/Switch.js
adekbadek: In pure HTML forms, Enter submits the form - I'd keep this functionality and not hijack Enter on input elements. I was surprised by that too, but I in the end I think it's nice that a form can be submitted by hitting one button. WDYT?

src/components/Switch/README.md
sohkai: > a switch and a checkbox are both boolean value controls

Agreed; let's consolidate on `checked` then :).

PR #625, opened 2019-10-17
useTheme in all components

delfipolito:


src/components/TransactionProgress/TransactionProgress.js
```diff
@@ -10,70 +9,81 @@ import { Info } from '../Info'
...
+          {slow && (
+            <SlowTransaction>
+              <Info.Alert>Slow transaction. Retry with more gas</Info.Alert>
+            </SlowTransaction>
+          )}
+          <Link
+            external
+            href={transactionHashUrl}
+            css={`
+              color: #21aae7;
```
bpierre: We could probably remove this now that `Link` is providing its own color:

```suggestion
```

src/components/Radio/RadioList.js
```diff
@@ -60,11 +61,11 @@ class RadioList extends React.Component {
 }
 
 const Title = styled.h2`
-  margin-bottom: 5px;
+  margin-bottom: ${GU / 2}px;
```
sohkai: In general we try to use `0.5 * GU` for these cases, as it's more similar to css (you can do `0.5px`, but not `px/2`).

src/components/Radio/RadioListItem.js
```diff
@@ -3,22 +3,33 @@ import PropTypes from 'prop-types'
...
+}) {
+  const theme = useTheme()
+  return (
+    <Label>
+      <RadioWrapper id={index} />
+      <LabelBox
+        css={`
+          border: 1px ${theme.border} solid;
+          &:hover {
+            border-color: ${theme.accent.alpha(0.35)};
```
sohkai: @bpierre I think I may have gotten confused, but in the case of borders do we want to find the exact colour instead of using alpha?

bpierre: We generally do, and we also remove the hover states! But this component need to be realigned for aragonDS.

PR #630, opened 2019-10-22
devbox: add “Everything”

bpierre:
This is useful to have an overview of the kit, and specifically to see how the themes are applied.


devbox/apps/index.js
```diff
@@ -20,6 +20,7 @@ export { default as Distribution } from './Distribution'
 export { default as DropDown } from './DropDown'
 export { default as EmptyStateCard } from './EmptyStateCard'
 export { default as EthIdenticon } from './EthIdenticon'
+export { default as Everything } from './Everything'
```
sohkai: Wondering if add an exception for `Everything` to put it on the bottom of the page (rather than alphabetically sorted with everything else)?

bpierre: Yes let’s do that! https://github.com/aragon/aragon-ui/pull/630/commits/44b5e3ec4778c8a186851cf0e4c6550d5359d5bc

PR #631, opened 2019-10-22
TextInput: fix placeholder and readonly colors

bpierre:


src/components/Input/TextInput.js
```diff
@@ -26,22 +26,29 @@ const TextInput = React.forwardRef(({ multiline, type, ...props }, ref) => {
...
           : ''}
 
         &:focus {
           outline: none;
           border-color: ${theme.selected};
         }
+
         &:read-only {
-          color: transparent;
-          text-shadow: 0 0 0 ${theme.surfaceContentSecondary};
```
sohkai: I remember having to do this text-shadow hack quite a long time ago in https://github.com/aragon/aragon-ui/pull/67#discussion_r156247300; do you know if it's not necessary anymore?

src/components/Input/TextInput.js
```diff
@@ -26,22 +26,29 @@ const TextInput = React.forwardRef(({ multiline, type, ...props }, ref) => {
...
           border-color: ${theme.border};
         }
+
+        &::placeholder {
+          color: ${theme.hint};
+          opacity: 1;
+        }
+
         &:invalid {
+          border-color: ${theme.negative};
```
sohkai: I remember dong something similar, but we may have decided to avoid styling it directly here: https://github.com/aragon/aragon-ui/pull/576#discussion_r321010346.

src/components/Input/TextInput.js
bpierre: Oooh I see, I didn’t remember and couldn’t understand why it was there. No I don’t notice the issue, but it might be OS related. Does it work for you on macOS?

src/components/Input/TextInput.js
bpierre: Oh right I forgot about this. Let’s keep it unstyled for now then :+1:

bpierre: Done in https://github.com/aragon/aragon-ui/pull/631/commits/05852170a1fcf3a3c3e7edb1d90890de7addfbcb

src/components/Input/TextInput.js
sohkai: Does not seem to be an issue on macOS's latest Firefox!

It seems to be behaving the same across Chrome, Firefox, and Safari now :).

PR #636, opened 2019-10-27
AutoComplete doc: suggest to limit the number of items

jyash97:
When we have more items on `AutoComplete` the wrapper expands with the items. Adding max-height fixes this. Hope this is a good PR for the system. Here is the screenshot of the UI before and after style changes:

- Before: 
Before

- After
After

### DIFF
--- a/src/components/AutoComplete/AutoComplete.js
+++ b/src/components/AutoComplete/AutoComplete.js
@@ -176,8 +176,6 @@ const Items = React.forwardRef(({ theme, ...props }, ref) => (
       padding: 0;
       margin: 0;
       list-style: none;
-      max-height: 250px;
-      overflow-y: scroll;
 
       & ${Item}:first-child {
         border-top-left-radius: 3px;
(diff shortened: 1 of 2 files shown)


src/components/AutoComplete/AutoComplete.js
```diff
@@ -176,6 +176,8 @@ const Items = React.forwardRef(({ theme, ...props }, ref) => (
       padding: 0;
       margin: 0;
       list-style: none;
+      max-height: 250px;
+      overflow-y: scroll;
```
bpierre: I am not really comfortable with an arbitrary max height: I think we should let authors define this themselves by setting the amount of suggestions directly (e.g. `items={items.slice(0, 10)}`).

For cases where the viewport is smaller than the content, which can always happen, it will get solved by implementing an automatic scroll behavior inside `Popover`, and using it internally for `Autocomplete` (which remain to be done).

Perhaps we could add a note in the `items` documentation, to recommend limiting the items to a certain number?

bpierre: @jyash97 I made the change so we can merge this PR now. We can keep discussing this separately, either here or in a new issue :-)

PR #637, opened 2019-10-27
Checkbox + Radio: update dimensions

jyash97:
Before: 
Before

After:
After

### DIFF
--- a/src/components/Input/Checkbox.js
+++ b/src/components/Input/Checkbox.js
@@ -174,8 +174,8 @@ const Bullet = ({ color }) => (
   <span
     css={`
       display: block;
-      width: 9px;
-      height: 9px;
+      width: 8px;
+      height: 8px;
       border-radius: 50%;
       background: ${color};
     `}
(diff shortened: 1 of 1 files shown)


bpierre:
Hi @jyash97, could you please indicate your browser version? It looks like an issue with subpixel rendering on Blink, which I thought was solved in recent versions. Thanks!

jyash97:
Hey I found this one: Chrome version - 78.0.3904.70 (Official Build) (64-bit)

bpierre:
Hi @jyash97, I could reproduce it! It seems Chrome doesn’t have subpixel rendering on macOS (except when using 3D transforms) and maybe Windows, while it does on the Linux release (which is the one I use :s).

I also noticed that the checkbox and radio dimensions were incorrect: they have been updated in the design system but weren’t up to date in aragonUI. I added a commit to the PR to do this.

Let me know if you notice any other issue!

jyash97:
Hey @bpierre thanks for the update. I am glad that the issue has been fixed :)

If possible you can check #636

PR #648, opened 2019-11-01
Checkbox + Radio: add a disabled state

bpierre:
controls(1)

### DIFF
--- a/src/components/Input/Checkbox.md
+++ b/src/components/Input/Checkbox.md
@@ -33,14 +33,21 @@ class App extends React.Component {
 - Type: `Boolean`
 - Default: `false`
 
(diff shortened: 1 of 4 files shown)


src/components/Input/Checkbox.md
```diff
@@ -33,14 +33,21 @@ class App extends React.Component {
 - Type: `Boolean`
 - Default: `false`
 
-Set to true to check the checkbox.
+Set to `true` to check the checkbox.
+
+### `disabled`
+
+- Type: `Boolean`
+- Default: `false`
```
sohkai: We may as well also convert these to the new table format :)

Issue #629, opened 2019-10-22
tokenIconUrl(): support ETH


sohkai:
The easiest way to do so would be using the zero address (`0x00...00`) and resolving that to https://github.com/trustwallet/assets/blob/master/blockchains/ethereum/info/logo.png.

Otherwise, we could also think of a slightly different API for "native" assets (and eventually, EVM-compatible chains... like Aragon Chain).

Issue #624, opened 2019-10-16
tokenIconUrl(): use https://github.com/trustwallet/assets


bpierre:
https://github.com/TrustWallet/tokens is now deprecated, https://github.com/trustwallet/assets should be used instead.

See https://github.com/aragon/aragon-ui/blob/newstyle/src/utils/web3.js#L144

bpierre:
Fixed by https://github.com/aragon/aragon-ui/pull/651

PR #651, opened 2019-11-04
tokenIconUrl(): support ETH + update source

bpierre:
Fixes https://github.com/aragon/aragon-ui/issues/629 and https://github.com/aragon/aragon-ui/issues/624

I went for the `0x000…` solution for now, I think we could do that for now and support other chains whenever it will become needed.

### DIFF
--- a/src/utils/web3.js
+++ b/src/utils/web3.js
@@ -142,7 +142,7 @@ export function blockExplorerUrl(
 /**
  * Get the address of a token icon
  *
- * @param {string} address The contract address of the token, or an empty address (0x000…) to get the Ethereum icon.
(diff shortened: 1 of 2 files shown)


src/utils/web3.js
```diff
@@ -138,12 +142,19 @@ export function blockExplorerUrl(
 /**
  * Get the address of a token icon
  *
- * @param {string} address The contract address of the token.
+ * @param {string} address The contract address of the token, or an empty address (0x000…) to get the Ethereum icon.
```
sohkai:
```suggestion
 * @param {string} address The contract address of the token, or the zero address (0x000…) to get the Ethereum icon.
```

PR #652, opened 2019-11-05
Use Popover from ContextMenu (WIP)

bpierre:


bpierre:
Closing for now, will reopen after having updated Popover to Popper.js 2.x.

PR #659, opened 2019-11-10
tokenIconUrl(): new source format (checksum addresses)

bpierre:
The format has recently changed on the [source repository](https://github.com/trustwallet/assets/pull/502), and token icons can now only be accessed through checksum addresses.

web3-utils [is huge](https://bundlephobia.com/result?p=web3-utils@1.2.2), so I copied the `toChecksumAddress()` function there and made it use [js-sha3](https://github.com/emn178/js-sha3), which is [quite small](https://bundlephobia.com/result?p=js-sha3@0.8.0) and [used by ethers.js](https://github.com/ethers-io/ethers.js/blob/76a8e503dd47b8da1961fb85a6ca28d007f49585/utils/keccak256.js), so it seems like a reasonable choice for the time being.

@sohkai let me know what you think! Another approach would be to only support checksum addresses for `TokenBadge`.

### DIFF
--- a/src/utils/web3.js
+++ b/src/utils/web3.js
@@ -189,14 +189,8 @@ export function blockExplorerUrl(
  * @return {string} The generated URL, or an empty string if the parameters are invalid.
  */
 export function tokenIconUrl(address = '') {
-  address = address.trim().toLowerCase()
-
-  if (!address) {
-    return ''
-  }
-
   try {
-    address = toChecksumAddress(address)
+    address = toChecksumAddress(address.trim())
   } catch (err) {
     return ''
   }
--- a/src/utils/web3.js
+++ b/src/utils/web3.js
@@ -1,3 +1,4 @@
+import sha3 from 'js-sha3'
 import { warn } from './environment'
 
 const EMPTY_ADDRESS = '0x0000000000000000000000000000000000000000'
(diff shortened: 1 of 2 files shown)


src/utils/web3.js
```diff
@@ -40,6 +41,46 @@ const BLOCK_EXPLORERS = {
...
+
+  if (!/^(0x)?[0-9a-f]{40}$/i.test(address)) {
+    throw new Error(
+      'Given address "' + address + '" is not a valid Ethereum address.'
+    )
+  }
+
+  address = address.toLowerCase().replace(/^0x/i, '')
+
+  const addressHash = sha3.keccak_256(address).replace(/^0x/i, '')
```
sohkai: Perhaps we can just directly import `keccak_256`? I always find it annoying to read `sha3.keccak_256` because keccak256 isn't really sha3.

src/utils/web3.js
```diff
@@ -152,6 +193,12 @@ export function tokenIconUrl(address = '') {
     return ''
   }
 
+  try {
+    address = toChecksumAddress(address)
```
sohkai: We could optimize the code a bit, and just remove the first `address` assignment:

```js
try {
  address = toChecksumAddress(address.trim())
} catch () {
  ...
}
```

src/utils/web3.js
bpierre: It’s not exported, but we can pretend it is :laughing:  https://github.com/aragon/aragon-ui/pull/659/commits/27954241e72281adc57981d0afbfb1c3a05f11f0

src/utils/web3.js
bpierre: Yes! https://github.com/aragon/aragon-ui/pull/659/commits/359f7d532f20a7bb950060923ea1d24f6cf1f306

src/utils/web3.js
```diff
@@ -40,6 +43,46 @@ const BLOCK_EXPLORERS = {
...
+ * https://github.com/ethereum/web3.js/blob/22df832303e349f8ae02f0392e56abe10e1dfaac/packages/web3-utils/src/index.js#L287-L315
+ * And was adapted to use js-sha3 rather than soliditySha3.js from web3.js, in
+ * order to avoid adding the BN.js and underscore dependencies.
+ *
+ * @method toChecksumAddress
+ * @param {String} address the given HEX address
+ * @return {String}
+ */
+function toChecksumAddress(address) {
+  if (typeof address === 'undefined') {
```
sohkai: This first check probably also isn't required, since we'll just throw an error in this case (most likely better than just returning an empty string?)

bpierre: Yes good point, not sure what I was thinking :thinking:

https://github.com/aragon/aragon-ui/pull/659/commits/48646dc06e6e555c5fb9389f34372015eda2c472

PR #661, opened 2019-11-11
Tag: support 0 for numbers passed as labels (without digit limit)

bpierre:


src/components/Tag/Tag.js
```diff
@@ -77,10 +77,10 @@ function getSize(size, { uppercase, discMode, iconAndLabel }) {
   `
 }
 
-function useLabel({ label, limitDigits }) {
+function useLabel({ label = '', limitDigits }) {
   const finalLabel = useMemo(() => {
     if (limitDigits === false) {
-      return label || ''
+      return String(label) || ''
```
sohkai: My only concern is with this; wouldn't this break if you used a node as the `label` prop?

bpierre: It definitely would! :see_no_evil: 

Fixed, thanks! https://github.com/aragon/aragon-ui/pull/661/commits/203a974febb2b36b50b181b9a2869fd4d0e1659d

PR #665, opened 2019-11-19
Deprecate Text and font()

bpierre:
- Deprecate `font()` and `Text`.
- Remove them from any component, example or devbox app.
- `TransactionProgress`: remove `Text` + use `GU` (rather than `rem`).
- Remove `PartitionBar` demo (replaced by `Distribution`).


bpierre:
@sohkai OK I removed all the references to `Text` except from deprecated components using it. I didn’t realize we were still using it so much!

PR #655, opened 2019-11-07
Gallery update for 1.0

bpierre:
- Gallery: style + code refresh.
- Update dependencies.
- Categories refresh (WIP).
- Add sections: Getting Started, Spacing, Colors, Text styles.
- Move the non-component docs into the docs/ directory.
- Add docs for Bar, Box, Header, Split.
- Update icons docs.
- Remove deprecated (or soon to be) components from the gallery.

PR #666, opened 2019-11-19
Deprecate observe()

bpierre:


gallery/src/routes.js
```diff
@@ -58,7 +58,6 @@ import PageButtonBase from './pages/PageButtonBase'
 import PageRedraw from './pages/PageRedraw'
 import PageRoot from './pages/PageRoot'
 import PageRedrawFromDate from './pages/PageRedrawFromDate'
-import PageObserve from './pages/PageObserve'
```
bpierre: I removed the page like from the gallery, to make it like the other deprecated components in 1.0.

After 1.0, we should keep the deprecated components on the gallery, with a deprecation notice.

sohkai: Yep, agreed! I don't even think we really need to keep it in the gallery; the upgrade instructions is good enough.

bpierre:
> Otherwise, let's also add this to the upgrade.md.

Doing this in https://github.com/aragon/aragon-ui/pull/655 :+1:

PR #669, opened 2019-11-21
Gallery dropdown content bug fix & demo enhancement

### LINKED ISSUE #670: Gallery Bug: Content displays on top left when dropdown is clicked
Hey aragon-ui contributors! Due to the 1.0 release (congrats! 🚀) I was checking out the gallery for the new/updated components and noticed a slight bug on the dropdown page: when you click the dropdown, its content is displayed on the top left. It seems to be caused by either the Frame / Resizable components that are rendered if the prop `container` is `true` on the `PageDemo` component. Setting it as `false` and removing the `Container` component on the demo seems to solve the issue. I've made a pull request with the solution I implemented. See #669 

An image of the issue:
Screen Shot 2019-11-20 at 10 51 44 PM

Evalir:
- Set `container` as `false` on the attached Demo component on the PageDropdown component.
- Remove the `Container` component
- Remove the extra div with css for centering
- Pass the `selected` prop to the dropdown for displaying an item.
- Set the default state of the demo to -1 to display the default placeholder.

This PR closes #670

### DIFF
--- a/gallery/src/pages/PageDropDown.js
+++ b/gallery/src/pages/PageDropDown.js
@@ -18,31 +18,19 @@ class PageDropDown extends React.Component {
 
     return (
       <Page title={title} readme={readme}>
-        <Page.Demo height={350}>
-          <Container>
-            <div
-              style={{
-                display: 'flex',
-                height: '100%',
-                width: '100%',
-                paddingTop: '40px',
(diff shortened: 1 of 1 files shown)


Evalir:
No problem! 🦅

PR #673, opened 2019-11-22
Solve RootPortal page bug

### LINKED ISSUE #674: Gallery bug: Root portal page doesn't load
The Root Portal component route in the gallery is not loading; the page stays blank instead but doesn't show any error. 

**How to reproduce**
- Head to the [Aragon UI Page](https://ui.aragon.org/)
- On the advanced part of the components, click the root page component.

Evalir:
- Changed size prop from "compact" (deprecated) to "small". This was making the page not render due to an uncaught error.

This PR closes #674

### DIFF
--- a/gallery/src/pages/PageRootPortal.js
+++ b/gallery/src/pages/PageRootPortal.js
@@ -34,11 +34,7 @@ const PageRootPortal = ({ title }) => {
       >
         This component is rendered {inside ? 'through' : 'without'} RootPortal.
       </div>
-      <Button
-        mode="secondary"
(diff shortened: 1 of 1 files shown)


Evalir:
@bpierre Yeah; it seems to come from the `sizeStyles` function relying on being passed a valid prop to then get the appropriate styles from the `SIZE_STYLES` array. 😃 Maybe a solution would be to introduce a check on the props on the `sizeStyles` function or the `Button` component itself to see if they're valid and if not, set them to a default value.

PR #683, opened 2019-12-10
SyncIndicator: new design

guidovizoso:
- Align FloatIndicator to the right according to Figma designs provided,
- Switch FloatingIndicator's color schemes (from light to dark).
- Modify LoadingRing to display either current or new ring if used from SyncIndicator.
- Add "estimate" prop to SyncIndicator. (no default provided)

image


bpierre:
Hi @guidovizoso, sorry for having taken so long. It’s looking really good!

I added a few commits, mostly to add these:

- Full width mode on small viewports.
- Temporarily hide the indicator when a toast is visible (so they don’t overlap each other).
- Add a `shift` prop (so we can have the help button on the client).

Let me know what you think!

guidovizoso:
@bpierre it's looking great! There's some things I didn't knew because I'm getting familiar with the code at the same time I'm developing but I'll have all of these in mind for my next PR!

src/components/FloatIndicator/README.md
```diff
@@ -33,3 +33,10 @@ Set to `true` to display the indicator.
 | `React node` | None          |
 
 The content of the indicator.
+
+### `shift`
+
+- Type: `Number`
+- Default value: `0`
+
+Use this property to shift the position of the float indicator from its “end position” (right on left to right languages).
```
sohkai:
```suggestion
Use this property to shift the position of the float indicator from its “end position” (the right side for left to right languages).
```

src/components/SyncIndicator/README.md
```diff
@@ -33,3 +33,10 @@ Set to `true` to display the indicator.
 | `String` | `"Syncing data…"` |
 
 The label displayed by the indicator.
+
+### `shift`
```
sohkai: Let's update this to the new table format and add the new props

src/components/SyncIndicator/SyncIndicator.js
```diff
@@ -1,19 +1,36 @@
...
-      <LoadingRing animated />
-      <span css="margin-left: 5px">{label}</span>
+    <FloatIndicator visible={visible} shift={shift} {...props}>
+      <LoadingRing />
+      <div
+        css={`
+          margin-left: ${1.5 * GU}px;
+        `}
+      >
+        {children || `${label} ${estimate ? `Estimate: ${estimate}` : ''} 🙏`}
```
sohkai: I'm just wondering if we really need both of `label` and `estimate` as props.

src/components/LoadingRing/LoadingRing.js
```diff
@@ -12,52 +13,99 @@ const spin = keyframes`
...
+
+  const mode = modeProp || (insideFloatIndicator ? 'half-circle' : 'two-parts')
+
+  const borderWidth =
+    mode === 'half-circle' ? BORDER_WIDTH_SINGLE : BORDER_WIDTH_DOUBLE
+  const size = (mode === 'half-circle' ? SIZE_MEDIUM : SIZE_SMALL) - borderWidth
+  const length = Math.PI * size
+
+  const gapLength = mode === 'two-parts' ? (length / 2) * 0.5 : length / 2
+  const dashLength = mode === 'two-parts' ? length / 2 - gapLength : length / 2
```
sohkai: Small thing, maybe this block would be more readable put inside a `useMemo()` and separated with an if statement at the top rather than ternaries for each variable.

src/components/LoadingRing/LoadingRing.js
```diff
@@ -12,52 +13,99 @@ const spin = keyframes`
...
+  const [insideFloatIndicator] = useInside('FloatIndicator')
+
+  const mode = modeProp || (insideFloatIndicator ? 'half-circle' : 'two-parts')
+
+  const borderWidth =
+    mode === 'half-circle' ? BORDER_WIDTH_SINGLE : BORDER_WIDTH_DOUBLE
+  const size = (mode === 'half-circle' ? SIZE_MEDIUM : SIZE_SMALL) - borderWidth
+  const length = Math.PI * size
+
+  const gapLength = mode === 'two-parts' ? (length / 2) * 0.5 : length / 2
```
sohkai: > (length / 2) * 0.5

Seems a bit redundant 🤷‍♂️

bpierre:
@sohkai 

> The move to using SVG was quite interesting; is it much more performant?

It might be slightly more performant, but the change was really made to fix its rendering and to consolidate the two modes.

The previous one had been hacked together using `overflow: hidden`, which was bothering me since a _long_ time :laughing: :



The new version is using the gradient, and has fixed angles:

image

src/components/SyncIndicator/SyncIndicator.js
bpierre: Yes I was hesitating a bit with this one… I like the idea of keeping the “🙏” as part of the component since it was designed this way. Perhaps we could remove `estimate` and do this:

```
{children || `${label} 🙏`} 
```

Evalir: Yeah, I actually was thinking a bit about the 🙏emoji being there; it looks pretty good and its part of the design. :) Maybe we could also make it optional? And regarding the props: I actually think having just one prop as Pierre says is fine; it gives the component more flexibility.

src/components/SyncIndicator/README.md
bpierre: Done in https://github.com/aragon/aragon-ui/pull/683/commits/ae1a667473e4664c2c8e1a94c77ef0da487b08ee

src/components/LoadingRing/LoadingRing.js
bpierre: Done (and simplified!) in https://github.com/aragon/aragon-ui/pull/683/commits/d335cd551c96f89c31491fabdb522d39d0281fec

src/components/LoadingRing/LoadingRing.js
bpierre: Done https://github.com/aragon/aragon-ui/commit/d335cd551c96f89c31491fabdb522d39d0281fec

src/components/SyncIndicator/SyncIndicator.js
guidovizoso: I'll get into it. Is adding a prop called "praise" a bit too much?

bpierre: @guidovizoso Yes I think `praise` could become problematic if we modify the emoji one day. Another option would be to name the prop `adornment` and make it default to `"🙏"`.

But since setting a `label` is a very common usage of this component, removing or setting the emoji to something else could be as simple as moving from `label` to `children`:

```jsx
// Default emoji
<SyncIndicator
  label="Fetching data… Estimate: 3 hours"
/>

// No emoji
<SyncIndicator>
  Fetching data… Estimate: 3 hours
</SyncIndicator>

// Custom emoji
<SyncIndicator>
  Fetching data… Estimate: 3 hours 💪
</SyncIndicator>
```

bpierre:
Thank you @guidovizoso :100:

I added one last update to use `children` by default, all good otherwise, merging! :+1:

PR #688, opened 2019-12-18
Move to yarn.lock + update dependencies

bpierre:
- Update dependencies (fix a security alert).
- Remove WebpackMonitor (unmaintained, little benefit on the gallery).
- Replace the webpack UglifyJS plugin by the Terser plugin (the UglifyJS plugin is now unmaintained).
- Move to yarn.lock files.
- Prettier: formatting change.


package.json
```diff
@@ -68,16 +68,16 @@
...
     "rollup-plugin-commonjs": "^10.0.0",
     "rollup-plugin-filesize": "^6.0.1",
     "rollup-plugin-node-resolve": "^5.0.0",
     "rollup-plugin-progress": "^1.1.1",
-    "rollup-plugin-sizes": "^0.5.1",
-    "rollup-plugin-uglify": "^6.0.0",
-    "rollup-plugin-url": "^2.0.1",
-    "rollup-plugin-visualizer": "^1.1.1",
+    "rollup-plugin-sizes": "^1.0.1",
+    "rollup-plugin-uglify": "^6.0.4",
```
sohkai: Should we also move to the terser rollup plugin?

package.json
```diff
@@ -68,16 +68,16 @@
...
+    "rollup-plugin-url": "^3.0.1",
```
sohkai: It looks like there are also deprecation warnings for `rollup-plugin-url` and `rollup-plugin-node-resolve`; should we update those as well?

package.json
bpierre: Ah this isn’t used anymore, I think we can remove it. We shouldn’t need to compress aragonUI since it will get bundled by apps and compressed at this stage. It also makes it easier to inspect if the tool importing aragonUI doesn’t support sourcemaps.

package.json
bpierre: Oh yes I didn’t see that! I thought `yarn outdated` or `yarn upgrade-interactive` would show that, but it seems that no command shows the deprecated status of dependencies − apart from running `yarn info | grep deprecated` on them individually… are you doing this manually, or do you know a secret command? :smile:

sohkai: > are you doing this manually, or do you know a secret command?

I just see the warnings during install :).

PR #699, opened 2020-01-03
Support for non-DOM environments (Next.js)

izqui:


izqui:
Thanks @bpierre!

izqui:
@welcome it was really all Pierre's work at the end but thanks anyway :)

PR #700, opened 2020-01-06
Handle auto value through the rowHeight prop on the CardLayout component

### LINKED ISSUE #635: CardLayout: auto height
See discussion about it: https://github.com/aragon/aragon-ui/commit/b2d1da7f7cce7ed3dd7cf843296c41ac13f67100#r35598639

/cc @chadoh

Evalir:
- Adds the ability to set `"auto"` on the `rowHeight` prop, handling the CSS properly instead of setting auto values in a hacky way for the card.

Closes #635 , discussion [here.](https://github.com/aragon/aragon-ui/commit/b2d1da7f7cce7ed3dd7cf843296c41ac13f67100#r35625086)

### DIFF
--- a/src/components/CardLayout/CardLayout.js
+++ b/src/components/CardLayout/CardLayout.js
@@ -7,6 +7,7 @@ import { Inside } from '../../utils/inside'
 function CardLayout({ children, columnWidthMin, rowHeight, ...props }) {
   const { layoutName } = useLayout()
   const fullWidth = layoutName === 'small'
+  const gridAutoRowValue = rowHeight === 'auto' ? rowHeight : `${rowHeight}px`
 
   return (
     <Inside name="CardLayout">
@@ -19,7 +20,7 @@ function CardLayout({ children, columnWidthMin, rowHeight, ...props }) {
             ${fullWidth ? 'auto-fit' : 'auto-fill'},
             minmax(${columnWidthMin}px, 1fr)
           );
-          grid-auto-rows: ${rowHeight}px;
+          grid-auto-rows: ${gridAutoRowValue};
           align-items: start;
           padding: 0 ${fullWidth ? 2 * GU : 0}px ${3 * GU}px;
           margin: 0 auto;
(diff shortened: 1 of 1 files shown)


src/components/CardLayout/CardLayout.js
```diff
@@ -7,6 +7,7 @@ import { Inside } from '../../utils/inside'
 function CardLayout({ children, columnWidthMin, rowHeight, ...props }) {
   const { layoutName } = useLayout()
   const fullWidth = layoutName === 'small'
+  const gridAutoRowValue = rowHeight === 'auto' ? rowHeight : `${rowHeight}px`
```
bpierre: Shouldn’t it be `auto`?

```suggestion
  const gridAutoRowValue = rowHeight === 'auto' ? 'auto' : `${rowHeight}px`
```

src/components/CardLayout/CardLayout.js
```diff
@@ -35,7 +36,7 @@ function CardLayout({ children, columnWidthMin, rowHeight, ...props }) {
 CardLayout.propTypes = {
   children: PropTypes.node,
   columnWidthMin: PropTypes.number,
-  rowHeight: PropTypes.number,
+  rowHeight: PropTypes.oneOfType([PropTypes.string, PropTypes.number]),
```
bpierre: I think we can do this to validate `auto` and only `auto` :smile: 

```suggestion
  rowHeight: PropTypes.oneOfType([PropTypes.oneOf(['auto']), PropTypes.number]),
```

src/components/CardLayout/CardLayout.js
Evalir: Yes; my thought process was that if `rowHeight` is indeed equal to 'auto', then we should "leave it the same", instead of applying string interpolation. However, it is more clear to set it to auto.

src/components/CardLayout/CardLayout.js
Evalir: Nice!

src/components/CardLayout/CardLayout.js
bpierre: Oh right! Sorry for some reason I thought that it was using the default number alone rather of `auto` by looking at it. All good, let’s keep it the way you did it!

PR #701, opened 2020-01-06
Deprecate Countdown, replace all internal Countdown instances with Timer

### LINKED ISSUE #687: Deprecate CountDown
`CountDown` should get deprecated and replaced by [`Timer`](https://github.com/aragon/aragon-ui/blob/master/src/components/Timer/Timer.js)

Evalir:
- Adds a deprecation warning for the `Countdown component`
- Switches `TransactionProgress` to use `Timer` instead of `Countdown`.

Closes #687


src/components/Countdown/Countdown.js
```diff
@@ -1,9 +1,18 @@
 import React from 'react'
 import PropTypes from 'prop-types'
 import Timer from '../Timer/Timer'
+import { warnOnce } from '../../utils'
 
 class Countdown extends React.Component {
+  deprecationWarning() {
+    warnOnce(
+      'Countdown',
+      '"Countdown" and its variants have been deprecated. Please use "Timer" instead.'
```
bpierre: I might be wrong but I don’t think we have any variant?

```suggestion
      '"Countdown" have been deprecated. Please use "Timer" instead.'
```

Evalir: Yep, we actually don't.

Evalir:
Should be good to go! 🚀

PR #704, opened 2020-01-14
LineChart: function component + full width by default

bpierre:


src/components/LineChart/LineChart.js
```diff
@@ -1,236 +1,230 @@
...
+  labelColor: '#6d777b',
+  lines: [],
+  label: index => index + 1,
+  color: (index, { lines }) =>
+    `hsl(${(index * (360 / lines.length) + 40) % 360}, 60%, 70%)`,
+}
+
+export default props => {
+  const { width } = useViewport()
+  return <LineChart vw={width} {...props} />
```
Evalir: What is the `vw` prop used for?

izqui:
@evalir it is 'aragonUI' not 'Aragon UI' any more ;)

src/components/LineChart/README.md
```diff
@@ -49,41 +50,57 @@ function App() {
 - Type: `Number`
 - Default: `-1`
 
+The expected total (x axis of the chart). When not provided, the length of `values` is used instead.
+
 ### `width`
 
 - Type: `Number`
 - Default: `300`
```
izqui: Change this?

src/components/LineChart/LineChart.js
bpierre: Oops, just something I was trying, I removed it!

bpierre:
@izqui I was going to say that you should use `width` to adapt the initial rendering, but then I thought it would probably be better to do it automatically. It should work now, without having to set `width`.

src/components/LineChart/README.md
bpierre: Updated yes!

src/components/LineChart/README.md
```diff
@@ -94,10 +111,28 @@ Pass the values that will be used to draw the lines. The color atribute in Subty
...
+
+Example:
+
+```jsx
+<LineChart label={index => ((index % 26) + 10).toString(36)} />
+```
+
+This function gets called to render a label on the x axis.
 
 ### `color`
```
sohkai: A bit of a nitpick but we might've wanted to take the chance to also update the README to use the new table format :).

PR #395, opened 2019-06-15
Markdown rendering / Normalized Html styling

corydickson:
Addresses issue #386 

I had to update the `rollup` config with a JSON plugin due to a failed installation. Let me know if you are able to reproduce.

Also since we are using `jsx` everywhere I don't think we'll need to sanitize the input.

PR #709, opened 2020-01-30
Tree-shake ready

bpierre:
This optimization will make it possible to include components like [Markdown](https://github.com/aragon/aragon-ui/pull/395) without affecting the performance of apps that are not using it.

Some numbers:

| App                       | Before    | After     | Decrease           | Notes                                                                                                                                                                                              |
| ------------------------- | --------- | --------- | ------------------ | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| Basic app (Main + Button) | 144.21 KB | 76.99 KB  | 46.6%              |                                                                                                                                                                                                    |
| aragon/aragon (Client)    | 1040 KB   | 906.79 KB | 12.8% 🧐           |                                                                                                                                                                                                    |
| Finance                   | 390 KB    | 307 KB    | 21.3% :open_mouth: |                                                                                                                                                                                                    |
| Tokens                    | 348.15 KB | 196.73 KB | 43.5% :scream:     | Also removed [web3-utils](https://bundlephobia.com/result?p=web3-utils@1.2.6), which was breaking with Parcel’s tree shaking. [See pull request](https://github.com/aragon/aragon-apps/pull/1076). |


src/components/index.js
```diff
@@ -1,17 +1,10 @@
...
-export { DataView } from './DataView/DataView'
-export { DiscButton } from './DiscButton/DiscButton'
-export { Info } from './Info'
-export { Layout, useLayout } from './Layout/Layout'
-export { Link, ExternalLink, SafeLink } from './Link'
-export { Pagination } from './Pagination/Pagination'
-export { SearchInput } from './Input/SearchInput'
-export { Tag } from './Tag/Tag'
+export { ExternalLink, SafeLink } from './Link/LinkDeprecated'
 export { ToastHub, Toast, useToast } from './ToastHub/ToastHub'
```
Evalir: Should we move this down to the corresponding part of the file?

src/theme-legacy/index.js
```diff
@@ -1 +1 @@
-export { theme as default, theme, themeDark, brand, colors } from './palettes'
+export { theme, themeDark, brand, colors } from './palettes'
```
sohkai: Is this a breaking change?

rollup.config.js
```diff
@@ -1,62 +1,81 @@
...
+    input: {
+      index: 'src/index.js',
+      components: 'src/components/index.js',
+      hooks: 'src/hooks/index.js',
+      providers: 'src/providers/index.js',
+      icons: 'src/icons/index.js',
+      themeLegacy: 'src/theme-legacy/index.js',
+      theme: 'src/theme/index.js',
+      utils: 'src/utils/index.js',
+      style: 'src/style/index.js',
```
sohkai: Wondering if we should also use `glob` to make this list; do you think we'll easily forget to update it in the future?

src/components/index.js
bpierre: Done in https://github.com/aragon/aragon-ui/pull/709/commits/1aa7a5795ee3e8faa9fdfd0e3c2688f6d00e9b8e and https://github.com/aragon/aragon-ui/pull/709/commits/cdecbed342e623559a5fc762d3e20c4020bbd241

src/theme-legacy/index.js
bpierre: No it was only for internal imports, `default` was never exported: https://github.com/aragon/aragon-ui/blob/master/src/index.js#L5

rollup.config.js
bpierre: Yes I think we can do this. We’ll have to be careful if one day we add a private `dir/index.js` here, but we are not doing it right now.

rollup.config.js
```diff
@@ -50,10 +60,26 @@ export default (async () => {
...
+        exports: 'named',
       },
     ],
     external: ['react', 'react-dom', 'styled-components'],
-    treeshake: true,
+    treeshake: production
+      ? {
+          moduleSideEffects: false,
+
+          // Reminder: getters are going to be used to deprecate theme-legacy, so
```
sohkai: Maybe we can add a TODO or a companion comment in `theme.js` to remember to do this?

PR #710, opened 2020-01-31
Popover: reject closing

bpierre:
I’m wondering if a more transparent approach could be to wait for the next tick, then focus again if `visible` is still `true`.

### DIFF
--- a/src/components/Popover/Popover.js
+++ b/src/components/Popover/Popover.js
@@ -47,7 +47,7 @@ class PopoverBase extends React.Component {
   componentDidMount() {
     this._document = this._popperElement.current.ownerDocument
     this._document.addEventListener('keydown', this.handleEscape)
-    this._cardElement.current.focus()
+    this.focus()
     this.initPopper()
   }
 
@@ -67,6 +67,12 @@ class PopoverBase extends React.Component {
     ) {
       this.destroyPopper()
       this.initPopper()
+    }
+  }
+
+  focus() {
+    if (this._cardElement.current) {
+      this._cardElement.current.focus()
     }
   }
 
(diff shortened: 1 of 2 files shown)


bpierre:
> Not sure on the context, but wouldn't it be weird if there was a small lag between the `onClose()` and re-focus?

It shouldn’t be perceptible for the user, it would be just waiting for the component to update (or not) its `visible` prop. It also only addresses the case where the popover doesn’t close after losing focus, so even with a lag, it wouldn’t be visible.

I think a safer way to do it than `setTimeout(0)` would be to rely on React:

```jsx
  componentDidUpdate(prevProps, prevState) {
    // …

    if (prevState.closeAttempts !== this.state.closeAttempts) {
      this.focus()
    }
  }

  attemptClose() {
    this.props.onClose()
    this.setState(state => ({ closeAttempts: state.closeAttempts + 1 })
  }
```

sohkai:
> It shouldn’t be perceptible for the user, it would be just waiting for the component to update (or not) its visible prop.

I guess the follow-up would be whether there's something we're afraid of with calling `.focus()` immediately? That the parent component may want to update state in `onClose()` and if we call `.focus()` immediately, we may miss new props?

bpierre:
> I guess the follow-up would be whether there's something we're afraid of with calling .focus() immediately? That the parent component may want to update state in onClose() and if we call .focus() immediately, we may miss new props?

It’s more about losing the newly focused element if the `onClose()` gets accepted. If we refocus immediately, the focus sequence should look like this:

- Focus is on the opened popover.
- The focus moves to an element outside of the popover, which triggers the `onClose()`.
- Focus on the popover again (assuming we are doing the immediate focus).
- `onClose` gets accepted (the `visible` prop updates).
- The Popover element gets removed at the end of the transition.
- The focus is lost, back to `document.body` ☹️

So the idea was to wait for the `visible` prop to propagate to a new render, also using something like `closeAttempts` so we get a new render even if the `visible` prop doesn’t change. Then, we can decide to focus again or not, depending on the `visible` state.

Another option would be to save `document.activeElement`, focus immediately, and if `visible` gets updated, move the focus back to where it was. That should work too, but it would move the focus unnecessarily.

PR #720, opened 2020-02-19
JSDOM (Jest) compatibility

bpierre:


src/components/Slider/Slider.js
```diff
@@ -13,8 +13,10 @@ const PADDING = 5
 const MIN_WIDTH = HANDLE_SIZE * 10
 const HEIGHT = Math.max(HANDLE_SIZE, BAR_HEIGHT) + PADDING * 2
 
+// The check on window.DOMRect is needed for the JSDOM environment, which has
+// window but not window.DOMRect. JSDOM is used by default in Jest.
 const DEFAULT_RECT =
-  typeof window === 'undefined'
+  typeof window === 'undefined' || typeof window.DOMRect === 'undefined'
```
sohkai:
```suggestion
  typeof window === 'undefined' || typeof window.DOMRect !== 'function'
```

Perhaps this would be the most safe?

bpierre: Yes!

PR #722, opened 2020-02-20
Dataview empty states

rperez89:
Based on this API proposal: https://gist.github.com/bpierre/97c2dda99a8151cb144170e3e681ce09

This PR introduces the `emptyState` prop and deprecates `statusEmpty`, `statusEmptyFilters`, `statusEmptySearch`, `statusLoading`.

### emptyState

Use this prop to customize the empty state of `DataView`. The type of empty state currently active is determined by the `status` prop.

This prop accepts two types: a configuration object or a function.

#### Configuration object

The object can define one of the possible values of `status` as keys, and an object.

This object can contain the following values:

- `title`, `subtitle` and `illustration` are React nodes that can override any of these parts. Set to `null` to disable.
- `displayLoader` is a boolean that can be used to display a loader before the title.
- `clearLabel` is the content of the clearing link, and can contain any React node. Set to `null` to disable. See also `onStatusEmptyClear`.

Any undefined `status`, or undefined key on that object will use the default. `null` can be passed to disable a specific part of the empty state.

The default values are the following:

```jsx
<DataView
  emptyState={{
    default: {
      displayLoader: false,
      title: 'No data available.',
      subtitle: null,
      illustration: <img src="empty-state-illustration-blue.png" alt="" />,
      clearLabel: null,
    },
    loading: {
      displayLoader: true,
      title: 'No data available.',
      subtitle: null,
      illustration: <img src="empty-state-illustration-blue.png" alt="" />,
      clearLabel: null,
    },
    'empty-filters': {
      displayLoader: false,
      title: 'No results found.',
      subtitle: 'We can’t find any item matching your filter selection.',
      illustration: <img src="empty-state-illustration-red.png" alt="" />,
      clearLabel: 'Clear filters',
    },
    'empty-search': {
      displayLoader: false,
      title: 'No results found.',
      subtitle: 'We can’t find any item matching your search query.',
      illustration: <img src="empty-state-illustration-red.png" alt="" />,
      clearLabel: 'Clear filters',
    },
  }}
/>
```

#### Function mode

The function allows to completely override the empty state content. It takes the current `status` as a unique parameter, and is expected to return a React node, or `null` to let DataView display the default.

Use it this way:

```jsx
<DataView
  emptyState={status => {
    if (status === 'loading') {
      return <div>Loading!</div>
    }
    // Use the default for other `status` values.
    return null
  }}
/>
```

### DIFF
--- a/src/components/DataView/DataView.js
+++ b/src/components/DataView/DataView.js
@@ -241,6 +241,36 @@ function useSelection(entries, selection, onSelectEntries) {
   }
 }
 
+function useEmptyStateValue(
+  status,
+  defaultEmptyStates,
+  emptyStateConfigurator
+) {
+  if (!emptyStateConfigurator) {
+    return defaultEmptyStates[status]
+  }
+
+  const functionMode = typeof emptyStateConfigurator === 'function'
+  const customEmptyState = functionMode
+    ? emptyStateConfigurator(status)
+    : emptyStateConfigurator[status]
+
+  // text shortcut for object mode
+  if (!functionMode && React.isValidElement(customEmptyState)) {
+    return { ...defaultEmptyStates[status], text: customEmptyState }
+  }
+
+  // override all (illustration and text) with function mode
+  if (React.isValidElement(customEmptyState)) {
+    return customEmptyState
+  }
+
+  return {
+    ...defaultEmptyStates[status],
+    ...customEmptyState,
+  }
+}
+
 const DataView = React.memo(function DataView({
   page,
   entries,
@@ -314,49 +344,30 @@ const DataView = React.memo(function DataView({
   const { layoutName } = useLayout()
   const publicUrl = usePublicUrl()
 
-  const emptyStateNode = useCallback(
-    status => {
-      let dataviewStates = prepareDefaultStates(
-        theme,
-        publicUrl,
-        statusEmpty,
-        statusLoading,
-        statusEmptyFilters,
-        onStatusEmptyClear,
-        statusEmptySearch
-      )
-      if (emptyStateConfigurator) {
-        const customEmptyState = emptyStateConfigurator(status)
-
-        if (React.isValidElement(customEmptyState)) {
-          dataviewStates[status] = customEmptyState
-        } else {
-          dataviewStates[status] = {
-            ...dataviewStates[status],
-            ...customEmptyState,
-          }
-        }
-      }
-
-      return (
-        <DefaultEmptyStateNode
-          status={status}
-          publicUrl={publicUrl}
-          states={dataviewStates}
-          onStatusEmptyClear={onStatusEmptyClear}
-        />
-      )
-    },
-    [
-      emptyStateConfigurator,
-      onStatusEmptyClear,
+  const defaultEmptyStates = useMemo(() => {
+    return prepareDefaultStates(
+      theme,
       publicUrl,
       statusEmpty,
-      statusEmptyFilters,
-      statusEmptySearch,
       statusLoading,
-      theme,
-    ]
+      statusEmptyFilters,
+      onStatusEmptyClear,
+      statusEmptySearch
+    )
+  }, [
+    onStatusEmptyClear,
+    publicUrl,
+    statusEmpty,
+    statusEmptyFilters,
+    statusEmptySearch,
+    statusLoading,
+    theme,
+  ])
+
+  const emptyStateValue = useEmptyStateValue(
+    status,
+    defaultEmptyStates,
+    emptyStateConfigurator
   )
 
   const listMode =
@@ -473,7 +484,7 @@ const DataView = React.memo(function DataView({
               padding: ${8 * GU}px 0;
             `}
           >
-            {emptyStateNode(status)}
+            <EmptyStateNode status={status} emptyState={emptyStateValue} />
           </div>
         </div>
       )}
@@ -496,22 +507,21 @@ const DataView = React.memo(function DataView({
   )
 })
 
-function DefaultEmptyStateNode({ status, publicUrl, states }) {
-  // const theme = useTheme()
-  if (React.isValidElement(states[status])) {
-    return states[status]
+function EmptyStateNode({ status, emptyState }) {
+  if (React.isValidElement(emptyState)) {
+    return emptyState
   }
 
   return (
     <>
       {(() => {
         // Empty state: illustration part
-        if (React.isValidElement(states[status].illustration)) {
-          return states[status].illustration
+        if (React.isValidElement(emptyState.illustration)) {
+          return emptyState.illustration
         }
         return (
           <img
-            src={states[status].illustration}
+            src={emptyState.illustration}
             alt=""
             height={20 * GU}
             css={`
@@ -534,11 +544,11 @@ function DefaultEmptyStateNode({ status, publicUrl, states }) {
               >
                 No results found.
               </p>
-              {states[status].text}
+              {emptyState.text}
             </>
           )
         }
-        return states[status].text
+        return emptyState.text
       })()}
     </>
   )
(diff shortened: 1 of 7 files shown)


src/components/DataView/DataView.js
```diff
@@ -234,6 +236,25 @@ const DataView = React.memo(function DataView({
...
+      statusLoading={statusLoading}
+      statusEmptyFilters={statusEmptyFilters}
+      statusEmptySearch={statusEmptySearch}
+      onStatusEmptyClear={onStatusEmptyClear}
+    />
+  )
+
+  const emptyStateNode = useCallback(
+    (status, defaultEmptyState) => {
+      return emptyStates(status) || defaultEmptyState
```
bpierre: I like this better than the solution of passing the default node as a second parameter! And I think we could accept any falsy value, I don’t think there is any reason to not render any empty state. Even if someone wants to do this, it is would still be possible to return an empty fragment from `emptyStates()`.

Now we need to add the transformation from the value being returned by `emptyState()` into the final element. I think a good idea could be to have the default values being declared as `{ illustration, text }` objects, override them with the value returned by `emptyState()`, then pass them to the empty state component.

What do you think?

src/components/DataView/DataView.js
```diff
@@ -185,6 +185,7 @@ const DataView = React.memo(function DataView({
   statusEmptyFilters,
   statusEmptySearch,
   onStatusEmptyClear,
+  emptyStates,
```
bpierre: What would you think about naming this `emptyState`, as in “empty state configurator / generator”?

src/components/DataView/DataView.js
```diff
@@ -163,6 +241,36 @@ function useSelection(entries, selection, onSelectEntries) {
...
+    return defaultEmptyStates[status]
+  }
+
+  const functionMode = typeof emptyStateConfigurator === 'function'
+  const customEmptyState = functionMode
+    ? emptyStateConfigurator(status)
+    : emptyStateConfigurator[status]
+
+  // text shortcut for object mode
+  if (!functionMode && React.isValidElement(customEmptyState)) {
```
bpierre: We should also test if `typeof customEmptyState === 'string'` (`React.isValidElement()` doesn’t check for other accepted nodes like strings or `null`).

src/components/DataView/DataView.js
```diff
@@ -14,6 +14,84 @@ import { ListView } from './ListView'
...
+            display: flex;
+            align-items: center;
+          `}
+        >
+          <LoadingRing
+            css={`
+              margin-right: ${2 * GU}px;
+            `}
+          />{' '}
+          Loading data…
```
bpierre: I think `statusLoading.text` should be set here directly, and the default structure (`<p>`, `<LoadingRing>`) could be separated from the default state. This structure would only get overridden by the function mode, otherwise the structure would always get used.

src/components/DataView/DataView.js
```diff
@@ -14,6 +14,84 @@ import { ListView } from './ListView'
...
+  statusEmpty,
+  statusLoading,
+  statusEmptyFilters,
+  onStatusEmptyClear,
+  statusEmptySearch
+) {
+  return {
+    default: {
+      illustration: publicUrl + illustrationBlueImage,
+      text: statusEmpty || (
```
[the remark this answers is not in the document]
rperez89: sure!

src/components/DataView/DataView.js
```diff
@@ -234,6 +344,32 @@ const DataView = React.memo(function DataView({
...
+    statusLoading,
+    theme,
+  ])
+
+  const emptyStateValue = useEmptyStateValue(
+    status,
+    defaultEmptyStates,
+    emptyStateConfigurator
+  )
+
```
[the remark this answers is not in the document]
rperez89: agreed on moving it to its own file, but i thought about this that you were saying to move the logic also inside the file and right now is super simple it just receive the status and what needs to be displayed, do you think that might be better to pass down to it all the others parameters like `publicUrl, statusEmpty, statusLoading, statusEmptyFilters, etc` ? in my understanding is simpler to send only the data processed and that's it , but i might be missing something

src/components/DataView/DataView.js
rperez89: not sure if i understood your idea here since the previous behavior was with the loading ring only for the default mode
```  if (status === 'loading') {
                return (
                  statusLoading || (
                    <p
                      css={`
                        ${textStyle('title2')};
                        display: flex;
                        align-items: center;
                      `}
                    >
                      <LoadingRing
                        css={`
                          margin-right: ${2 * GU}px;
                        `}
                      />{' '}
                      Loading data…
                    </p>
                  )
                )
              }
```

src/components/DataView/DataView.js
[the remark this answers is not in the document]
bpierre: Yes I think it could be nice to move `useEmptyStateValue()` and `prepareDefaultStates()` there, so that everything related to the empty state is together?

src/components/DataView/DataView.js
bpierre: I’m thinking about making our internal default object identical to the object users would set, and removing the HTML structure from it:

```jsx
function prepareDefaultStates(
  publicUrl,
  statusEmpty,
  statusLoading,
  statusEmptyFilters,
  statusEmptySearch
) {
  return {
    default: {
      illustration: publicUrl + illustrationBlueImage,
      description: statusEmpty || 'No data available.',
    },
    loading: {
      illustration: publicUrl + illustrationBlueImage,
      description: statusLoading || 'Loading data…',
    },
    'empty-filters': {
      illustration: publicUrl + illustrationRedImage,
      description: statusEmptyFilters || 'We can’t find any item matching your filter selection.',
    },
    'empty-search': {
      illustration: publicUrl + illustrationRedImage,
      description: statusEmptySearch || 'We can’t find any item matching your search query.',
    },
  }
}
```

Then we could have this structure defined in the `EmptyState` component directly. But if users pass a function to `emptyState` (`functionMode === true`), then the entire structure of the empty state would be skipped.

src/components/DataView/DataView.js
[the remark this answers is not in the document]
rperez89: yes that also make sense

src/components/DataView/EmptyState.js
```diff
@@ -127,93 +81,65 @@ function EmptyState({
...
+        {emptyState.title && (
+          <h1
+            css={`
+              ${textStyle('title2')};
+              display: flex;
+              align-items: center;
+              justify-content: center;
+            `}
+          >
+            {emptyState.displayLoader && (
```
rperez89: here we are allowing the user to introduce a bad configuration, i.e displaying an empty state result with a loading indicator, thats why i left the status check for this case.

src/components/DataView/EmptyState.js
```diff
@@ -5,119 +5,73 @@ import { GU, textStyle } from '../../style'
...
+    // eslint-disable-next-line react/prop-types
+    const Illustration = ({ path }) => (
+      <img src={publicUrl + path} alt="" height={20 * GU} />
+    )
+    return {
+      default: {
+        displayLoader: false,
+        title: 'No data available.',
+        subtitle: null,
+        illustration: <Illustration path={illustrationBlueImage} />,
```
rperez89: which is the Pro about  setting the illustration here instead of how we were doing it before?

src/components/DataView/EmptyState.js
bpierre: Ah yes true, but I think we could let this be the responsibility on authors at this point? They shouldn’t have to touch this flag, but they can if they want.

Also I’m thinking that we could add `displayClearAction` for consistency, rather than implicitly relying on `clearLabel` to be set?

src/components/DataView/EmptyState.js
bpierre: It is to remove the possibility to set the illustration as a string, to remove a case in the API, and because users shouldn’t really need to change the image. I commented about it here https://github.com/aragon/aragon-ui/pull/722#discussion_r389027236

rperez89:
Cooler, LGTM

PR #726, opened 2020-03-02
CI: Add codesandbox CI

Evalir:
Adds a CI pipeline for having automated deployments of the devbox with the latest changes on every PR.


Evalir:
Will close this for now; sadly, Codesandbox CI seems not to like huge builds and will get stuck if it faces too much load. We will revisit later when the tooling has matured.

PR #727, opened 2020-03-03
Dependencies: Add use-inside as a dependency and remove the local utils file

### LINKED ISSUE #723: Utils: Extract useInside() to its own library
As discussed offline between @bpierre and me, it would be a good idea to extract `useInside` to it's own library, as it presents a good solution for tracking components inside a specific tree and providing useful data through the hook. [Implementation here.](https://github.com/aragon/aragon-ui/blob/master/src/utils/inside.js)

Todos:
- [x] Extract it to its own package - [done 🎉 ](https://github.com/aragon/use-inside)
- [ ] Add `use-inside` as a dependency to aragonUI

Evalir:
Closes #723 , Adding `use-inside` as a dependency rather than having the library locally in the `utils/` folder.


src/icons/components/IconSwap.js
```diff
@@ -0,0 +1,32 @@
+import React from 'react'
```
sohkai: Not sure if this is a new icon, but let's move it out of this PR if it is.

.gitignore
```diff
@@ -5,6 +5,8 @@ npm-debug.log
 package-lock.json
 yarn-error.log
 /stats.html
-
+src/icons/.DS_Store
+.DS_Store
+src/.DS_Store
```
bpierre: These are not specific to the project, they should be declared [in local .gitignore files](https://help.github.com/en/github/using-git/ignoring-files#configuring-ignored-files-for-all-repositories-on-your-computer).

gallery/yarn.lock
```diff
@@ -3,7 +3,7 @@
```
bpierre: I think this will have to be updated after the next release, as it doesn’t contain `use-inside` otherwise.

Evalir: True!

Issue #719, opened 2020-02-14
Add number formatting utilities


sohkai:
From an offline discussion:

- `formatNumber()`: use `Intl.formatNumber()`
- `formatTokenAmount()`: build on top of `formatNumber()` and add some options to make it easier to format token amounts (e.g. see [this function](https://github.com/aragon/aragon-apps/blob/2ced965a1be54fb38809e332eba5b39322f5c32c/future-apps/payroll/app/src/utils/formatting.js#L12-L33))

ECWireless:
Should the options be the exact same as the payroll formatting options (formatCurrency) you gave in the example, or are there specific options you have in mind for the UI formatTokenAmount util?

bpierre:
@ECWireless I think we could adapt them to make it a bit simpler to use:

#### `pow`, `multiplier` and `decimals`
I think we could replace these three params by a single `decimals = 18`, since it is both defined by ERC-20 and widely used. Similarly to `amount`, this should be either a number, or any object with a `toNumber()` method (i.e. we should call `Number()` on it).

#### `amount`
This should be either a string, a `BigInt` instance, or any object with a .toString() method (i.e. we should call `String()` on it).

#### `rounding`
I think we could rename this to something like `displayDecimals`, `roundToDecimals`, `decimalPlaces`, or just `precision`?

#### `isIncoming`
I don’t think this is needed, we can use a sign on the BigInt + `displaySign` (or `signDisplay`) to determine that.

#### `displaySign`
We might want to name it `signDisplay` to align it with [the new `Intl.NumberFormat` API](https://github.com/tc39/proposal-unified-intl-numberformat#iii-sign-display).

<hr>

@sohkai @ECWireless  what would you guys think about naming the function `formatAmount()`? It might sound a bit generic, but I think there is enough context for authors to know what it is about. It would also make it clearer that we can use it to format currencies like USD as well.

ECWireless:
Okay, most of that makes sense to me. The only thing I'm fuzzy on is: if `pow` and `multiplier` are being removed, couldn't decimals and rounding be combined into a simple `precision` option? (Since they essentially accomplish the same thing?)

bpierre:
@ECWireless the two would be different. The idea would be to pass a token amount represented as an integer, requiring to indicate its number of decimals. Then, for display reasons, we want to round this number of decimals to a certain number.

[...]

PR #728, opened 2020-03-04
Add number formatting utilities

ECWireless:
Adding a `formatNumber()` and `formatTokenAmount()` utility for issue #719


ECWireless:
Wondering if there are any other changes I need to make?

src/utils/format.js
```diff
@@ -13,3 +15,41 @@ export const formatIntegerRange = (
...
+ * @param {number} returnToDecimals Rounds the number to a given decimal place
+ * @param {boolean} isIncoming Decides if a plus or negative should be used
+ * @param {boolean} signDisplay Decides if the sign should be displayed
+ * @returns {string} Formatted token
+ */
+export function formatTokenAmount(
+  amount,
+  symbol,
+  decimals = 18,
+  rountToDecimals = 2,
```
Evalir: I think this should be `roundToDecimals`. 😃 If not, we should choose a better name.
```suggestion
  roundToDecimals = 2,
```

src/utils/format.js
```diff
@@ -13,3 +15,41 @@ export const formatIntegerRange = (
...
+  isIncoming = true,
+  signDisplay = false
```
Evalir: Nitpick: I believe `displaySign` or `signed` makes more sense. Thoughts? @bpierre

src/utils/format.js
```diff
@@ -13,3 +15,41 @@ export const formatIntegerRange = (
...
+) {
+  const number = round(
+    (amount / Math.pow(10, decimals)),
+    rountToDecimals
+  )
+  const formattedNumber = formatNumber(number)
+  const sign = signDisplay ? (isIncoming ? '+' : '-') : ''
+  return `${sign}${formattedNumber} ${symbol}`
+}
```
Evalir: Let's add a newline after this.

src/utils/math.js
```diff
@@ -97,3 +97,19 @@ export function randomInt(min, max) {
...
+ *  - https://stackoverflow.com/a/19722641/1375656
+ *
+ * Fixed for NaNs on really small values
+ *
+ * @param {number} num Number to round
+ * @param {number} [places=2] Number of places to round to
+ * @returns {number} Rounded number
+ */
+export function round(num, places = 2) {
+  const rounded = Number(Math.round(num + 'e+' + places) + 'e-' + places)
```
Evalir: If we're doing string concatenation, let's try to use backticks for readability. 😃Also, we're using type coercion here; maybe there's a way of using JS properties to do this in a nicer way?

src/utils/format.js
bpierre: I agree `displaySign` sounds more natural, but I suggested this name [to be aligned with the new `Intl.NumberFormat` API](https://github.com/aragon/aragon-ui/issues/719#issuecomment-591516226). That being said, I don’t think it would be a big issue to ignore this and just pick what feels more natural to us.

I would personally avoid `signed` though: I think it could create a confusion with the concept of signed / unsigned integers.

Evalir: Ooh I see; in that case, it makes sense to be more aligned with APIs like Intl. Let's keep `signDisplay` then 👍

src/utils/math.js
ECWireless: How is something like this?

```
function round(num, places = 2) {
  return Math.round( ( Number(num) + Number.EPSILON ) * Math.pow(10, places) ) / Math.pow(10, places)
}
```

src/utils/format.js
```diff
@@ -13,3 +15,41 @@ export const formatIntegerRange = (
...
+ * @returns {string} Formatted number
+ */
+export function formatNumber(number) {
+  return new Intl.NumberFormat('en-US', {minimumFractionDigits: 0,}).format(number)
+}
+
+/**
+ * Builds on top on formatNumber(), and extends it with options for formatting tokens
+ *
+ * @param {number} amount Number to round
```
bpierre: I think it could be a good idea to also support integers defined as `String`, `BigInt`, or `BigInt`-like objects (JSBI, BN.js, BigNumber), because its value can be greater than `Number.MAX_SAFE_INTEGER`. The easiest would be to call `String()` on the passed value, and validate that the result is indeed an integer.

We also probably need a library to to round the value, perhaps we should add [JSBI](https://github.com/GoogleChromeLabs/jsbi) as a dependency to do it?

ECWireless: I've tried simply calling String() while passing a BigInt, and everything seems to work fine; even the rounding. Is there a reason why adding JSBI would definitely be necessary?

Everything seems to work without it, but maybe I'm missing the point of adding the library.

bpierre: Yes I think it would be needed to round properly in browsers that don’t support `BigInt`.

Let’s say we want to represent `2839.381295879707883945 ETH`, that would be `2839381295879707883945` in wei. On most systems, this number is too large to be represented using `Number`, and since `BigInt` can’t be used for compatibility reasons, a solution would be to use a string.

We would add the decimal point (`.`) by manipulating the string: `value.slice(0, decimals) + "." + value.slice(decimals, decimals + rountToDecimals)`. It would work well to round the number down, but it wouldn’t be able to round up:

```jsx
formatTokenAmount(
  '2839331295879707883945',
  'ETH',
  18,
  1
)

// Outputs 2,839.3 as expected

formatTokenAmount(
  '2839381295879707883945',
  'ETH',
  18,
  1
)

// Outputs 2,839.3 instead of rounding up to 2,839.4
```

That’s where `BigInt` or an implementation of it can help, so we can do the calculations using big integers directly.

I added some tests (that don’t pass): https://github.com/aragon/aragon-ui/pull/728/commits/794acbb9b4473ead052e1f8c80fc3bd2bbbfb550

Let me know what you think!

bpierre: Hey @ECWireless I was wondering if we could help in any way with this PR? Let me know :smiley:

ECWireless: I guess I'm wondering if it would be best to spit the value and run the integer through formatNumber(), while then running the fractional through a BigInt conversion? To be honest, I've never used BigInt implementations, so getting a good starting point would definitely help. I'll keep looking into too to see if I can solve it on my own.

bpierre: @ECWireless I just added a commit to support BigInt-like numbers. Not having `BigInt` involved having to add a custom `Intl.NumberFormat` and a function to divide and round at the same time, so we can round the decimals. Let me know what you think!

ECWireless: Super informative to read through the commit! Definitely think I understand the thought-process much more now. I tried running tests to understand everything a bit more, though, and it looks like the './characters' file wasn't committed? Maybe I've just missed it somehow though...

bpierre: You’re right, sorry about that! I added the missing file :+1:

ECWireless: Nice, I went through everything again, including the tests, and couldn't find anything to tweak. Was more educational than anything for me, especially the custom `Intl.NumberFormat`, which was tripping me up the most originally.

bpierre: Awesome, thanks for having taken the time to review it!

src/utils/format.js
```diff
@@ -1,15 +1,122 @@
...
+
+  if (displaySign) {
+    result = `${negative ? '-' : '+'}${result}`
+  }
+
+  if (symbol) {
+    result = `${result}${NO_BREAK_SPACE}${symbol}`
+  }
+
+  return result
```
sohkai, line 121: _Maybe_ this looks a bit simpler... but maybe not:

```js
return [
  displaySign ? negative ? '-' : '+' : '',
  result,
  rightPart && rightPart !== '0' ? `.${rightPart}` : '',
  symbol ? `${NO_BREAK_SPACE}${symbol}` : ''
].join('')
```

src/utils/format.test.js
```diff
@@ -0,0 +1,121 @@
...
+    expect(formatNumber(1234567.1234567)).toEqual('1,234,567.1234567')
+  })
+
+  test('should work with strings', () => {
+    expect(formatNumber('123456123456')).toEqual('123,456,123,456')
+    expect(formatNumber('1234567.1234567')).toEqual('1,234,567.1234567')
+  })
+
+  test('should work with BigInt', () => {
+    expect(formatNumber(BigInt('123456123456239873298739287'))).toEqual(
```
sohkai, line 45: I wonder if we should also specify our node engine in the package.json to be 12+, if we're expecting this to be available locally.

src/utils/format.test.js
```diff
@@ -0,0 +1,121 @@
...
+    )
+  })
+
+  test('should handle BigInt-like numbers', () => {
+    expect(formatTokenAmount(BigInt('2839381295879707883945'), 18, 8)).toEqual(
+      '2,839.38129588'
+    )
+  })
+
+  test('should handle non-decimal numbers', () => {
```
sohkai, line 74:
```suggestion
  test('should handle large integers', () => {
```

Or maybe "large numbers without decimals"?

src/utils/format.js
```diff
@@ -1,15 +1,122 @@
...
-  if (count > max) {
+  if (value > max) {
     return `${parseInt(max, 10)}${maxSuffix}`
   }
-  return count.toString()
+  return String(value)
+}
+
+/**
+ * Formats an integer for display purposes.
```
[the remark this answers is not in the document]
bpierre: Numbers sounds good yes!

src/utils/format.js
```diff
@@ -1,15 +1,122 @@
...
+ *
+ * @param {BigInt|string|number} dividend Integer to be divided + rounded
+ * @param {BigInt|string|number} diviser  Diviser
+ * @returns {string}
+ */
+function divideRoundBigInt(dividend, diviser) {
+  dividend = JSBI.BigInt(dividend)
+  diviser = JSBI.BigInt(diviser)
+  return JSBI.divide(
+    JSBI.add(dividend, JSBI.divide(diviser, JSBI.BigInt(2))),
```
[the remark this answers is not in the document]
bpierre: © internet (can’t remember where I found it :smile:)

src/utils/format.test.js
bpierre, line 45: @sohkai oops I deleted your comment about specifying Node 12+ as an engine version.

Yes good idea :+1:

src/utils/format.js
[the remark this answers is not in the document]
bpierre: https://github.com/aragon/aragon-ui/pull/728/commits/0e55ee854bdd53c2d24e0c4de1caa241f72ebcc9

src/utils/format.js
```diff
@@ -1,15 +1,122 @@
...
+ * integers expressed as string, or BigInt-like objects.
+ *
+ * @param {BigInt|string|number} number Number to convert
+ * @returns {string}
+ */
+export function formatNumber(number) {
+  const numAsString = String(number)
+  const [integer, decimals] = numAsString.split('.')
+  let result = ''
+  for (let i = 0, len = integer.length; i < len; i++) {
```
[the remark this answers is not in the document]
bpierre: Not sure it makes it more readable, but it’s ok for such a small piece of code :point_right: https://github.com/aragon/aragon-ui/pull/728/commits/9d40bc3405923670f450172075f04e9d82623064

src/utils/format.js
```diff
@@ -1,15 +1,122 @@
...
+}
+
+/**
+ * Formats a token amount for display purposes.
+ *
+ * @param {BigInt|string|number} amount              Number to round
+ * @param {BigInt|string|number} decimals            Decimal placement for amount
+ * @param {BigInt|string|number} roundToDecimals     Rounds the number to a given decimal place
+ * @param {string}               options.symbol      Symbol for the token amount
+ * @param {boolean}              options.displaySign Decides if the sign should be displayed
```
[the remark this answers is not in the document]
bpierre: Done https://github.com/aragon/aragon-ui/pull/728/commits/1544376f8a7a1dd6fc9042ca763ca74383fa2838

src/utils/format.js
```diff
@@ -1,15 +1,122 @@
...
+  amount = JSBI.BigInt(String(amount))
+  decimals = JSBI.BigInt(String(decimals))
+  roundToDecimals = JSBI.BigInt(String(roundToDecimals))
+
+  const _0 = JSBI.BigInt(0)
+  const _10 = JSBI.BigInt(10)
+  const negative = JSBI.lessThan(amount, _0)
+
+  if (negative) {
+    amount = JSBI.subtract(_0, amount)
```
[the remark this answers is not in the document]
bpierre: Yes! https://github.com/aragon/aragon-ui/pull/728/commits/9849045bcfa6b3bd7bcd22238fabf51b36742b8a

src/utils/format.js
bpierre: :+1:  https://github.com/aragon/aragon-ui/pull/728/commits/e4910f52ed4d254d47cd3ff93c5d6c583021ff25

src/utils/format.test.js
bpierre: https://github.com/aragon/aragon-ui/pull/728/commits/827a9ac45c958d44bbba072df39defc3155498c5

src/utils/format.test.js
bpierre: https://github.com/aragon/aragon-ui/pull/728/commits/878fa619d98824eb37c22db62d3fcc68bcf1ecc1

src/utils/format.js
```diff
@@ -40,14 +40,17 @@ export function formatIntegerRange(
...
-      result = ',' + result
-    }
-    result = integer[len - i - 1] + result
-  }
-  return result + (decimals ? `.${decimals}` : '')
+
+  return [...integer].reduceRight(
+    (result, digit, index, { length }) => {
+      const position = length - index - 1
+      if (position > 0 && position % 3 === 0) {
```
sohkai, line 47: Ah jeez, I thought `reduceRight()` would change the index so that it started "backwards" from `0`. That was one of the main simplifications I had wanted with it :).

What do you think instead of doing this with `[...integer].reverse().reduce()`? That way the position is the index :).

src/utils/format.js
```diff
@@ -1,15 +1,119 @@
...
+ * @param {BigInt|string|number} roundToDecimals     Rounds the number to a given decimal place
+ * @param {boolean}              options.displaySign Decides if the sign should be displayed
+ * @param {string}               options.symbol      Symbol for the token amount
+ * @returns {string}
+ */
+export function formatTokenAmount(
+  amount,
+  decimals = 18,
+  roundToDecimals = 2,
+  { symbol = '', displaySign = false } = {}
```
sohkai, line 86: Ah, I missed this in the first pass around... but I wonder if it would be worth making `decimals` and `roundToDecimals` part of the options object?

src/utils/format.test.js
```diff
@@ -0,0 +1,121 @@
...
+  })
+
+  test('should handle BigInt-like numbers', () => {
+    expect(formatTokenAmount(BigInt('2839381295879707883945'), 18, 8)).toEqual(
+      '2,839.38129588'
+    )
+  })
+
+  test('should handle large numbers without decimals', () => {
+    expect(formatTokenAmount(BigInt('2839000000000000000000'), 18, 8)).toEqual(
```
sohkai, line 75: Might be worthwhile adding a test like `formatTokenAmount(BigInt('2839000000010000000000'), 18, 8)` or something, just to see the decimal truncation after the last significant digit.

src/utils/format.js
```diff
@@ -1,15 +1,119 @@
...
+  const _10 = JSBI.BigInt(10)
+  const negative = JSBI.lessThan(amount, _0)
+
+  if (negative) {
+    amount = JSBI.unaryMinus(amount)
+  }
+
+  const amountConverted = divideRoundBigInt(
+    amount,
+    JSBI.exponentiate(_10, JSBI.subtract(decimals, roundToDecimals))
```
sohkai, line 102: Just wondering what happens if this goes negative. Should we guard against this and cut it to 0?

src/utils/format.js
```diff
@@ -1,15 +1,119 @@
...
+}
+
+/**
+ * Divide and round two big integers.
+ *
+ * @param {BigInt|string|number} dividend Integer to be divided + rounded
+ * @param {BigInt|string|number} divisor  Divisor
+ * @returns {string}
+ */
+function divideRoundBigInt(dividend, diviser) {
```
sohkai, line 63: Ahh, we forgot to update the parameter name here :)

src/utils/format.js
```diff
@@ -1,15 +1,119 @@
...
+/**
+ * Formats a number for display purposes.
+ *
+ * This function is not using Intl.NumberFormat() to be compatible with big
+ * integers expressed as string, or BigInt-like objects.
+ *
+ * @param {BigInt|string|number} number Number to convert
+ * @returns {string}
+ */
+export function formatNumber(number) {
```
sohkai, line 40: I wonder if we should also include a mechanism in this utility to do rounding; perhaps it'd just involve moving the rounding bit in `formatTokenAmount()` to here for the `decimals`.

Having it exposed here it also make it easier for us to format strings to be _lower_ precision than before, e.g.:

- Start with a token amount (BN to string, via `formatTokenAmount()`), adjusted to 18 decimals of full precision
- When we want to display a shorter rounded amount, use this string with `formatNumber()`

Or do you think we should just recommend people to always keep the non-decimal adjusted amounts and use `formatTokenAmount()`?

src/utils/format.js
bpierre: Done in https://github.com/aragon/aragon-ui/pull/760/commits/e5399977ff62dd2728812ea23b45d8ff7327f582

src/utils/format.js
bpierre: Done in https://github.com/aragon/aragon-ui/pull/760/commits/7ed7f5109073ba0cbca45c83ba279cca569a9325

- Removed the default `18` from `decimals` to make it a required parameter.
- `roundToDecimals` has moved into `options.digits`.

src/utils/format.js
bpierre: I think we could throw in that case, what do you think? https://github.com/aragon/aragon-ui/pull/760/commits/12b1e72804b30d1ecde51f0940e1bec53b167277

src/utils/format.test.js
bpierre: Done and good find! It made visible an issue with the way leading & trailing zeros were handled: https://github.com/aragon/aragon-ui/pull/760/commits/6435c36c47a6df33b7e6ac1abd0c0a976f00e3a2

src/utils/format.js
bpierre: This shouldn’t be needed with [`TokenAmount`](https://github.com/aragon/aragon-ui/pull/757), what do you think?

sohkai: Yep, I think `TokenAmount` solves this nicely!

PR #729, opened 2020-03-06
New Component: Details

### LINKED ISSUE #675: Add Toggle component
It would be nice if we had a generic Toggle for showing / hiding components:

Screen Shot 2019-11-22 at 2 27 10 AM

See https://github.com/aragon/aragon/blob/master/src/components/MenuPanel/MenuPanel.js#L142

Jfelix61:
Closes #675 .

## Changes

- Added `ToggleComponents` component and README

## Demo
ezgif com-video-to-gif


src/components/ToggleComponents/README.md
```diff
@@ -0,0 +1,52 @@
+# TransactionBadge
```
Evalir: Let's properly update the description for this component. :)

src/components/index.js
```diff
@@ -71,6 +71,7 @@ export { default as TextInput } from './Input/TextInput'
 export { default as Timer } from './Timer/Timer'
 export { default as TokenBadge } from './TokenBadge/TokenBadge'
 export { default as TransactionBadge } from './TransactionBadge/TransactionBadge'
+export { default as ToggleComponent } from './ToggleComponents/ToggleComponents'
```
Evalir: Let's alphabetize this import by placing it above the `TokenBadge` component. :)

src/components/ToggleComponents/ToggleComponents.js
```diff
@@ -0,0 +1,200 @@
...
+
+const openedState = {
+  key: 'APPS_OPENED_STATE',
+  isOpen: function() {
+    return localStorage.getItem(this.key) === '1'
+  },
+  set: function(opened) {
+    localStorage.setItem(this.key, opened ? '1' : '0')
+  },
+}
```
Evalir: Should we keep this pattern for the internal component state when porting this to aragonUI? If so, I think we should pass `key` as a prop. 😃What do you think @bpierre ?

src/components/ToggleComponents/ToggleComponents.js
```diff
@@ -0,0 +1,200 @@
...
+
+function ToggleComponents({ children, itemBaseHeight, label, ...props }) {
```
Evalir: Let's rename this to just `Toggle`, and the same with the file name.

src/components/ToggleComponents/ToggleComponents.js
```diff
@@ -0,0 +1,200 @@
...
+  itemBaseHeight: PropTypes.number,
+  label: PropTypes.string.isRequired,
+}
+
+ToggleComponents.defaultProps = {
+  itemBaseHeight: 5 * GU,
+  label: '',
+}
+
+function AppsToggle(props) {
```
Evalir: Let's remove the `Apps` prefix and use something more generic, such as `ItemsToggle`, as we're generalizing the use of this component outside the `MenuPanel`.

src/components/ToggleComponents/README.md
```diff
@@ -0,0 +1,52 @@
...
+
+A component to hide/show components in a list
+
+## Usage
+
+```jsx
+import { ToggleComponent } from '@aragon/ui'
```
Evalir: Let's properly update this when we change the file name from the component itself; this is not matching the current name.

src/components/ToggleComponents/ToggleComponents.js
```diff
@@ -0,0 +1,200 @@
...
+    </h1>
+  )
+}
+
+Heading.propTypes = {
+  children: PropTypes.node,
+  label: PropTypes.node,
+}
+
+const AppsToggleArrow = props => (
```
Evalir: Ditto; let's make this name more generic.

src/components/ToggleComponents/ToggleComponents.js
```diff
@@ -0,0 +1,200 @@
...
+        align-items: center;
+        justify-content: center;
+      `}
+    >
+      <IconDown size="tiny" />
+    </div>
+  </animated.div>
+)
+
+const AppsToggleShadow = props => (
```
Evalir: Same here; generic naming.

devbox/apps/ToggleComponents.js
```diff
@@ -0,0 +1,66 @@
...
+const testApps = [
+  { id: '1', name: 'foo', icon: IconAlert },
+  { id: '2', name: 'bar', icon: IconAragon },
+  { id: '3', name: 'zeu', icon: IconAttention },
+]
+
+function App() {
+  return (
+    <div>
+      <ToggleComponent label="Random Items" itemBaseHeight={4 * GU}>
```
Evalir: Let's improve the example by resembling the usage in the [`MenuPanel`](https://github.com/aragon/aragon/blob/master/src/components/MenuPanel/MenuPanel.js#L231) regarding the copies and the buttons. :) Also, one example is just fine!

src/components/Details/README.md
```diff
@@ -0,0 +1,51 @@
...
+
+The content to reveal. Note that it this node is always mounted even when hidden, but this might change in the future.
+
+### `opened`
+
+| Type      | Default value |
+| --------- | ------------- |
+| `Boolean` | None          |
+
+When set, allows to control the state of `Details` from the outside ([controlled component](https://reactjs.org/docs/forms.html#controlled-components)). Leave unset to let `Details` handle its own opened / close state.
```
sohkai, line 43:
```suggestion
When set, allows to control the state of `Details` from the outside (as a [controlled component](https://reactjs.org/docs/forms.html#controlled-components)). Leave unset to let `Details` manage its own opened / closed state.
```

src/components/Details/Details.js
```diff
@@ -0,0 +1,208 @@
...
+
+  // Animate after the initial render
+  const animate = useRef(false)
+  useEffect(() => {
+    animate.current = true
+  }, [])
+
+  // Use height: 'auto' when opened
+  const [forceHeight, setForceHeight] = useState(false)
+  const handleRest = useCallback(() => setForceHeight(!opened), [opened])
```
sohkai, line 64:
```suggestion
  const handleRest = useCallback(() => setForceHeight(opened => !opened), [])
```

Would this also work?

src/components/Details/Details.js
```diff
@@ -0,0 +1,208 @@
...
+                flex-direction: column;
+                justify-content: flex-end;
+              `}
+              style={{
+                opacity: openProgress,
+                height: forceHeight
+                  ? openProgress.interpolate(
+                      v => `${v * contentHeight.current}px`
+                    )
+                  : 'auto',
```
sohkai, line 180: Interesting, is this to force it to fill the space when doing the transition?

src/components/Details/Details.js
```diff
@@ -0,0 +1,208 @@
...
+      to={{ openProgress: Number(opened) }}
+      immediate={!animate}
+      onRest={handleRest}
+      onStart={handleStart}
+      native
+    >
+      {({ openProgress }) => (
+        <section
+          css={`
+            margin: ${insideSidePanel ? `${2 * GU}px ${-3 * GU}px 0` : '0'};
```
sohkai, line 83: The negative margins weren't in the version on aragon/aragon I think—do we need them?

devbox/apps/Details.js
```diff
@@ -0,0 +1,32 @@
+import React, { useState } from 'react'
+import { Details, GU, Info, SidePanel } from '@aragon/ui'
+
+function App() {
+  return (
+    <SidePanel title="Details demo">
```
sohkai, line 6: It'd be nice to see a non-sidepanel version as well!

src/components/Details/Details.js
bpierre: It would yes :+1:

src/components/Details/Details.js
bpierre: Yes, this is to cancel the `SidePanel` padding.

src/components/Details/Details.js
bpierre: The `auto` is here so that we can update the height of the content when opened, or right before opening, without breaking it visually. This is needed on the aragon/aragon menu, so we can dynamically add “Console” :-)

devbox/apps/Details.js
bpierre: This component is a bit special because it is the first one we add to aragonUI without having a use case for apps − that’s why I wanted to make it work in `SidePanel`, so that there is at least one way to use it, outside of custom cases.

Now I think we could try to make it work in `<Box>`, kind of similar to what we are doing on the error screen… let me see if that works.

src/components/Details/Details.js
sohkai: Ohh gotcha, I didn't notice they were still there for the SidePanel! Thought it was for the else case!

src/components/Details/Details.js
bpierre: …no it doesn’t haha, we need to use the `opened` from the previous render cycle. I reverted => https://github.com/aragon/aragon-ui/pull/729/commits/97906a11086214c358b5ef3e1d00288056282a84

devbox/apps/Details.js
bpierre: Done in https://github.com/aragon/aragon-ui/pull/729/commits/3a8968a102886f6b83134e2807f7352c49fe8db6 and https://github.com/aragon/aragon-ui/pull/729/commits/7ada1f9733b0a221d6250343a23a4254c1d510de

src/components/Details/Details.js
sohkai: Sounds good, might be useful to leave a comment!

PR #732, opened 2020-03-08
Add @testing-library/react

Jfelix61:
## Changes 

- Added react-testing-library
- Added initial tests for `Badge` and `Button`

## Current problem

`use-inside` package makes `Button` test fail. It seems to be an issue with how it is bundled with commonjs (perhaps its entry point?). Tried compiling it with Babel however it doesn't seem to solve the issue. Since the library is not available in Github I will have to leave for the maintainers at Aragon to check on it. 

Aside from this problem, `Badge` test is working fine and tests can be written for it.


sohkai:
Thanks for taking a shot at this @Jfelix61!

> Since the library is not available in Github

Actually, it is! It's here: [`use-inside`](https://github.com/aragon/use-inside).

Jfelix61:
> Actually, it is! It's here: [`use-inside`](https://github.com/aragon/use-inside).

Great I'll be checking it and see what could be going wrong.

bpierre:
Hi @Jfelix61, thanks for working on this! I added a few commits to sync the branch with `master` and to use CommonJS modules. It should work now, let me know if I can help with anything!

Jfelix61:
Thanks for this @bpierre! I'll keep adding tests.

bpierre:
Hey @Jfelix61! I was wondering if we should merge the setup already, and add the tests in another PR. What do you think?

Jfelix61:
> Hey @Jfelix61! I was wondering if we should merge the setup already, and add the tests in another PR. What do you think?

I think the same, that way we can keep adding tests in different PRs.

bpierre:
Awesome, let’s merge it then! :tada:

Jfelix61:
Great!

PR #733, opened 2020-03-10
TransactionProgress: fix prop type error

### LINKED ISSUE #721: Website: React error when trying to see the "TransactionProgress" example
When the `Show Component` button for the demo is clicked, React throws an error that causes the site to go completely blank, needing a full refresh. Seems like we're missing a prop or something, as when decoding the error, I got this:

> Element type is invalid: expected a string (for built-in components) or a class/function (for composite components) but got: undefined.

nadalfederer:
~closes #721~

Also move `Info` to a default export.


src/components/Info/index.js
```diff
@@ -5,9 +5,4 @@ function Warning(props) {
   return <Info mode="warning" {...props} />
 }
 
-// Backward compatibility
-Info.Action = Info
-Info.Permissions = Warning
```
nadalfederer: There are no usage for Action and Permission in the project. So, I think it's better to delete them to avoid errors

sohkai: These were for backwards compatibility with previous versions of the library.

Please revert this change; we will deprecate and remove these in the future, on a separate version.

bpierre:
Hey @nadalfederer, thank you for this PR! While working on it, I realized the problem wasn’t here anymore, but there was still a a prop type error that I fixed, so I updated the title of the PR to reflect that. Merging it!

PR #734, opened 2020-03-10
Timer show icon prop

fabriziovigevani:
Adding this new prop so we can choose to not show the clock icon.

This will be useful for the court dashboard. See https://github.com/aragon/court-dashboard/pull/227#discussion_r389214082

### DIFF
--- a/src/components/Timer/README.md
+++ b/src/components/Timer/README.md
@@ -26,33 +26,42 @@ const App = () => (
 
 ### `end`
 
-- Type: `Date`
+| Type      | Default value |
+| --------- | ------------- |
+| `Date`    | None          |
 
 The end of the countdown, as a `Date` instance.
 
 ### `start`
 
-- Type: `Date`
+| Type      | Default value |
+| --------- | ------------- |
+| `Date`    | None          |
 
 The start of a timer, as a `Date` instance.
 
 ### `format`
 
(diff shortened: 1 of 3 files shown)


src/components/index.js
```diff
@@ -70,8 +70,12 @@ export { default as TextCopy } from './TextCopy/TextCopy'
 export { default as TextInput } from './Input/TextInput'
 export { default as Timer } from './Timer/Timer'
 export { default as TokenBadge } from './TokenBadge/TokenBadge'
-export { default as TransactionBadge } from './TransactionBadge/TransactionBadge'
-export { default as TransactionProgress } from './TransactionProgress/TransactionProgress'
+export {
```
sohkai: I feel like we must not be using the same prettier config; these files seem to be going back and forth in formatting :).

src/components/Timer/README.md
```diff
@@ -49,3 +49,10 @@ Format output in years 'y', months 'M', days 'd', hours 'h', minutes 'm', second
 - Default: `false`
 
 Display the units on the left side of the timer when they are equal to zero.
+
+### `showIcon`
```
sohkai: Since we're touching this file, we might as well update its format to be in the same table format as most other component's README.mds :).

src/components/index.js
fabriziovigevani: Yes, surprised me that when pushing the changes the `lint` gave me those errors

PR #744, opened 2020-04-02
Hooks: useFocusEnter and useFocusLeave

### LINKED ISSUE #475: Add useFocusEnter() / useFocusLeave()
See https://github.com/aragon/aragon-ui/pull/391#discussion_r313862603

rperez89:
This pr closes #475 

see [discussion](https://github.com/aragon/aragon-ui/pull/391#discussion_r313862603)


src/hooks/useFocusLeave.js
```diff
@@ -1,9 +1,9 @@
 import { useCallback, useRef } from 'react'
 
 /* eslint-disable react-hooks/rules-of-hooks */
-export function useOnBlur(cb, ref = useRef()) {
+export function useFocusLeave(cb, ref = useRef()) {
```
bpierre, line 4: We should keep `useOnBlur()` for a while to warn users:

```jsx
function useOnBlur(...params) {
  warnOnce('useOnBlur', 'useOnBlur() is deprecated, please use useFocusLeave() instead.')
  return useFocusLeave(...params)
}
```

src/hooks/useFocusEnter.js
```diff
@@ -0,0 +1,16 @@
+import { useCallback, useRef } from 'react'
+
+/* eslint-disable react-hooks/rules-of-hooks */
+export function useFocusEnter(cb, ref = useRef()) {
+  /* eslint-enable react-hooks/rules-of-hooks */
```
bpierre, line 5: Nitpicking: should we declare `useRef()` so we can remove the ESLint comments?

src/hooks/useFocusEnter.js
```diff
@@ -0,0 +1,16 @@
...
+  const handleFocusEnter = useCallback(
+    event => {
+      if (ref.current.contains(event.target)) {
+        cb(event)
+      }
+    },
+    [cb, ref]
```
bpierre, line 12: `ref` shouldn’t be needed, it never changes:

```suggestion
    [cb]
```

src/hooks/useFocusEnter.js
```diff
@@ -0,0 +1,16 @@
+import { useCallback, useRef } from 'react'
+
+/* eslint-disable react-hooks/rules-of-hooks */
+export function useFocusEnter(cb, ref = useRef()) {
+  /* eslint-enable react-hooks/rules-of-hooks */
+  const handleFocusEnter = useCallback(
+    event => {
+      if (ref.current.contains(event.target)) {
```
bpierre, line 8: We should check for `ref.current` and `ref.current.contains`:
```suggestion
      if (ref.current?.contains?.(event.target)) {
```

src/hooks/useFocusEnter.js
rperez89: could you expand this a little? you mean declaring inside the function?

bpierre: Oh I was assuming that the `eslint-disable` was needed because `useRef()` was declared here instead of the function body, but I realize now that it is because of its conditional nature.

To avoid calling `useRef()` conditionally but still accept an optional ref, we could do this:

https://github.com/aragon/aragon-ui/blob/master/src/components/SidePanel/SidePanel.js#L284-L288

PR #745, opened 2020-04-02
Slider: Add onTouchStart to slider circle

### LINKED ISSUE #718: Slider: Handle gets cut off at ends
For example, see:

Screen Shot 2020-02-14 at 9 07 32 PM

I would expect the handle to always be interactive, but at the end, if there are margins or paddings, it becomes unresponsive up to the "half way" point at the slider's edge.

rperez89:
This PR closes #718 

slider

### DIFF
--- a/src/components/Slider/Slider.js
+++ b/src/components/Slider/Slider.js
@@ -165,8 +165,6 @@ class Slider extends React.Component {
                 >
                   <animated.div
                     style={this.getHandleStyles(pressProgress)}
-                    onMouseDown={this.dragStart}
-                    onTouchStart={this.dragStart}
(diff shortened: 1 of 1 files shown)


src/components/Slider/Slider.js
```diff
@@ -205,7 +208,6 @@ const Bar = styled(animated.div)`
 `
 
 const HandleClip = styled.div`
-  pointer-events: none;
```
bpierre, line 208: I think this change only should be enough? But a problem it introduces is that it makes the entire `HandleClip` area clickable (which is bigger than the slider + its padding).

Issue #193, opened 2018-07-11
Expose math utils


sohkai:
Could also expose some of the math utils used in the apps.

PR #750, opened 2020-04-08
Export all the utils

bpierre:
New exports:

- ~`cachedMap()`~
- `prefixUrl()`
- `ensureTrailingSlash()`
- `map()`, `norm()`, `lerp()`, `clamp()`
- `random()`, `randomInt()`

Closes https://github.com/aragon/aragon-ui/issues/193


sohkai:
Just not sure about `cachedMap()`—do you think that'll be really useful to others?

bpierre:
> Just not sure about cachedMap()—do you think that'll be really useful to others?

I’m not sure it would be useful, but the intention is to export all of `utils/` without exception. Maybe we could create a `lib/` directory for the things we don’t want to expose, and put `cachedMap()` in there?

sohkai:
> Maybe we could create a lib/ directory for the things we don’t want to expose, and put cachedMap() in there?

Yep, that sounds good!

bpierre:
@sohkai Done!

Issue #476, opened 2019-08-14
Theme: warn when a color is deprecated or inexistant


bpierre:
See https://github.com/aragon/aragon-ui/pull/391#discussion_r313839246

PR #751, opened 2020-04-08
Theme: warn when invalid colors are being accessed

bpierre:
Fixes https://github.com/aragon/aragon-ui/issues/476

Warn in the following cases:

- Deprecated colors.
- Unknown colors.
- Invalid theme.

Also:

- Move TabBarLegacy to `useTheme()`.
- Move deprecated colors in `Theme.js`, rather than in every theme.

### DIFF
--- a/src/components/Tabs/TabBarLegacy.js
+++ b/src/components/Tabs/TabBarLegacy.js
@@ -1,7 +1,6 @@
 import React from 'react'
 import PropTypes from 'prop-types'
 import styled from 'styled-components'
-import { theme } from '../../theme-legacy'
 import { font, unselectable, noop } from '../../utils'
 
 class TabBar extends React.Component {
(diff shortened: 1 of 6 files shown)


src/theme/Theme.js
```diff
@@ -32,9 +40,7 @@ function getTheme(theme) {
   return { ...baseTheme, ...theme }
 }
 
-const ThemeContext = React.createContext(
-  convertThemeColors(getTheme(THEME_DEFAULT))
-)
+const ThemeContext = React.createContext(prepareTheme(getTheme(THEME_DEFAULT)))
```
sohkai, line 43: Do we need to set the default if we're already providing it in `Theme`?

src/components/Tabs/TabBarLegacy.js
```diff
@@ -92,43 +91,83 @@ class TabBar extends React.Component {
   }
 }
 
-const Bar = styled.ul`
-  display: flex;
-  border-bottom: ${p => (p.border ? `1px solid ${theme.contentBorder}` : '0')};
-`
+/* eslint-disable react/prop-types */
```
sohkai, line 95: Looks like we could remove this extra line

src/theme/Theme.js
```diff
@@ -12,13 +12,21 @@ const EMBEDDED_THEMES = { dark, light }
 const THEME_DEFAULT = 'light'
 
 const RESERVED_KEYS = ['_appearance', '_name']
+const DEPRECATED_COLORS = new Map([
+  ['error', 'negative'],
+  ['success', 'positive'],
+])
 
 const COLOR_FALLBACK = '#FF00FF'
```
sohkai, line 20: Just wondering if we should use one of the light theme's values for this instead of hardcoding it?

src/components/Tabs/TabBarLegacy.js
```diff
@@ -1,7 +1,6 @@
 import React from 'react'
```
sohkai, line 1: 🙌Is this the last internal component still using the old theme?

bpierre: It is yes!

src/theme/Theme.js
bpierre: The idea is really to make it visibly wrong, that’s the color being used: `#FF00FF`

src/components/Tabs/TabBarLegacy.js
bpierre: It disables `react/prop-types` for all the internal components below and not only the first one, so I think we could leave the empty line?

src/theme/Theme.js
bpierre: We don’t, let’s remove it!

bpierre: https://github.com/aragon/aragon-ui/pull/751/commits/fdb00eaea638035a1b55ea1c8128d94c885d6a62

src/components/Tabs/TabBarLegacy.js
sohkai: Ah gotcha, missed that it was for the rest of the file!

PR #752, opened 2020-04-09
useKeyDown: propagate event's keycode to handler

sohkai:
In cases where there are multiple keycodes listened to, this allows the handler to know which key was pressed and how to handle it.

### DIFF
--- a/src/hooks/useKeyDown.js
+++ b/src/hooks/useKeyDown.js
@@ -6,7 +6,7 @@ export function useKeyDown(key, callback) {
   const handleKeyDown = useCallback(
     event => {
       if (keys.includes(event.keyCode)) {
-        callback(event.keyCode)
+        callback(event.keyCode, event)
       }
     },
     [callback, keys]
--- a/src/hooks/useKeyDown.js
+++ b/src/hooks/useKeyDown.js
@@ -3,7 +3,7 @@ import { useCallback, useEffect, useMemo } from 'react'
 export function useKeyDown(key, callback) {
   const keys = useMemo(() => (Array.isArray(key) ? key : [key]), [key])
 
-  const handlekeyDown = useCallback(
+  const handleKeyDown = useCallback(
     event => {
       if (keys.includes(event.keyCode)) {
         callback(event.keyCode)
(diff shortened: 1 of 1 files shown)


src/hooks/useKeyDown.js
```diff
@@ -3,10 +3,10 @@ import { useCallback, useEffect, useMemo } from 'react'
 export function useKeyDown(key, callback) {
   const keys = useMemo(() => (Array.isArray(key) ? key : [key]), [key])
 
-  const handlekeyDown = useCallback(
+  const handleKeyDown = useCallback(
     event => {
       if (keys.includes(event.keyCode)) {
-        callback()
+        callback(event.keyCode)
```
bpierre, line 9: I’m wondering if we should pass the [KeyboardEvent](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent) as a second parameter, or even directly?

sohkai: Yes, was thinking this too... do you think it'd be better to just be more transparent?

sohkai: Also wonder: do you think we should do a `e.preventDefault()` here when we capture it?

bpierre: Do you mean more transparent by only passing the event?

I’m thinking that from the other side, the difference would be having to do `({ keyCode }) => {` rather than `keyCode => {`, which is acceptable, but it wouldn’t allow to pass a `useState` setter directly:

```jsx
function MyComponent() {
  const [key, setKey] = useState(-1)

  // if the key is passed first
  useKeyDown([KEY_DOWN, KEY_UP], setKey)

  // if the event is passed first
  useKeyDown([KEY_DOWN, KEY_UP], ({ keyCode }) => {
    setKey(keyCode)
  })
}
```

It’s also the reason why components like Checkbox pass their value directly in the `onChange` callback, but in that case it’s a bit simpler: the main information is whether the checkbox is checked or not, while the main information of a keyboard event is a combination of different entries like `key`, `code`, `shiftkey` etc.

In the case of `useKeyDown`, we are filtering by `keyCode` already, so I think it makes sense to pass it as a first parameter for this reason, and perhaps the event could be passed as a second parameter?

sohkai: > Do you mean more transparent by only passing the event?

Yes, this is what I meant.

> Main information of a keyboard event is a combination of different entries like key, code, shiftkey etc.

Good point, I hadn't considered these other keys!

-------

> In the case of useKeyDown, we are filtering by keyCode already, so I think it makes sense to pass it as a first parameter for this reason, and perhaps the event could be passed as a second parameter?

Yes, let's do this then!

sohkai: Updated in f7bd10d.

PR #755, opened 2020-04-11
Dropdown: elements min width

rperez89:
Not sure if this is the desired behavior but opening to start the discussion

Issue : https://github.com/aragon/aragon-apps/issues/1047


src/components/DropDown/DropDown.js
```diff
@@ -186,7 +178,7 @@ const DropDown = React.memo(function DropDown({
           padding-left: ${2 * GU}px;
           padding-right: ${1.5 * GU}px;
           width: ${width || (wide ? '100%' : 'auto')};
-          min-width: ${wide ? 'auto' : `${placeholderMinWidth}px`};
+          min-width: auto;
```
sohkai, line 181: I think we need this to be set directly, because using `auto` won't automatically adjust for the width of all items.

The original behaviour was intended for the dropdown's width to fit all listed items, even if they were not selected.

src/components/DropDown/DropDown.js
```diff
@@ -9,8 +9,6 @@ import { useViewport } from '../../providers/Viewport/Viewport'
 import ButtonBase from '../ButtonBase/ButtonBase'
 import Popover from '../Popover/Popover'
 
-const MIN_WIDTH = 128
```
Evalir, line 12: Should we try and use `GU` here?

src/components/DropDown/DropDown.js
```diff
@@ -111,26 +109,20 @@ const DropDown = React.memo(function DropDown({
...
       return
     }
-    setPlaceholderMinWidth(el.clientWidth)
     setMeasureWidth(false)
   }, [])
   useEffect(() => {
     setMeasureWidth(true)
   }, [vw, items])
 
+  console.log('measured width ', measureWidth)
```
Evalir, line 125: Let's remember to remove this when this is ready to review. 😃

PR #757, opened 2020-04-12
Add TokenAmount

bpierre:
This class will help to consolidate a token amount into its own type. For now it is mostly about using its `format()` method, but we can imagine adding other things in the future.

We might also want to make some aragonUI components aware of it, e.g. a `<TransferAmount />` component could be used to represent the transferred amounts in the Finance app, and accept `TokenAmount` instances.

About this, any thoughts on exporting a class to use with `new TokenAmount()` vs. a `tokenAmount()` function that would do the same thing? Since this is primarily a components library, I was thinking that we might want to avoid exporting names that might conflict with component names in the future.

### DIFF
--- a/src/utils/TokenAmount.js
+++ b/src/utils/TokenAmount.js
@@ -11,9 +11,9 @@ class TokenAmount {
 
   /**
    * Create a TokenAmount.
-   * @param {BigInt|string|number} amount   The amount as an integer (e.g. in Wei for Ethers).
-   * @param {BigInt|string|number} decimals The token decimals (e.g. 18 for Ethers).
-   * @param {string} options.symbol         The token symbol (e.g. ETH for Ethers).
+   * @param {BigInt|string|number} amount   The amount as an integer (e.g. in Wei for Ether).
+   * @param {BigInt|string|number} decimals The token decimals (e.g. 18 for Ether).
+   * @param {string} options.symbol         The token symbol (e.g. ETH for Ether).
    */
   constructor(amount, decimals, { symbol = '' } = {}) {
     amount = toJsbi(amount)
@@ -29,7 +29,7 @@ class TokenAmount {
   }
 
   /**
-   * Get the amount as an integer (e.g. in Wei for Ethers).
+   * Get the amount as an integer (e.g. in Wei for Ether).
    * @returns {BigInt}
    */
   amount() {
@@ -37,7 +37,7 @@ class TokenAmount {
   }
 
   /**
-   * Get the amount as an integer (e.g. in Wei for Ethers).
+   * Get the amount as an integer (e.g. in Wei for Ether).
    * @returns {string}
    */
   amountString() {
--- a/src/utils/TokenAmount.js
+++ b/src/utils/TokenAmount.js
@@ -52,7 +52,8 @@ class TokenAmount {
    * @returns {string}
    */
   format({ sign = false, symbol = false, digits = 2 } = {}) {
-    return formatTokenAmount(this.#amount, this.#decimals, digits, {
+    return formatTokenAmount(this.#amount, this.#decimals, {
+      digits,
       sign,
       symbol: symbol ? this.#symbol : '',
     })
--- a/src/utils/TokenAmount.js
+++ b/src/utils/TokenAmount.js
@@ -1,6 +1,7 @@
 /* global BigInt */
 
 import JSBI from 'jsbi'
+import { toJsbi } from './math'
 import { formatTokenAmount } from './format'
 
 class TokenAmount {
@@ -15,8 +16,15 @@ class TokenAmount {
    * @param {string} options.symbol         The token symbol (e.g. ETH for Ethers).
    */
   constructor(amount, decimals, { symbol = '' } = {}) {
-    this.#amount = JSBI.BigInt(amount)
-    this.#decimals = JSBI.BigInt(decimals)
+    amount = toJsbi(amount)
+    decimals = toJsbi(decimals)
+
+    if (JSBI.lessThan(decimals, 0)) {
+      throw new Error('TokenAmount: decimals cannot be negative')
+    }
+
+    this.#amount = amount
+    this.#decimals = decimals
     this.#symbol = symbol
   }
 
--- a/src/utils/TokenAmount.js
+++ b/src/utils/TokenAmount.js
@@ -1,5 +1,3 @@
-/* global BigInt */
-
 import JSBI from 'jsbi'
 import { toJsbi } from './math'
 import { formatTokenAmount } from './format'
@@ -29,18 +27,11 @@ class TokenAmount {
   }
 
   /**
-   * Get the amount as an integer (e.g. in Wei for Ether).
-   * @returns {BigInt}
-   */
-  amount() {
-    return BigInt(this.#amount.toString())
-  }
-
-  /**
-   * Get the amount as an integer (e.g. in Wei for Ether).
+   * Get the amount of the token without the decimals (e.g. in Wei for Ether),
+   * as a string integer.
    * @returns {string}
    */
-  amountString() {
+  amount() {
     return this.#amount.toString()
   }
 
--- a/src/utils/TokenAmount.js
+++ b/src/utils/TokenAmount.js
@@ -1,0 +1,92 @@
+/* global BigInt */
+
+import JSBI from 'jsbi'
+import { formatTokenAmount } from './format'
+
+class TokenAmount {
+  #amount
+  #decimals
+  #symbol
+
+  /**
+   * Create a TokenAmount.
+   * @param {BigInt|string|number} amount   The amount as an integer (e.g. in Wei for Ethers).
+   * @param {BigInt|string|number} decimals The token decimals (e.g. 18 for Ethers).
+   * @param {string} options.symbol         The token symbol (e.g. ETH for Ethers).
+   */
+  constructor(amount, decimals, { symbol = '' } = {}) {
+    this.#amount = JSBI.BigInt(amount)
+    this.#decimals = JSBI.BigInt(decimals)
+    this.#symbol = symbol
+  }
+
+  /**
+   * Get the amount as an integer (e.g. in Wei for Ethers).
+   * @returns {BigInt}
+   */
+  amount() {
+    return BigInt(this.#amount.toString())
+  }
+
+  /**
+   * Get the amount as an integer (e.g. in Wei for Ethers).
+   * @returns {string}
+   */
+  amountString() {
+    return this.#amount.toString()
+  }
+
+  /**
+   * Get the decimals of the token.
+   * @returns {number}
+   */
+  decimals() {
+    return this.#decimals.toNumber()
+  }
+
+  /**
+   * Formats the token amount for display purposes.
+   * @param {string} options.displaySign     Whether to display the sign or not.
+   * @param {string} options.displaySymbol   Whether to display the token symbol or not.
+   * @param {string} options.digits          The number of digits to appear after the decimal point.
+   * @returns {string}
+   */
+  format({ sign = false, symbol = false, digits = 2 } = {}) {
+    return formatTokenAmount(this.#amount, this.#decimals, digits, {
+      sign,
+      symbol: symbol ? this.#symbol : '',
+    })
+  }
+
+  /**
+   * Returns an object to be serialized by JSON.stringify().
+   * @returns {object}
+   */
+  toJSON() {
+    return {
+      amount: this.#amount.toString(),
+      decimals: this.#decimals.toString(),
+      symbol: this.#symbol,
+    }
+  }
+
+  /**
+   * Instanciate a new TokenAmount from the data serialized by JSON.stringify().
+   * @returns {object}
+   */
+  static fromJSON(jsonData) {
+    try {
+      const { amount, decimals, symbol } = JSON.parse(jsonData)
+      if (amount === undefined || decimals === undefined) {
+        throw new Error()
+      }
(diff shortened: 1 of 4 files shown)


sohkai:
> About this, any thoughts on exporting a class to use with new TokenAmount() vs. a tokenAmount() function that would do the same thing? Since this is primarily a components library, I was thinking that we might want to avoid exporting names that might conflict with component names in the future.

I like leaving it as a class—that was it is and introducing a factory just to hide the export is contrived. I wonder if we could hide it under a namespace somehow, e.g. `import { TokenAmount } from '@aragon/ui/utils'`? Or maybe under a `formatters` namespace or export?

src/utils/TokenAmount.js
```diff
@@ -0,0 +1,92 @@
...
+import { formatTokenAmount } from './format'
+
+class TokenAmount {
+  #amount
+  #decimals
+  #symbol
+
+  /**
+   * Create a TokenAmount.
+   * @param {BigInt|string|number} amount   The amount as an integer (e.g. in Wei for Ethers).
```
sohkai, line 13: Nitpick, but I think the plural for "Ether" is still "Ether"

src/utils/TokenAmount.js
```diff
@@ -0,0 +1,92 @@
...
+    this.#decimals = JSBI.BigInt(decimals)
+    this.#symbol = symbol
+  }
+
+  /**
+   * Get the amount as an integer (e.g. in Wei for Ethers).
+   * @returns {BigInt}
+   */
+  amount() {
+    return BigInt(this.#amount.toString())
```
sohkai, line 28: Should we return a copy of the JSBI object instead?

Or maybe do a quick availability check for BigInt existing natively?

src/utils/TokenAmount.js
```diff
@@ -0,0 +1,92 @@
...
+  }
+
+  /**
+   * Formats the token amount for display purposes.
+   * @param {string} options.displaySign     Whether to display the sign or not.
+   * @param {string} options.displaySymbol   Whether to display the token symbol or not.
+   * @param {string} options.digits          The number of digits to appear after the decimal point.
+   * @returns {string}
+   */
+  format({ sign = false, symbol = false, digits = 2 } = {}) {
```
sohkai, line 54: I mentioned this last in #728, but [wondering if should we adopt the same API in `formatTokenAmount()`](https://github.com/aragon/aragon-ui/pull/728#discussion_r406650540)?

src/utils/TokenAmount.js
```diff
@@ -0,0 +1,92 @@
...
+   */
+  amount() {
+    return BigInt(this.#amount.toString())
+  }
+
+  /**
+   * Get the amount as an integer (e.g. in Wei for Ethers).
+   * @returns {string}
+   */
+  amountString() {
```
sohkai, line 35: Just wondering if we _really_ need to expose this API now. It's still fairly accessible with `amount()` with just an extra `.toString()`

src/utils/TokenAmount.js
bpierre: I would avoid returning a different type implicitly: JSBI and BigInt have different APIs and we should be able to rely on one of them from the other side.

What about adding a `.amountJsbi()` method? I’d like to keep `.amount()` for `BigInt` only, as it is the only one we should use eventually.

src/utils/TokenAmount.js
bpierre: The idea is to only use `amountString()` in browsers until `BigInt` gets supported by Safari (calling `.amount()` would throw on Safari now), and use `.amount()` in Node.js.

src/utils/TokenAmount.js
bpierre: Ah yes! I agree `roundToDecimals`, we could even call it `digits` there too (from [Number#toFixed](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toFixed#Parameters)).

For `decimals`: I made it a required param in the `TokenAmount` constructor, thinking that it could be a good idea to always set it and avoid situations where code could break with non-eighteen decimals tokens.

What do you think?

src/utils/TokenAmount.js
sohkai: Oh I had assumed JSBI and BigInt were as close as possible API-wise.

What do you think about just returning a string here instead as the default API? I think we'll still be using strings to represent BNs for a while until we've all moved to BigInt... this will make it less confusing / failure-prone if we wanted to transfer a TokenAmount into one of the other BN libraries (especially web3 / ethers—they always accept strings, but the BN types are a mishmash).

src/utils/TokenAmount.js
sohkai: > For decimals: I made it a required param in the TokenAmount constructor, thinking that it could be a good idea to always set it and avoid situations where code could break with non-eighteen decimals tokens.

Yes, unfortunately we can never assume 18 decimals so we should be clear and up-front about requiring the decimals.

src/utils/TokenAmount.js
bpierre: Oh interesting, I always thought it was quantifiable unlike the other ether, and similar to the dollar.

sohkai: Yeah... AFAIK there's no usage of "Ethers", e.g. on https://ethereum.org/eth/

src/utils/TokenAmount.js
bpierre: We could, but the downside is that we will have to break compatibility at some point if we want to keep using `amount()` to return a BigInt. Something I like about `amountString()` is that it is just verbose enough to push users towards using `amount()` as soon as they can :smile:

What do you think about using a `asString` boolean parameter? `amount()` would return a `BigInt` while `amount(true)` would return a `String`.

About BigInt more generally, I think the most important thing would be to get support in Safari, because we can’t use BigInt at all without it, even as an intermediary step. I am not as worried for libraries support, because users can convert the type themselves before feeding it to their library. I’m also not too worried about libraries adding support for it: they all support string integers already, so supporting BigInt only requires a `String(value)` on the parameter.

For Safari, it seems that support for BigInt is coming in the next version :tada::tada::tada: https://developer.apple.com/safari/technology-preview/release-notes/

Ethers.js v5 will have full support for BigInt: https://github.com/ethers-io/ethers.js/issues/594#issuecomment-526807810

Web3.js seems to have [plans to support BigInt](https://github.com/ethereum/web3.js/issues/2679#issuecomment-482984819), but it doesn’t seem to be done yet.

sohkai: I'd honestly default to returning strings as the "base denominator" BN type until we see massive user adoption of BigInts, just so people can fall into the ["pit of success"](https://blog.codinghorror.com/falling-into-the-pit-of-success/) today.

I greatly prefer this over forcing users to learn how to use a new type / look up why their toolchain doesn't support BigInts natively and have to do a string cast. This should be easy.

I really don't think the string serialization into a BigInt will cost that much; if someone is serializing _that_ many strings, they probably have much bigger performance problems than the small waste of us doing a roundtrip back/forth BigInts/strings.

------

> if we want to have amount() return a BigInt

As you can tell, I'm not convinced we will want to do this at least in the next two years.

bpierre: > I'd honestly default to returning strings as the "base denominator" BN type until we see massive user adoption of BigInts, just so people can fall into the "pit of success" today.

I think it’s also a valid choice, let’s do this for now then. We have a bunch of other things returning strings rather than BigInt (e.g. `useWallet()`), so we could wait and move them all at the same time whenever it makes sense.

> As you can tell, I'm not convinced we will want to do this in the next two years. Is there a reason that makes you think exposing the BigInt type is going to be preferred in that time frame?

Two years seems really far to me if Safari support is right around the corner, but time will tell! I anticipate a quick adoption of BigInt in the web3 ecosystem once Safari supports it, at least as input values. In I might be wrong, but I think it provides clear benefits:

- No need to import and ship one or multiple libraries to deal with big integers (on both sides, libraries and apps).
- Being able to use operators directly e.g. `amount = 38n * 10n ** 18n` vs. `amount = (new BN('38')).mul((new BN('10')).pow(new BN('18'))`.
- More than operators, being able to use the rest of the language with it, e.g. `Intl.NumberFormat`.
- For libraries, no need to export non-standard types, provide the libraries to work with them, or have to deal with version incompatibilities (e.g. [Ethers’s embedded bignumber utility](https://github.com/ethers-io/ethers.js/blob/master/utils/bignumber.js)).
- Stop doing conversions everywhere between the different types (not so much for the performances cost than for the additional complexity).
- On input, supporting BigInt for libraries is trivial and can be done today, even without Safari support.
- On output (assuming support in Safari is there), it would break compatibility for existing libraries. But for authors already using BigInt, first class support for it might become an important factor when choosing a library.

src/utils/TokenAmount.js
bpierre: I removed it as this is what `.amount()` does now.

src/utils/TokenAmount.js
bpierre: Done in https://github.com/aragon/aragon-ui/pull/757/commits/5431650d51665a1fed0c8acf47d78e4d6b474d02

bpierre:
@sohkai I forgot to answer this comment:

> I like leaving it as a class—that was it is and introducing a factory just to hide the export is contrived.

I agree about the nice semantics of a constructor, my main concern is really about conflicting with a potential `<TokenAmount />` component in the future. But I also see other reasons why we would want to use a factory:

- To be consistent with [color()](https://github.com/aragon/aragon-ui/blob/master/src/utils/color.js) (same reason, `<Color />` might exist one day), and more generally, trying to reserve PascalCase for React components.
- `new` can look a bit awkward when a temporary object is needed:

```jsx
// factory
<Tag
  background={color('#FFF').alpha(0.5)}
  label={tokenAmount(amount, decimals).format()}
/>

// constructor
<Tag
  background={(new Color('#FFF')).alpha(0.5)}
  label={(new TokenAmount(amount, decimals)).format()}
/>
```

Also factories are pretty common in JS, even within the language recently (Symbol(), BigInt()), so I don’t think it would be confusing for users.

> I wonder if we could hide it under a namespace somehow, e.g. import { TokenAmount } from '@aragon/ui/utils'?

This doesn’t work, because the `main` & `module` fields of package.json only apply to the root… or we would need a `@aragon/ui/cjs/utils` and a `@aragon/ui/utils` ☹️

Also we would users would still have to do things like this if `TokenAmount` becomes a component one day:

```jsx
import { TokenAmount } from '@aragon/ui'
import { TokenAmount as TokenAmountSomething } from '@aragon/ui/utils'

function App() {
  const amount = new TokenAmountSomething(…)
  return <TokenAmount amount={amount} />
}
```

> Or maybe under a `formatters` namespace or export?

That could be an idea yes! I guess we would also need to introduce modules like `math`, `date` etc. in that case. The only issue is that it will make it impossible to tree shake inside these modules, but I guess we could split them if it becomes needed.

Now in case of conflict, users would still deal with similar names, but at least they wouldn’t have to rename their imports:

```jsx
// using utils for now because I’m not sure what could be a category for the TokenAmount class
import { utils, TokenAmount } from '@aragon/ui'

function App() {
  const amount = new utils.TokenAmount(…)
  return <TokenAmount amount={amount} />
}
```

Evalir:
> Or maybe under a `formatters` namespace or export?

I believe this is the way to go, for a few reasons:
- As @bpierre said, lets us introduce new modules for `date`, `math` utils and etc; these might be useful to expose as standalone packages.
- This also lets us do something which I think is quite nice: being able to release separate packages from aragonUI that contain these utils. [Material UI](https://material-ui.com/getting-started/installation/) does this, and I believe it gives a good separation of concerns over what is where, because right now there's a lot to think about when you import components, utilities and hooks from aragonUI. See [this pattern](https://github.com/aragon/aragon-apps/blob/master/apps/voting/app/src/components/VoteStatus.js#L4-L9) (how we do it, alphabetized) vs. [this pattern](https://github.com/1Hive/apiary/blob/master/website/src/pages/orgs.js#L5-L21), which separates components, hooks and utilities (Even though I would've personally put all components together, but, I still think it's a nice approach and you get the idea 😄).

> my main concern is really about conflicting with a potential `<TokenAmount />` component in the future.

I feel such a component would be unnatural (at least for me) — I expect a component to be purely visual from an "end-user" standpoint, and `<TokenAmount />` for me feels like trying to accomplish the same thing as having a formatting function, bundled with some HTML for display, which would end up quite coupled to the design system. 

For the reason mentioned above, I prefer the class approach; it's clear to me, and also, thinking about how _I_ would use it, I expect to pass this newly created `TokenAmount` instance to various places; I feel more comfortable having a class/object that I feel is "malleable" rather than a factory, which I feel it forces me to re-declare this variable to reuse it (doing something like `tokenAmount().format()`. doesn't feel too nice for me).

bpierre:
Closing for now, `TokenAmount` will move in its own micro library https://github.com/aragon/aragon-ui/issues/772

sohkai:
To come back to this point @Evalir:

> This also lets us do something which I think is quite nice: being able to release separate packages from aragonUI that contain these utils. Material UI does this, and I believe it gives a good separation of concerns

Agreed, and I think what we're seeing internally for frontend libraries (leading to #772) is:

- aragonUI is our collection of easily reusable React components
- A number of supporting packages (`use-x`, `token-amount`, etc.) that are single purpose and may be useful even without aragonUI
  - Over time, some utilities of aragonUI may be moved out into supporting packages, rather than publishing them as aragonUI modules
  - It is not clear how we could best organize these yet, but single repos for each is _never wrong_.

PR #758, opened 2020-04-13
CircleGraph: fix the default label for <1%

bpierre:


src/components/CircleGraph/CircleGraph.js
```diff
@@ -16,14 +16,15 @@ function labelDefault(animValue, value) {
...
-    ? parts
-    : {
-        ...parts,
-
-        // Do not display the prefix when we are only animating between 0 and 1.
-        prefix: percentage < 1 && percentage > 0 ? '<' : '',
-      }
+  const lessThanOne =
+    percentage > 0 &&
+    percentage <= 1 &&
```
sohkai, line 21: Shouldn't this be `<`?

bpierre: Ah yes! :+1:

PR #760, opened 2020-04-14
formatTokenAmount() updates

bpierre:
This PR is to address @sohkai’s comments in https://github.com/aragon/aragon-ui/pull/728.

### DIFF
--- a/src/utils/format.test.js
+++ b/src/utils/format.test.js
@@ -104,6 +104,18 @@ describe('formatTokenAmount()', () => {
     ).toEqual(`4,442,839.38129587970788395${NO_BREAK_SPACE}ANT`)
   })
 
+  test('should throw when a negative number is used for decimals', () => {
+    expect(() => {
+      formatTokenAmount(BigInt('2839000000010000000000'), -1)
+    }).toThrow()
+  })
+
+  test('should throw when a negative number is being used for digits', () => {
+    expect(() => {
+      formatTokenAmount(BigInt('2839000000010000000000'), 18, { digits: -2 })
+    }).toThrow()
+  })
+
   test('should display the sign', () => {
     expect(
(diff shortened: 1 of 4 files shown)


src/utils/math.test.js
```diff
@@ -0,0 +1,9 @@
+/* global BigInt */
+
+import { divideRoundBigInt } from './math'
+
+describe('divideRoundBigInt()', () => {
+  test('should round the result', () => {
```
sohkai, line 6: Might be worth adding another test just to smoke test the rounding works with a different base other than 10 (e.g. 5)

src/utils/format.test.js
```diff
@@ -50,72 +50,111 @@ describe('formatNumber()', () => {
...
+    ).toEqual('2,839.00000001')
+    expect(
+      formatTokenAmount(BigInt('2839000000010000000000'), 18, { digits: 9 })
+    ).toEqual('2,839.00000001')
+    expect(
+      formatTokenAmount(BigInt('2839000000010000000000'), 18, { digits: 7 })
+    ).toEqual('2,839')
+  })
+
+  test('should throw when a negative number is used for decimals', () => {
```
sohkai, line 125: Nitpick: what do you think about moving these "failure" expectations to the end?

src/utils/math.test.js
bpierre: Done https://github.com/aragon/aragon-ui/pull/760/commits/43ce5377977c2c288b3a8a1c3d7cb2ac38fc7df8

src/utils/format.test.js
bpierre: Done https://github.com/aragon/aragon-ui/pull/760/commits/ef9198496209ce5e45575bc0472b32dd9d8cc97c

PR #761, opened 2020-04-14
Add JSDoc comments

bpierre:


src/utils/miscellaneous.js
```diff
@@ -1,9 +1,23 @@
+/**
+ * Does nothing.
+ */
 export function noop() {}
 
+/**
+ * A function that only returns the first parameter passed to it.
+ * @param {any} x The first parameter
+ * @retruns {any}
```
sohkai, line 9: Small typo on `return` :)

bpierre: Noooo

Issue #731, opened 2020-03-07
Use Storybook for an isolated dev environment


Jfelix61:
What do you guys think of substituting `devbox` for [Storybook](https://storybook.js.org/)? 

## Motivations

Adding storybook can ease the development and visual testing of individual components. Also, can be used to showcase different use cases of these components.

bpierre:
I think it could work for `devbox/`, which it is directed at developers, but I am a bit reluctant to the idea of using it to showcase components, as we want something a little bit more branded than a Storybook instance. `gallery/` is doing that right now, but we want to redesign a bit and add better demos there, some coming from `devbox/`.

One concern I have with Storybook is how demos can get tied to the Storybook API, making it difficult to integrate somewhere else, but I guess we could take care of separating the demos ourselves. The integration of their visual testing service seems interesting, though I’m not sure it would be worth the cost since we are using scoped styles everywhere and visual regressions are not happening often.

If you want to give it a try, I think we could start by taking one of the devbox demos, and see how it would look as a Storybook story. From there, we could decide if moving to Storybook would be worth it. What do you think?

Jfelix61:
I was thinking that instead of starting from zero using storybook, we could use [lerna](https://github.com/lerna/lerna) and leave devbox as a package, similar to what [material-ui](https://github.com/mui-org/material-ui) does. 

My main problem when I was developing the `ToggleComponent` was updating the changes coming from the src which lerna can fix easily.

bpierre:
> My main problem when I was developing the ToggleComponent was updating the changes coming from the src which lerna can fix easily.

I should write this somewhere, but you should get the changes automatically by doing:

```console
cd aragon-ui/
yarn link
cd devbox/
yarn link @aragon/ui
```

Then:

```console
# shell session 1
yarn start
```

```console
# shell session 2
yarn devbox
```

I’m not sure Lerna would help a lot since both `devbox/` and `gallery/` are private packages, and I think it can add a bit of friction for new contributors. Do you think it could help with anything else than linking `@aragon/ui` to `devbox/` and `gallery`?

If not, perhaps we could add a script to do the linking automatically?

Jfelix61:
Thanks! I think creating a script would be the better choice, and perhaps later add lerna if needed.

sohkai:
> I should write this somewhere, but you should get the changes automatically by doing:

Maybe we should write a script and automatically do this on a local, non-CI install?

PR #763, opened 2020-04-23
devbox: update instructions

sohkai:
See https://github.com/aragon/aragon-ui/issues/731#issuecomment-596145523 (should we close it for this?).

Updates the README and adds a warning when starting the devbox if `@aragon/ui` isn't linked.

Screen Shot 2020-04-23 at 12 19 27 PM

### DIFF
--- a/devbox/package.json
+++ b/devbox/package.json
@@ -6,10 +6,11 @@
   "scripts": {
     "build": "npm run prebuild && webpack --mode production && npm run postbuild",
     "build:dev": "npm run prebuild && webpack --mode development && npm run postbuild",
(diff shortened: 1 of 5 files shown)


devbox/package.json
```diff
@@ -30,6 +31,7 @@
     "file-loader": "^5.0.2",
     "html-webpack-plugin": "^3.2.0",
     "react-axe": "^3.3.0",
+    "symlinked": "^0.5.0",
```
sohkai, line 34: [This is not the most complex package](https://github.com/ryanve/symlinked/blob/master/index.js), but I think it's fine as a dev dependency?

PR #769, opened 2020-05-19
re-align RadioListItem button to top of component

stvtortora:
Fixes aragon/aragon#1340

### DIFF
--- a/src/components/Radio/RadioListItem.js
+++ b/src/components/Radio/RadioListItem.js
@@ -25,7 +25,6 @@ const RadioListItem = React.memo(function RadioListItem({
         id={index}
         css={`
           flex-shrink: 0;
-          margin-top: 1px;
         `}
       />
       <div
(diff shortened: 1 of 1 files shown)


stvtortora:
Hi, any update on this one?

bpierre:
Hi @stvtortora, sorry for the delay, I am checking now.

I noticed the spacing wasn’t exactly the same as the Figma file (1px vs. 4px). The default on `Radio` is 4px already, so I think we could remove the line altogether?

stvtortora:
Hi @bpierre. No worries. I removed that line, but I don't think its what we want because no default value took its place. As far as I can tell, there does not appears to be a file (or any code for that matter) with "Figma" in this repository.

bpierre:
> As far as I can tell, there does not appears to be a file (or any code for that matter) with "Figma" in this repository.

Ah sorry I should have been more explicit, I was referring to the link to Figma in the issue: https://www.figma.com/file/bZ1j7omfhdMCbtFahP3nEJ9B/Client?node-id=525%3A1141

I was just testing and it seems correct now?

stvtortora:
@bpierre Ah, ok, thanks for clarifying! It looks good to me too.

PR #770, opened 2020-05-24
TextInput: refresh adornment API

### LINKED ISSUE #656: TextInput: refresh the adornment API
See discussion: https://github.com/aragon/aragon-ui/pull/612#discussion_r342338026

Might also fix https://github.com/aragon/aragon-ui/issues/534 at the same time.

stvtortora:
Fixes #656 

Do changes need to be made where this component is utilized throughout your other repositories?

As a heads up, I accidentally added the changes from https://github.com/aragon/aragon-ui/pull/769 to this PR because I initially forgot to create a new branch.

### DIFF
--- a/devbox/apps/Input.js
+++ b/devbox/apps/Input.js
@@ -27,25 +27,23 @@ function App() {
         </form>
         <TextInput
           css="width: 80px"
-          adornment="ETH"
-          adornmentPosition="end"
-          adornmentSettings={{
-            width: 55,
-            padding: 8,
+          adornment={{
+            end: 'ETH',
+            endWidth: 55,
+            endPadding: 8,
           }}
         />
         <TextInput
           css="width: 80px"
-          adornment="$"
-          adornmentPosition="start"
-          adornmentSettings={{
-            width: 36,
-            padding: 8,
+          adornment={{
+            start: '$',
+            startWidth: 36,
+            startPadding: 8,
           }}
         />
         <TextInput />
-        <TextInput adornment={<IconBlank />} adornmentPosition="start" />
-        <TextInput adornment={<IconBlank />} adornmentPosition="end" />
+        <TextInput adornment={{ start: <IconBlank /> }} />
+        <TextInput adornment={{ end: <IconBlank /> }} />
         <TextInput.Multiline />
         <SearchInput
           value={searchTerm}
--- a/devbox/apps/Input.js
+++ b/devbox/apps/Input.js
@@ -44,6 +44,27 @@ function App() {
         <TextInput />
         <TextInput adornment={{ start: <IconBlank /> }} />
         <TextInput adornment={{ end: <IconBlank /> }} />
+        <TextInput
+          css="width: 80px"
+          adornment="ETH"
+          adornmentPosition="end"
+          adornmentSettings={{
+            width: 55,
+            padding: 8,
+          }}
+        />
+        <TextInput
+          css="width: 80px"
+          adornment="$"
+          adornmentPosition="start"
+          adornmentSettings={{
+            width: 36,
+            padding: 8,
+          }}
+        />
+        <TextInput />
+        <TextInput adornment={<IconBlank />} adornmentPosition="start" />
+        <TextInput adornment={<IconBlank />} adornmentPosition="end" />
         <TextInput.Multiline />
         <SearchInput
           value={searchTerm}
--- a/src/components/Input/TextInput.js
+++ b/src/components/Input/TextInput.js
@@ -188,7 +188,6 @@ const WrapperTextInput = React.forwardRef(({ adornment, ...props }, ref) => {
 WrapperTextInput.propTypes = {
   ...TextInput.propTypes,
   adornment: PropTypes.oneOf([
-    PropTypes.node,
     PropTypes.shape({
       start: PropTypes.node,
       startWidth: PropTypes.number,
@@ -197,7 +196,12 @@ WrapperTextInput.propTypes = {
       endWidth: PropTypes.number,
       endPadding: PropTypes.number,
     }),
+    
+    // deprecated
+    PropTypes.node,
   ]),
+
+  // deprecated
   adornmentPosition: PropTypes.oneOf(['start', 'end']),
   adornmentSettings: PropTypes.shape({
     width: PropTypes.number,
--- a/src/components/Input/TextInput.js
+++ b/src/components/Input/TextInput.js
@@ -84,7 +84,7 @@ TextInput.defaultProps = {
   type: 'text',
 }
 
-const Adornment = ({ adornment, position, padding }) => {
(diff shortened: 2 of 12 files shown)


devbox/apps/Input.js
```diff
@@ -25,6 +25,25 @@ function App() {
...
+            startWidth: 36,
+            startPadding: 8,
+          }}
+        />
+        <TextInput />
+        <TextInput adornment={{ start: <IconBlank /> }} />
+        <TextInput adornment={{ end: <IconBlank /> }} />
         <TextInput
           css="width: 80px"
           adornment="ETH"
```
sohkai, line 49: For these older examples, perhaps we can either remove them or update them to use the new API?

src/components/Input/TextInput.js
```diff
@@ -84,66 +84,120 @@ TextInput.defaultProps = {
...
+const Adornment = React.memo(({ adornment, position, padding }) => {
+  const theme = useTheme()
+  return (
+    <div
+      css={`
+        position: absolute;
+        top: 0;
+        bottom: 0;
+        height: 100%;
+        ${position === 'end' ? 'right' : 'left'}: ${padding}px;
```
sohkai, line 96: This wasn't covered previously, but we should double check that `padding` has been defined before setting this.

src/components/Input/TextInput.js
```diff
@@ -84,66 +84,120 @@ TextInput.defaultProps = {
...
+  adornment: PropTypes.node,
+  position: PropTypes.oneOf(['start', 'end']),
+  padding: PropTypes.number,
+}
+
+Adornment.defaultProps = {
+  padding: 4,
+}
+
+const deprecationWarning = (adornmentPosition, adornmentSettings) => {
```
sohkai, line 118: Let's inline this logic, since it's not used anywhere else :).

src/components/Input/TextInput.js
```diff
@@ -84,66 +84,120 @@ TextInput.defaultProps = {
...
+      startWidth: PropTypes.number,
+      startPadding: PropTypes.number,
+      end: PropTypes.node,
+      endWidth: PropTypes.number,
+      endPadding: PropTypes.number,
+    }),
+  ]),
   adornmentPosition: PropTypes.oneOf(['start', 'end']),
   adornmentSettings: PropTypes.shape({
     width: PropTypes.number,
```
sohkai, line 203:
```suggestion

  // deprecated
  adornmentPosition: PropTypes.oneOf(['start', 'end']),
  adornmentSettings: PropTypes.shape({
    width: PropTypes.number,
```

src/components/Input/TextInput.js
```diff
@@ -154,8 +208,6 @@ WrapperTextInput.propTypes = {
 WrapperTextInput.defaultProps = {
   ...TextInput.defaultProps,
   adornment: null,
```
sohkai, line 210: I'm not sure, but should we add `null` to the list of allowed types for `adornment`? Not sure how `PropTypes.oneOf()` handles this, but I could see it erroring.

src/components/Input/TextInput.js
```diff
@@ -84,66 +84,120 @@ TextInput.defaultProps = {
...
+  adornment: PropTypes.oneOf([
+    PropTypes.node,
+    PropTypes.shape({
+      start: PropTypes.node,
+      startWidth: PropTypes.number,
+      startPadding: PropTypes.number,
+      end: PropTypes.node,
+      endWidth: PropTypes.number,
+      endPadding: PropTypes.number,
+    }),
```
sohkai, line 199:
```suggestion
    PropTypes.shape({
      start: PropTypes.node,
      startWidth: PropTypes.number,
      startPadding: PropTypes.number,
      end: PropTypes.node,
      endWidth: PropTypes.number,
      endPadding: PropTypes.number,
    }),
    
    // deprecated
    PropTypes.node,
```

src/components/Input/TextInput.js
```diff
@@ -84,66 +84,120 @@ TextInput.defaultProps = {
...
+  if (!adornment) {
+    return <TextInput ref={ref} {...props} />
+  }
+
+  let adornmentConfig = adornment
+
+  const usingDeprecatedAPI =
+    React.isValidElement(adornment) ||
+    typeof adornment === 'string' ||
+    (typeof adornment === 'object' && adornment.constructor === Array)
```
sohkai, line 144: I would also assume a user wants the old API if they use `adornmentPosition` and `adornmentSettings`, and widen the `PropTypes.node` check a bit:

E.g. something like:

```js
(React.isValidElement(adornment) || typeof adornment !== 'object' || Array.isArray(adornment)) ||
  adornmentPosition ||
  adornmentSettings
```

The main part that changes for the `PropTypes.node` check is changing `typeof adornment === 'string'` to `typeof adornment !== object`. `PropTypes.node` covers numbers and potentially other renderable types.

src/components/Input/TextInput.js
```diff
@@ -84,66 +84,120 @@ TextInput.defaultProps = {
...
+      `}
+    >
+      <TextInput
+        ref={ref}
         css={`
-          display: inline-flex;
-          position: relative;
-          width: ${props.wide ? '100%' : 'max-content'};
+          ${start && `padding-left: ${startWidth}`}
+          ${end && `padding-right: ${endWidth}`}
```
sohkai, line 176: It looks like the backwards compatibility above might not reproduce this exactly, since it just uses the `adornmentSettings.padding` as the padding.

src/components/Input/TextInput.js
```diff
@@ -154,8 +208,6 @@ WrapperTextInput.propTypes = {
 WrapperTextInput.defaultProps = {
   ...TextInput.defaultProps,
   adornment: null,
-  adornmentPosition: 'start',
-  adornmentSettings: {},
 }
 
 // <input type=number> (only for compat)
```
sohkai, line 213: Could we also update this component's [documentation](https://github.com/aragon/aragon-ui/blob/master/src/components/Input/TextInput.md) with the new changes? We can remove (or mark) the deprecated props :).

src/components/Input/TextInput.js
Evalir: Actually, there's a `defaultProp` defined for padding 🙌 it's `4`.

src/components/Input/TextInput.js
Evalir: Sadly, React [doesn't seem to have `null`](https://reactjs.org/docs/typechecking-with-proptypes.html) as an available `propType`. :(

src/components/Input/TextInput.js
Evalir: Done in https://github.com/aragon/aragon-ui/pull/770/commits/aa2b33f07f063df32fcb9212ad86339e3ffcebf5

PR #777, opened 2020-06-10
format: add tests to handle input edge cases (zero decimals)

sohkai:
Only added additional test cases that we should handle for now!


bpierre:
@sohkai Done. For decimal units greater than digits, I was thinking that [we could reasonably tolerate it](https://github.com/aragon/aragon-ui/pull/777/commits/32869eff54937b9f910052df7759635a793a07da#diff-f94a805a1179ce07d38671f60c27aca4R143-R147), what do you think?

PR #780, opened 2020-06-11
Dark theme: Increase positive content contrast

andy-hook:
I found `positiveSurfaceContent` to lack contrast in the dark theme. It might just be the context i've been using it but even still it felt on the low side.

image


bpierre:
Agreed. @owisixseven @dizzypaty what do you think?

dizzypaty:
> Agreed. @owisixseven @dizzypaty what do you think?

I suggest we use `2CC68F` instead, which will provide even higher contrast and is already part of the theme. 

Thanks for catching this Andy!!

Issue #790, opened 2020-07-17
Modal: EscapeOutside causing premature modal closing with internal interactions


andy-hook:
I've run up against an issue multiple times where `EscapeOutside` checks for the existence of the target element too late in the event queue, causing a premature closing of the modal.

This is common when you want to hide an interactive element after it's been clicked. For example you might store some state on whether to show or hide a button:

```javascript
const [buttonVisible, setButtonVisible] = useState(true)

......
<Modal>
    {buttonVisible && <Button onClick={() => setButtonVisible(false)}/>}
</Modal>
```

`Button` is clicked (modal catches click event)
State is updated and `Button` is removed from DOM
`EscapeOutside` [searches the contents of the modal for the event.target node](https://github.com/aragon/aragon-ui/blob/400b147028f6444fa4ec4415a45ae00f0d0004ab/src/components/EscapeOutside/EscapeOutside.js#L33)
`EscapeOutside` deems this as an outside click because the `Button` was removed before the modal click fired.

This problem is often masked when using animations as nodes stick around long enough to be checked.

It might just be a case of using `onmousedown` instead of `click`. Another option could be to assign a handler to the background scrim that closes the modal, that way the intent is explicit and we sidestep dealing with `event.target`.

PR #796, opened 2020-07-24
Modal: Prevent EscapeOutside from causing problems with internal interactions

andy-hook:
resolves https://github.com/aragon/aragon-ui/issues/790

I've added a `useCapture` prop to `EscapeOutside` to allow us to optionally change the event propagation from bubbling to capture, this allows us to reverse execution order on the events, this is especially important for the `Modal` so that outside interaction checks can happen before others potentially affect render state.

More info: https://developer.mozilla.org/en-US/docs/Learn/JavaScript/Building_blocks/Events#Event_bubbling_and_capture

image

### DIFF
--- a/src/components/EscapeOutside/README.md
+++ b/src/components/EscapeOutside/README.md
@@ -32,7 +32,7 @@ The callback to call after clicking outside of the wrapped element, or when pres
 
 If `useCapture` is true, the event will be registered in the capturing phase and thus, propagated top-down instead of bottom-up as is the default.
 
-See (MDN web docs)[https://developer.mozilla.org/en-US/docs/Learn/JavaScript/Building_blocks/Events#Event_bubbling_and_capture] for more information on event bubbling vs capture.
+See the (MDN web docs)[https://developer.mozilla.org/en-US/docs/Learn/JavaScript/Building_blocks/Events#Event_bubbling_and_capture] for more information on event bubbling vs capture.
 
 ### `children`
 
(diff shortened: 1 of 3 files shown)


src/components/Modal/Modal.js
```diff
@@ -79,6 +79,7 @@ function Modal({
                 >
                   <EscapeOutside
                     role="alertdialog"
+                    useCapture
```
sohkai, line 82: Just not sure about making this the default since it may be considered a breaking change in behaviour.

However, since we don't provide a mechanism to turn this off in the Modal, I assume this is actually desired behaviour?

andy-hook: For the modal I can't think of a case where this wouldn't be the desired behaviour.

andy-hook:
@sohkai @bpierre – Is this good to go? would we be confident to ship it under a minor release?

src/components/EscapeOutside/README.md
```diff
@@ -0,0 +1,43 @@
...
+
+### `useCapture`
+
+| Type      | Default value |
+| --------- | ------------- |
+| `Boolean` | false         |
+
+If `useCapture` is true, the event will be registered in the capturing phase and thus, propagated top-down instead of bottom-up as is the default.
+
+See (MDN web docs)[https://developer.mozilla.org/en-US/docs/Learn/JavaScript/Building_blocks/Events#Event_bubbling_and_capture] for more information on event bubbling vs capture.
```
Evalir, line 35: Just a super big nitpick 😅 
```suggestion
See the (MDN web docs)[https://developer.mozilla.org/en-US/docs/Learn/JavaScript/Building_blocks/Events#Event_bubbling_and_capture] for more information on event bubbling vs capture.
```

andy-hook: 🙌

PR #797, opened 2020-07-27
Token amount component

delfipolito:
New component for rendering a token logo, an amount and the token symbol.
(Updated eslint version so it supports "bigint" as a valid-typeof rule )

image

image

image

image

image


src/components/TokenAmount/TokenAmount.js
```diff
@@ -0,0 +1,123 @@
...
+            {new TokenAmount(amount, decimals).format({ digits: digits })}
+          </span>
+        )}
+        <Symbol address={address} size={size} symbol={symbol} />
+      </div>
+    </UseTokenProvider>
+  )
+})
+
+const Logo = React.memo(function Logo({ address, size }) {
```
bpierre, line 54: We could probably drop the `React.memo()` here, as the parent is already memoized.

src/components/TokenAmount/TokenAmount.js
```diff
@@ -0,0 +1,123 @@
...
+            `}
+            src={token.iconUrl}
+          />
+        )
+      }
+    </ImageExists>
+  )
+})
+
+const Symbol = React.memo(function Symbol({ address, size, symbol }) {
```
bpierre, line 74: Same here, I think we could drop the `React.memo()`.

src/components/TokenAmount/TokenAmount.js
```diff
@@ -0,0 +1,123 @@
...
+  const token = useToken(address)
+  const theme = useTheme()
+
+  return (
+    <span
+      css={`
+        color: ${theme.surfaceContentSecondary};
+        ${textStyle('body2')};
+        line-height: 1;
+        ${size === 'large' && `align-self: flex-end;`}
```
bpierre, line 84: In `css`, it’s important to always do this:

```suggestion
        ${size === 'large' ? `align-self: flex-end;` : ''}
```

Otherwise it will convert to the `"false"` string in the CSS declaration block, which will break the next declaration if present.

src/components/TokenAmount/TokenAmount.js
```diff
@@ -0,0 +1,123 @@
...
+}
+
+TokenAmountComponent.propTypes = {
+  address: PropTypes.string.isRequired,
+  amount: PropTypes.any,
+  className: PropTypes.string,
+  decimals: PropTypes.number,
+  digits: PropTypes.number,
+  networkType: PropTypes.string,
+  size: PropTypes.string,
```
bpierre, line 110: We could use `Proptypes.oneOf()` here ([e.g. in Button](https://github.com/aragon/aragon-ui/blob/master/src/components/Button/Button.js#L313-L321)).

src/components/TokenAmount/TokenAmount.js
```diff
@@ -0,0 +1,123 @@
...
+
+Symbol.propTypes = {
+  address: PropTypes.string.isRequired,
+  size: PropTypes.string,
+  symbol: PropTypes.string,
+}
+
+TokenAmountComponent.propTypes = {
+  address: PropTypes.string.isRequired,
+  amount: PropTypes.any,
```
bpierre, line 105: Should we add a `Proptypes._bigIntish` [in the custom prop types](https://github.com/aragon/aragon-ui/blob/master/src/proptypes.js)?

src/components/TokenAmount/TokenAmount.js
```diff
@@ -0,0 +1,123 @@
...
+  className: PropTypes.string,
+  decimals: PropTypes.number,
+  digits: PropTypes.number,
```
bpierre, line 108: I think we could also accept a `BigInt`, `Number`, `String` or object with `.toString()` (i.e. the `Proptypes._bigIntish` type), since that’s what [token-amount does](https://github.com/aragon/token-amount/blob/master/src/format.js#L30-L31).

src/components/TokenAmount/TokenAmount.js
```diff
@@ -0,0 +1,123 @@
...
+  const isValidAddress = isAddress(address)
+  if (!isValidAddress) {
+    warn(`TokenAmount: provided invalid address (${address})`)
+  }
+
+  return (
+    <UseTokenProvider>
+      <div
+        className={className}
+        css={`
```
bpierre, line 30: Nitpick: we usually always move the `css` prop at the end (since it’s not really a prop but gets transformed by babel).

src/components/TokenAmount/TokenAmount.js
```diff
@@ -0,0 +1,123 @@
...
+        <Logo address={address} size={size} />
+        {amount && (
+          <span
+            css={`
+              padding-right: ${size === 'large' ? 0.5 * GU : 0.25 * GU}px;
+              ${size === 'large' ? textStyle('title2') : textStyle('body2')};
+              line-height: 1;
+            `}
+          >
+            {new TokenAmount(amount, decimals).format({ digits: digits })}
```
bpierre, line 45: We could use the static function since we are only formatting here:

```suggestion
            {TokenAmount.format(amount, decimals, { digits: digits })}
```

src/components/TokenAmount/TokenAmount.js
```diff
@@ -0,0 +1,123 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import { useToken, UseTokenProvider } from 'use-token'
+import TokenAmount from 'token-amount'
+import { ImageExists } from '../../hooks'
+import { GU, textStyle } from '../../style'
+import { isAddress, warn } from '../../utils'
+import { useTheme } from '../../theme'
+
+const TokenAmountComponent = React.memo(function TokenAmountComponent({
```
bpierre, line 10: We should at least name our function `TokenAmount` so that the name appears correct in other places:

```suggestion
const TokenAmountComponent = React.memo(function TokenAmount({
```

But I was wondering if we should rename `TokenAmount` into `TokenAmountLib`, and just use `TokenAmount` for the component? :thinking:

delfipolito: sounds good!

gallery/src/pages/PageTokenAmount.js
```diff
@@ -0,0 +1,43 @@
+import React from 'react'
```
bpierre, line 1: About this page, what would you think of making the demo as simple as we can?

It could look like this:



I made some changes here: https://git.io/JJodD

src/components/TokenAmount/README.md
```diff
@@ -0,0 +1,98 @@
...
+| -------- | ------------- |
+| `string` | None          |
+
+The mainnet token address.
+
+### `amount`
+
+| Type                     | Default value |
+| ------------------------ | ------------- |
+| `bigInt, number, string` | None          |
```
bpierre, line 48:
```suggestion
| `BigInt`, `Number`, `String` | None (required)          |
```

src/components/TokenAmount/README.md
```diff
@@ -0,0 +1,98 @@
...
+
+## Props
+
+### `address`
+
+| Type     | Default value |
+| -------- | ------------- |
+| `string` | None          |
+
+The mainnet token address.
```
bpierre, line 42:
```suggestion
The address of the token contract, on the specified `chainId` (only Ethereum Mainnet is supported at the moment).
```

src/components/TokenAmount/README.md
```diff
@@ -0,0 +1,98 @@
...
+| ------------------------ | ------------- |
+| `bigInt, number, string` | None          |
+
+The token amount.
+
+### `chainId`
+
+| Type     | Default value |
+| -------- | ------------- |
+| `number` | 1             |
```
bpierre, line 56:
```suggestion
| `Number` | `1`             |
```

src/components/TokenAmount/README.md
```diff
@@ -0,0 +1,98 @@
...
+
+Ethereum chain ID. We only support mainnet for now.
```
bpierre, line 58:
```suggestion
The [chain ID](https://chainid.network/) for the token. Only Ethereum Mainnet (`1`) is supported at the moment.
```

src/components/TokenAmount/README.md
```diff
@@ -0,0 +1,98 @@
...
+
+### `decimals`
+
+| Type     | Default value |
+| -------- | ------------- |
+| `number` | 18            |
```
bpierre, line 64: I’m just thinking now that this should be required, to prevent amounts to get displayed with a wrong number of decimals:

```suggestion
| `BigInt`, `Number`, `String` | None (required)            |
```

src/components/TokenAmount/README.md
```diff
@@ -0,0 +1,98 @@
...
+
+Decimal placement for amount.
```
bpierre, line 66:
```suggestion
The number of decimals for the token amount, [as specified by the token contract](https://eips.ethereum.org/EIPS/eip-20#decimals).
```

src/components/TokenAmount/README.md
```diff
@@ -0,0 +1,98 @@
...
+
+### `digits`
+
+| Type     | Default value |
+| -------- | ------------- |
+| `number` | 2             |
```
bpierre, line 72:
```suggestion
| `BigInt`, `Number`, `String` | `2`             |
```

src/components/TokenAmount/README.md
```diff
@@ -0,0 +1,98 @@
...
+
+Rounds the number to a given decimal place.
+
+### `iconUrl`
+
+| Type     | Default value |
+| -------- | ------------- |
+| `string` | none          |
```
bpierre, line 80:
```suggestion
| `String` | None          |
```

src/components/TokenAmount/README.md
```diff
@@ -0,0 +1,98 @@
...
+
+Overrides the iconUrl given by the token address.
```
bpierre, line 82:
```suggestion
Overrides the icon URL given by the token address.
```

src/components/TokenAmount/README.md
```diff
@@ -0,0 +1,98 @@
...
+
+### `size`
+
+| Type     | Default value |
+| -------- | ------------- |
+| `string` | medium        |
```
bpierre, line 88:
```suggestion
| `"large"` or `"medium"` | `"medium"`        |
```

src/components/TokenAmount/README.md
```diff
@@ -0,0 +1,98 @@
...
+
+Size of the amount text.
```
bpierre, line 90:
```suggestion
Size of the component.
```

src/components/TokenAmount/README.md
```diff
@@ -0,0 +1,98 @@
...
+
+### `symbol`
+
+| Type     | Default value |
+| -------- | ------------- |
+| `string` | none          |
```
bpierre, line 96:
```suggestion
| `String` | None          |
```

src/components/TokenAmount/TokenAmount.js
```diff
@@ -0,0 +1,127 @@
...
+  address: PropTypes.string,
+  size: PropTypes.oneOf(['large', 'medium']),
+  symbol: PropTypes.string,
+}
+
+TokenAmount.propTypes = {
+  address: PropTypes.string,
+  amount: PropTypes._bigIntish,
+  chainId: PropTypes.number,
+  decimals: PropTypes.number,
```
bpierre, line 113: As I was saying in another comment above, I think `decimals` should be required, at least until we have it in `useToken()`.

```suggestion
  decimals: PropTypes.number.isRequired,
```

src/components/TokenAmount/TokenAmount.js
```diff
@@ -0,0 +1,127 @@
...
+
+Symbol.propTypes = {
+  address: PropTypes.string,
+  size: PropTypes.oneOf(['large', 'medium']),
+  symbol: PropTypes.string,
+}
+
+TokenAmount.propTypes = {
+  address: PropTypes.string,
+  amount: PropTypes._bigIntish,
```
bpierre, line 111:
```suggestion
  amount: PropTypes._bigIntish.isRequired,
```

src/components/TokenAmount/TokenAmount.js
```diff
@@ -0,0 +1,127 @@
...
+      {symbol || token.symbol}
+    </span>
+  )
+}
+
+Icon.propTypes = {
+  address: PropTypes.string,
+  chainId: PropTypes.number,
+  iconUrl: PropTypes.string,
+  size: PropTypes.oneOf(['large', 'medium', 'small', 'mini']),
```
bpierre, line 100:
```suggestion
  size: PropTypes.oneOf(['large', 'medium']),
```

src/components/TokenAmount/TokenAmount.js
```diff
@@ -0,0 +1,127 @@
...
+  decimals: PropTypes.number,
+  digits: PropTypes.number,
+  iconUrl: PropTypes.string,
+  size: PropTypes.oneOf(['large', 'medium']),
+  symbol: PropTypes.string,
+}
+
+TokenAmount.defaultProps = {
+  chainId: 1,
+  decimals: 18,
```
bpierre, line 122:
```suggestion
```

bpierre:
And one last thing, I think we should correct the alignment of the amount in the `large` size:

image


This is how it could look by moving the amount down by 2px and the symbol up by 1px:

image


What do you think?

src/proptypes.js
```diff
@@ -53,8 +55,32 @@ function _element(props, propName, componentName) {
...
+function _bigIntish(props, propName, componentName) {
+  if (isEmpty(props[propName])) {
+    return null
+  }
+
+  if (typeof props[propName] === 'number') {
+    return null
+  }
+
+  if (typeof props[propName] === 'bigint') {
```
bpierre, line 67: ESLint gives me an error here, even though `bigint` is valid: https://eslint.org/docs/rules/valid-typeof

I think we might want to bump ESLint to the latest version (7.6.0) so it recognizes `bigint` as a valid `typeof` comparison.

src/components/TokenAmount/TokenAmount.js
```diff
@@ -47,6 +47,7 @@ const TokenAmount = React.memo(function TokenAmount({
               padding-right: ${size === 'large' ? 0.5 * GU : 0.25 * GU}px;
               ${textStyle(size === 'large' ? 'title2' : 'body2')};
               line-height: 1;
+              ${size === 'large' ? `margin-bottom: -2px;` : ''}
```
bpierre, line 50: Just to be safe, we could use `transform: translateY(-2px)` so that it can’t impact any surrounding element:

```suggestion
              ${size === 'large' ? `transform: translateY(2px);` : ''}
```

src/components/TokenAmount/TokenAmount.js
```diff
@@ -86,6 +87,7 @@ const Symbol = function Symbol({ address, size, symbol }) {
         ${textStyle('body2')};
         line-height: 1;
         ${size === 'large' ? `align-self: flex-end;` : ''}
+        ${size === 'large' ? `margin-bottom: 1px;` : ''}
```
bpierre, line 90:
```suggestion
        ${size === 'large' ? `transform: translateY(-1px);` : ''}
```

Issue #804, opened 2020-08-19
Layout: Expose provider independently of container


andy-hook:
Currently `<Layout/>` renders both a provider and a container element that set's widths based on breakpoints. Having these static widths applied is not always desired outside of client apps, and in some cases it would be nice to control all aspects of the layout independently of the provider eg: Wanting to update the display of an element at a specific breakpoint but it sits outside the bounds of the width limited container.

This can be worked around by combined usage of [useViewport](https://github.com/aragon/use-viewport) but because core layout elements in this kit depend on `Layout` you end up with a mixture of `useLayout` combined with `useViewport` which is not ideal and can cause overlap or misalignment if not careful.

In the short term I think it would be nice to expose access to just the provider while leaving `<Layout/>` with its current functionality.

PR #805, opened 2020-08-19
Layout: Expose Provider

andy-hook:
Resolves https://github.com/aragon/aragon-ui/issues/804


src/components/Layout/Layout.js
```diff
@@ -29,6 +36,10 @@ function getLayoutSize(parentWidth, breakpoints) {
   return sizes[0]
 }
 
+function useMergeBreakpoints(breakpoints) {
+  return useMemo(() => ({ ...BREAKPOINTS, ...breakpoints }), [breakpoints])
```
bpierre, line 40: `breakpoints` will probably change on every render if it’s passed to `LayoutProvider`. Should we use `JSON.stringify(breakpoints)` as a key?

( `JSON.stringify()` is probably faster than `Object.keys(a).join() + Object.values(a).join()`. )

src/components/Layout/Layout.js
```diff
@@ -50,12 +61,39 @@ function Layout({
...
+  parentWidth: PropTypes.number,
+}
+
+Layout.defaultProps = {
+  breakpoints: {},
+  paddingBottom: 3 * GU,
+}
+
+// Can be used to build an alternative Layout component
+Layout.__Context = LayoutContext
```
bpierre, line 92: If we plan to use it, should we make it part of the public API by removing the `__`?

bpierre (approved):
LGTM!

src/components/Layout/Layout.js
andy-hook: Definitely, good spot!

src/components/Layout/Layout.js
andy-hook: I was curious about why this exists. What was the original use case? it feels redundant.

bpierre: Oh I completely forgot I added this, I thought it was a new change. It was added [in this PR](https://github.com/aragon/aragon-ui/pull/538) that was made [for Apiary](https://github.com/1Hive/apiary/pull/52), to do [the exact same thing](https://github.com/1Hive/apiary/blob/3fb0075f32ad418d3f6bc42289c99a98c7de3484/website/src/components/Layout/Layout.js) we need to do now :smile:

Apiary is still using aragonUI 1.0 and this was a private feature, so I think we can just remove it and recommend them to move to `LayoutProvider` whenever they decide to upgrade aragonUI.

PR #806, opened 2020-08-20
TokenAmount: Hold token space during image load

andy-hook:
I ran into a problem where expandable elements such as those on `DataView` and `Accordion` were measuring height dimensions incorrectly because `TokenAmount` wasn't mounted prior and hadn't loaded token images to render space.

A solution is to always provide the dimensions and a placeholder during load, this also stops content hopping around.

image

This does present the question of whether a placeholder being present when a fake token is used is acceptable. I think we should optimise for the common use case (a token address has an asset) and if we need to we can give the option to hide the token graphic via a flag in the future.

### DIFF
--- a/src/components/TokenAmount/TokenAmount.js
+++ b/src/components/TokenAmount/TokenAmount.js
@@ -44,10 +44,9 @@ const TokenAmount = React.memo(function TokenAmount({
         {amount && (
           <span
             css={`
-              padding-right: ${size === 'large' ? 0.5 * GU : 0.25 * GU}px;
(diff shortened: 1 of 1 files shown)


bpierre (approved):
> This does present the question of whether a placeholder being present when a fake token is used is acceptable. I think we should optimise for the common use case (a token address has an asset) and if we need to we can give the option to hide the token graphic via a flag in the future.

Agreed, I think having a placeholder is also making the component more consistent when displayed in a list. /cc @dizzypaty

All good for me!

PR #810, opened 2020-09-25
DataView: Prevent expansion close on re-render

andy-hook:
Spring seems to have some internal logic for animating to `auto` which is nice and means the expanded node will keep flow on resize and when sorting entries.


bpierre (approved):
LGTM :100:

PR #811, opened 2020-09-28
DataView: Explicit auto over spring internals

andy-hook:
This is an alternative approach to https://github.com/aragon/aragon-ui/pull/810 without relying on spring to handle auto. This is superior in some ways because values always remain accurate and up to date, transitions are also preserved when reordering.

Only drawback is that the height of the expansion it remeasured more often than is needed (while it's open) but without changing the api we have no reliable and unique information to use as dependencies other than the content nodes (which could arguably be more expensive and more complex to compare)

ezgif-4-c8421b2b27ad

### DIFF
--- a/src/components/DataView/TableView.js
+++ b/src/components/DataView/TableView.js
@@ -333,8 +333,19 @@ function EntryExpansion({
 }) {
   const theme = useTheme()
 
-  const height = expansion.freeLayout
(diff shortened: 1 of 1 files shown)


sohkai (approved):
LGTM, thanks for the additional comments!

PR #817, opened 2021-02-17
Checkbox [fix]: prevent default onkeypress action

### LINKED ISSUE #626: Checkbox, Radio: avoid toggling on 'enter'
From https://github.com/aragon/aragon-ui/pull/620#discussion_r335479434:

> the behavior of Checkbox is wrong: the enter key should be used to submit forms (if any) when a checkbox is focused, not to toggle the checkbox itself ([reference](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/checkbox_role#Keyboard_interactions)).

eliobricenov:
### Description

Given that the `<Checkbox/>` component renders a `<span>` element with a `<button>` element as a parent, by default, the `Enter` key triggers the `keypress` event. 

This PR disables the default action on the `keypress` event that triggers the `click` event later on.

**Note**: the checkbox can still be checked with the `Space` key thanks to the `role` attribute.

Resolves #626

### DIFF
--- a/src/components/Input/Checkbox.js
+++ b/src/components/Input/Checkbox.js
@@ -35,7 +35,10 @@ class Checkbox extends React.PureComponent {
     if (checked) return 'true'
     return 'false'
   }
-  handleClick = () => {
+  handleKeyPress = e => {
+    e.preventDefault()
+  }
+  handleClick = e => {
     const { onChange, checked, indeterminate } = this.props
     onChange(indeterminate ? false : !checked)
   }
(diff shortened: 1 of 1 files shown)


Horus78:
Ufs...

nivida (changes requested):
The change itself looks good. But was it required to update the dependencies as for example ``rollup/plugin-node-resolve``? I especially ask this because it is a major version increase. 

Overall we can say that changes per PR should be kept at a minimum. This means the description of a PR and the related issue/task defines the scope of it and any further changes have to be done in a separate PR with a related task.

nivida:
@eliobricenov Thanks for the insights! :) Feel free to create a task on Linear related to the required dependency updates and add it to the circle. I think it will be a quick thing for you to open a PR for and we would better see the work you already have done :-)

nivida (approved):

PR #859, opened 2021-09-22
feature / DAO-324 

sepehr2github:
- remove input adornment border
- change search input adornment position

### DIFF
--- a/src/components/Input/SearchInput.js
+++ b/src/components/Input/SearchInput.js
@@ -51,7 +51,7 @@ const SearchInput = React.forwardRef(({ onChange, ...props }, ref) => {
           </ButtonIcon>
         )
       }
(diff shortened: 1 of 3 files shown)


src/components/Input/SearchInput.js
```diff
@@ -51,7 +51,7 @@ const SearchInput = React.forwardRef(({ onChange, ...props }, ref) => {
           </ButtonIcon>
         )
       }
-      adornmentPosition="end"
+      adornmentPosition="start"
       onChange={handleChange}
       {...props}
```
yuetloo, line 56: Is this going to change the searchInput adornment from "end" to "start"?  Is it possible to try to make this backward compatible by accepting option for "end" or "start" and default to "start"?

PR #871, opened 2022-01-07
Fix: Rewrite types definitions

kamikazebr:
List all componens name and status if already converted to ts

https://hackmd.io/y33nfwoTSAiPIm4Efo9kUQ

Aproxx 250 components, all done first phase!


kamikazebr:
My bad! Its suppose to go to another repo!

But btw i'm writing typescript definitions for the components.

